PR #73, opened 2023-02-15
Run dev publish on push

saskliutas:
Run dev prerelease publishing workflow on each push to default branch if changes were made to any of publishable packages. Added `[skip actions]` to version bump commit message to avoid recursively running this workflow.


grigasp (approved):
Nice, didn't know you can control action runs like that.

PR #80, opened 2023-02-22
Add missing css imports

saskliutas:
Added missing css imports in some `PresentationInstanceFilterBuilder` related components. Missing imports resulted in components being rendered without styling in cases when css classes used in those components where nor loaded globally.

Also updated beachball config. Looks like message property from config is used for both `publish` and `change` commands. This resulted in `pnpm change` script generating change files with same message.

### DIFF
--- a/.github/workflows/release-dev.yml
+++ b/.github/workflows/release-dev.yml
@@ -40,6 +40,6 @@ jobs:
       run: |
         git config --local user.email imodeljs-admin@users.noreply.github.com
         git config --local user.name imodeljs-admin
-        pnpm publish-packages-dev -y --branch ${{ github.ref_name }}
+        pnpm publish-packages-dev -y --branch ${{ github.ref_name }} --message "Version bump [skip actions]"
       env:
         NODE_AUTH_TOKEN: ${{ secrets.NPMJS_PUBLISH_ITWIN }}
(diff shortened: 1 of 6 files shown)


grigasp (approved):

mayank99:
This PR should be undone. Manually importing unlayered CSS from `@itwin/itwinui-css` causes styling issues downstream. And as of `@itwin/itwinui-react@2.12.x`, the entire iTwinUI stylesheet is automatically injected by default anyway, negating the need for such imports. [More details](https://github.com/iTwin/iTwinUI/wiki/Styling).

To tackle the root cause, instead of using classes from `@itwin/itwinui-css`, it is recommended to directly use components (`<Select>`, `<Input>`, `<Tag>`, etc) from `@itwin/itwinui-react`. If you find that some components do not fully meet your needs, you can open an issue in the [iTwinUI repo](https://github.com/iTwin/iTwinUI/issues).

saskliutas:
> This PR should be undone. Manually importing unlayered CSS from `@itwin/itwinui-css` causes styling issues downstream. And as of `@itwin/itwinui-react@2.12.x`, the entire iTwinUI stylesheet is automatically injected by default anyway, negating the need for such imports. [More details](https://github.com/iTwin/iTwinUI/wiki/Styling).
> 
> To tackle the root cause, instead of using classes from `@itwin/itwinui-css`, it is recommended to directly use components (`<Select>`, `<Input>`, `<Tag>`, etc) from `@itwin/itwinui-react`. If you find that some components do not fully meet your needs, you can open an issue in the [iTwinUI repo](https://github.com/iTwin/iTwinUI/issues).

Created an issue for this https://github.com/iTwin/presentation/issues/287. I will switch to `itwinui-react` components where possible.

PR #81, opened 2023-02-24
Make "Accept" button to be enabled when opening the filter dialog with pre-filled values

simasjar:
#77


packages/components/src/presentation-components/common/ContentBuilder.ts
```diff
@@ -53,6 +53,13 @@ export function createPropertyDescriptionFromFieldInfo(info: FieldInfo) {
     displayLabel: info.label,
   };
 
+  if ((info.type.typeName === StandardTypeNames.Number
+    || info.type.typeName === StandardTypeNames.Int
+    || info.type.typeName === StandardTypeNames.Float
+    || info.type.typeName === StandardTypeNames.Double)) {
```
saskliutas, line 59:
```suggestion
  if (descr.typeName === StandardTypeNames.Number
    || descr.typeName === StandardTypeNames.Int
    || descr.typeName === StandardTypeNames.Float
    || descr.typeName === StandardTypeNames.Double) {
```

packages/components/src/test/common/ContentBuilder.test.ts
```diff
@@ -127,4 +127,11 @@ describe("PropertyRecordsBuilder", () => {
     });
   });
 
+  it("Sets editor name when field info types typeName is Number", () => {
```
grigasp, line 130: the convention is to start test names in lowercase
```suggestion
  it("sets editor name when field info types typeName is Number", () => {
```

change/@itwin-presentation-components-3cbd3de2-872c-4136-81cc-54e6de289025.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Make \"Accept\" button to be enabled when opening the dialog with pre-filled values",
```
grigasp, line 3: This comment is going to appear in our consumer-facing changelog, so it should be as clear as possible. At the moment it would be impossible to tell what "Accept" button the comment refers to - I suggest prefixing the comment with `InstanceFilterDialog:`.

Also, I believe the change to set the editor to `StandardEditorNames.NumericInput` for all numeric properties might be a big deal to consumers. We should add another changelog entry for that. Looks like beachball doesn't support multiple change entries per file, so we might need a separate change file for it.

grigasp (changes requested):

change/@itwin-presentation-components-3cbd3de2-872c-4136-81cc-54e6de289025.json
simasjar: fixed

change/@itwin-presentation-components-3cbd3de2-872c-4136-81cc-54e6de289025.json
```diff
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Make \"Accept\" button to be enabled when opening the dialog with pre-filled values",
+  "comment": "InstanceFilterDialog: make \"Accept\" button to be enabled when opening the dialog with pre-filled values",
```
grigasp, line 3:
```suggestion
  "comment": "`InstanceFilterDialog`: Ensure \"Accept\" button is enabled when opening the dialog with pre-filled values",
```

change/@itwin-presentation-components-369ffc02-10c5-41c6-b119-e7c83364c597.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "The editor has changed for numeric types to StandardEditorNames.NumericInput",
```
grigasp, line 3:
```suggestion
  "comment": "Set editor to `StandardEditorNames.NumericInput` for all numeric properties",
```

grigasp (approved):

PR #83, opened 2023-02-27
Add white background to the chevron

### LINKED ISSUE #79: Fix navigation property value displayed below the drop down chevron (in z-index)

simasjar:
Navigation property value is displayed under the chevron when it's too long:
image

After this fix, property value would look like this:
image

closes #79

### DIFF
--- a/packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.tsx
+++ b/packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.tsx
@@ -159,7 +159,7 @@ function TargetSelectIndicatorsContainer<TOption extends OptionTypeBase>({ child
 }
 
 function TargetSelectDropdownIndicator<TOption extends OptionTypeBase>(_: IndicatorProps<TOption>) {
-  return <span className="iui-end-icon iui-actionable" >
+  return <span className="iui-end-icon iui-actionable" style={{ backgroundColor: "white", height: "22px" }} >
     <SvgCaretDownSmall />
   </span>;
 }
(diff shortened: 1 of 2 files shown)


packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.tsx
```diff
@@ -159,7 +159,7 @@ function TargetSelectIndicatorsContainer<TOption extends OptionTypeBase>({ child
 }
 
 function TargetSelectDropdownIndicator<TOption extends OptionTypeBase>(_: IndicatorProps<TOption>) {
-  return <span className="iui-end-icon iui-actionable" >
```
saskliutas, line 162: We probably need to use color variable that is used in [`iui-select` class](https://github.com/iTwin/iTwinUI/blob/bfe66dd80d26c119b7f605a7d57a83b207c12928/packages/itwinui-css/src/select/select.scss#L22) in order for this component to correctly work in different themes.

saskliutas (changes requested):

packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.tsx
simasjar: fixed

saskliutas: Also react-select component used here has [`styles`](https://react-select.com/styles#the-styles-prop) property that allows to set css styles on various inner components. I think it should be used for this `dropdownIndicator` too.

saskliutas (approved):

Issue #86, opened 2023-03-06
Add walkthrough documentation for setting up the four unified selection components


grigasp:
Our unified selection documentation page currently doesn't have any examples, just links a few sandboxes which are already outdated. Instead, we should have walkthrough docs for each component. The docs could also be referenced from `presentation-components` as examples for how to set up each component.

PR #102, opened 2023-03-17
Unified selection tests & snippets

grigasp:
Part of https://github.com/iTwin/presentation/issues/86

Markdowns: https://github.com/iTwin/itwinjs-core/pull/5228


.vscode/launch.json
```diff
@@ -74,6 +74,44 @@
...
+      ],
+      "outFiles": [
+        "${workspaceFolder}/packages/testing/lib/**/*.js"
+      ],
+      "env": {
+        "NODE_ENV": "development"
+      }
+    },
+    {
+      "name": "Tests: Testing",
```
saskliutas, line 102: Looks like it's duplicate

package.json
```diff
@@ -19,7 +19,7 @@
     "lint": "lage lint --continue",
     "lint:copyright": "node scripts/copyrightLinter.js",
     "prepare": "husky install",
-    "change": "beachball change",
+    "change": "beachball change --no-commit",
```
saskliutas, line 22: Is this intentional?

packages/full-stack-tests/src/components/unified-selection/PropertyGrid.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+      after(async () => {
+        await terminate();
+      });
+
+      it("renders unified selection property grid", async function () {
+        // __PUBLISH_EXTRACT_START__ Presentation.Components.UnifiedSelection.PropertyGrid
+        function MyPropertyGrid(props: { imodel: IModelConnection }) {
+          // create a presentation rules driven data provider; the provider implements `IDisposable`, so we
+          // create it through `useDisposable` hook to make sure it's properly cleaned up
+          const dataProvider = useDisposable(React.useCallback(() => {
```
saskliutas, line 41: There is no requirement to `import React from "react"` for JSX to work so I would prefer importing only react types and function that are used. ex: `import { useCallback } from "react"`

packages/full-stack-tests/src/components/unified-selection/PropertyGrid.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+          }, [props.imodel]));
+
+          // set up the data provider to be notified about changes in unified selection
+          const { isOverLimit, numSelectedElements } = usePropertyDataProviderWithUnifiedSelection({ dataProvider });
+
+          // width and height should generally we computed using ResizeObserver API or one of its derivatives
+          const [width] = React.useState(400);
+          const [height] = React.useState(600);
+
+          // data provider is going to be empty is no elements are selected
```
saskliutas, line 52:
```suggestion
          // data provider is going to be empty if no elements are selected
```

packages/full-stack-tests/src/components/unified-selection/PropertyGrid.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+  // find & expand the root category
+  const category = await waitFor(() => getRootPropertyCategory(container));
+  if (!category.querySelector(".iui-expanded"))
+    fireEvent.click(category.querySelector(".iui-expandable-block .iui-header")!);
+  await waitFor(() => expect(category.querySelector(".iui-expanded")).to.not.be.null);
+
+  // find the property label
+  await waitFor(() => {
+    const record = getPropertyRecordByLabel(container, propertyLabel);
+    expect(record.querySelector(`.components-property-record-value [title="${propertyValue}"]`)).to.not.be.null;
```
saskliutas, line 112: Is it really necessary to check that `.components-property-record-value` with title `propertyValue` is rendered under same parent as `propertyLabel`? Wouldn't it be enought to check that `propertyLabel` and `propertyValue` are rendered with `getByText`?

saskliutas (changes requested):

package.json
grigasp: Yes, kind of hate the idea of something automatically commiting for me. Also, not commiting is similar to what we have with `rush` in `itwinjs-core` and `appui`. Any objections?

packages/full-stack-tests/src/components/unified-selection/PropertyGrid.test.tsx
grigasp: agree, simplified

packages/full-stack-tests/src/components/unified-selection/PropertyGrid.test.tsx
saskliutas: I thinks there is no need to show that it's a `React` function as our components package is made specifically for `React` so I hoper package consumer knows about hooks used in this example. 

I want to have consistent imports inside whole repository, but if you think it gives some value to the example we can leave it.

package.json
saskliutas: I got used to it commiting automatically for me. 😄 I especially like it in case I forget to run change and CI catches it. As in that case I just need to run `pnpm change` and `git push`.

You can run `pnpm change --no-commit` in those cases when you don't want it to auto commit.

packages/full-stack-tests/src/components/unified-selection/PropertyGrid.test.tsx
grigasp: > I want to have consistent imports inside whole repository

makes sense, updated

package.json
grigasp: ok, I think I can get used to it

saskliutas (approved):

Issue #108, opened 2023-03-27
Fix `presentation-components` not transpiling with `itwinjs-core@3.6`


grigasp:
https://github.com/iTwin/presentation/actions/runs/4521812390/jobs/7963769742

```
Run pnpm lage build cover
info Lage task runner - let's make it
info @test-app/common build ▶️ start 
info @test-app/common build ✔️ done  - 3.27s
info @itwin/presentation-components build ▶️ start 
info @itwin/presentation-components build ❌ fail 
info 🏗 Summary
info 
info [Tasks Count] success: 1, skipped: 0, incomplete: 1
info ----------------------------------------------
ERR! [@itwin/presentation-components build] ERROR DETECTED
ERR! started
ERR! hash: 55432b3[6](https://github.com/iTwin/presentation/actions/runs/4521812390/jobs/7963769742#step:8:7)2e224a628da46fd3[7](https://github.com/iTwin/presentation/actions/runs/4521812390/jobs/7963769742#step:8:8)27a7249e6e4677[8](https://github.com/iTwin/presentation/actions/runs/4521812390/jobs/7963769742#step:8:9), cache hit? undefined
ERR! Running /opt/hostedtoolcache/node/18.15.0/x64/bin/npm run build
ERR! 
ERR! > @itwin/presentation-components@4.0.0-dev.27 build
ERR! > npm run -s copy:locale && npm run -s build:cjs && npm run -s build:esm
ERR! 
Error: ERR! src/presentation-components/tree/DataProvider.ts(25[9](https://github.com/iTwin/presentation/actions/runs/4521812390/jobs/7963769742#step:8:10),33): error TS2339: Property 'ResultSetTooLarge' does not exist on type 'typeof PresentationStatus'.
Error: ERR! src/test/tree/DataProvider.test.ts(3[10](https://github.com/iTwin/presentation/actions/runs/4521812390/jobs/7963769742#step:8:11),[11](https://github.com/iTwin/presentation/actions/runs/4521812390/jobs/7963769742#step:8:12)5): error TS2345: Argument of type '{ imodel: IModelConnection; rulesetOrId: string; parentKey: ECInstancesNodeKey; sizeLimit: number; }' is not assignable to parameter of type 'Paged<HierarchyRequestOptions<IModelConnection, NodeKey, RulesetVariable>>'.
[...]

PR #109, opened 2023-03-28
Fix transpilation with `itwinjs-core@3.6`

grigasp:
Fixes https://github.com/iTwin/presentation/issues/108


saskliutas (approved):

apps/test-app/frontend/src/index.tsx
```diff
@@ -23,26 +23,22 @@ Logger.initializeToConsole();
...
-  Object.assign(iModelAppOpts, {
     localization: new ITwinLocalization({
       initOptions: { lng: "en" },
     }),
-  });
+  };
 
   if (ProcessDetector.isElectronAppFrontend) {
-    // __PUBLISH_EXTRACT_START__ Presentation.Frontend.IModelAppStartup
     await ElectronApp.startup({ iModelApp: iModelAppOpts });
```
saskliutas, line 33: Looks like RpcInterfaces are not registered for electron app?

grigasp: yes... fixed

Issue #88, opened 2023-03-06
Error handling examples


grigasp:
With https://github.com/iTwin/itwinjs-core/pull/5168 we started to re-throw errors that occur when handling our requests. We need to either handle them at `presentation-components` level or add code examples for how to handle errors in consumer code.

PR #119, opened 2023-04-04
Add code snippets for error handling docs

grigasp:
Part of https://github.com/iTwin/presentation/issues/88


packages/full-stack-tests/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,40 @@
...
+      return this.state.error?.message ?? "Error";
+
+    // otherwise - render provided child component
+    return this.props.children;
+  }
+}
+// __PUBLISH_EXTRACT_END__
+
+export async function ensureHasError(container: HTMLElement, expectedError?: string) {
+  return waitFor(() => expect(getByText(container, expectedError ?? "Error")).to.not.be.undefined);
```
saskliutas, line 39: I think `getByText` throws if element if not found so there is no need to use `expect().to.not.be.undefined`

saskliutas (approved):

PR #120, opened 2023-04-04
Deprecate getNodeKey method on IPresentationTreeDataProvider

saskliutas:
Deprecated `getNodeKey` method in favor of getting node key directly from `PresentationTreeNodeItem` using `isPresentationTreeNodeItem` typeguard.


packages/components/src/presentation-components/tree/FilteredDataProvider.ts
```diff
@@ -142,6 +142,7 @@ export class FilteredPresentationTreeDataProvider implements IFilteredPresentati
   }
 
   public getNodeKey(node: TreeNodeItem): NodeKey {
+    // eslint-disable-next-line deprecation/deprecation
```
grigasp, line 145: why not switch to recommended way of doing this?

packages/components/src/presentation-components/tree/DataProvider.ts
```diff
@@ -183,10 +183,12 @@ export class PresentationTreeDataProvider implements IPresentationTreeDataProvid
 
   /**
    * Returns a [NodeKey]($presentation-common) from given [TreeNodeItem]($components-react).
-   * **Warning:** the `node` must be created by this data provider.
+   *
+   * **Warning**: Returns invalid [NodeKey]($presentation-common) if `node` is not a [[PresentationTreeNodeItem]].
```
grigasp, line 187: I think we should also add the `@deprecated` tag here, in addition to adding it on the interface. At the moment it's not clear the method is deprecated when looking at this piece of code.

grigasp (changes requested):

packages/components/src/presentation-components/tree/FilteredDataProvider.ts
saskliutas: This method is deprecated and will require that invalid key is returned if supplied node is not `PresentationTreeNodeItem` so I wanted to reuse implementation from `PresentationTreeDataProvider` which takes care of creating invalid `NodeKey`

grigasp (approved):

Issue #145, opened 2023-05-05
Add API release tags


grigasp:
Extract API from published packages to help us track public API and stay consistent between public / internal API.

PR #146, opened 2023-05-05
Extract API

grigasp:
Closes https://github.com/iTwin/presentation/issues/145

Prerequisites:
- [x] https://github.com/iTwin/itwinjs-core/pull/5486
- [x] https://github.com/planttheidea/micro-memoize/issues/97
   API Extractor is complaining about `micro-memoize` taking it's types from `.ts` rather than `.d.ts` with error:
   ```
   Error: {presentation-repo}/node_modules/.pnpm/micro-memoize@4.0.11/node_modules/micro-memoize/src/types.ts:1:1 - (ae-wrong-input-file-type) Incorrect file type; API Extractor expects to analyze compiler outputs with the .d.ts file extension. Troubleshooting tips: https://api-extractor.com/link/dts-error
   ```
   It seems there's no way to suppress the error from our side, so let's wait a bit to see what happens with the issue I filed. A couple of other possibilities is ignoring the error mentioned in api-extractor docs or using something other than `micro-memoize`. I don't like either of these two options.


saskliutas:
Looks like `pnpm-lock` is generated with `pnpm` 8.x. Should we bump pnpm version in root `package.json`?

grigasp:
> Looks like `pnpm-lock` is generated with `pnpm` 8.x. Should we bump `pnpm` version in root `package.json`?

Yes, let's do that separately after this PR completes

saskliutas (approved):

PR #156, opened 2023-05-30
Filter Dialog: Make dialog resizable, draggable, set larger min-width and min-height

simasjar:


packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
```diff
@@ -4,40 +4,43 @@
 *--------------------------------------------------------------------------------------------*/
 
 .presentation-instance-filter-dialog {
-  min-width: 50vw;
+  .presentation-instance-filter-dialog-content-container {
+    min-height: calc(var(--iui-size-2xl) * 5 + var( --iui-size-xs));
```
saskliutas, line 8: I think `min-height` can be simplified to `calc(var(--iui-size-2xl) * 6)`.

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
```diff
@@ -4,40 +4,43 @@
...
+    .presentation-instance-filter-title {
+      margin-bottom: 0;
+      padding: 6px;
+    }
 
-  .presentation-instance-filter-content {
-    max-height: 80vh;
-    overflow: hidden;
+    .presentation-instance-filter-content {
+      max-height: 80vh;
```
saskliutas, line 17: Can you check if this is still needed? If I remember correctly this was added to make sure that Dialog does not grow out off screen when adding new filter rules. However, Dialog is resizable now and it has static size.

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
```diff
@@ -4,40 +4,43 @@
...
+      overflow: hidden;
 
-    .presentation-property-filter-builder {
-      overflow: scroll;
-      max-height: 70vh;
+      .presentation-property-filter-builder {
+        overflow: scroll;
+        height: calc(100% - var(--iui-size-xl));
```
saskliutas, line 22: Looks like this does not divide height equally between `Class selector` and `Filter builder`. It seems that `Class selector` is slightly bigger than `--iui-size-xl`. 

I played with it locally and it looks okay with `Class selector: height: var(--iui-size-2xl), padding-top: 10px` (moved padding from bottom to the top). 
With this style applied `FilterBuilder` can use `height: calc(100% - var(--iui-size-2xl)`. 
What do you think?

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
```diff
@@ -4,40 +4,43 @@
...
-    overflow: hidden;
+    .presentation-instance-filter-content {
+      max-height: 80vh;
+      overflow: hidden;
 
-    .presentation-property-filter-builder {
-      overflow: scroll;
-      max-height: 70vh;
+      .presentation-property-filter-builder {
+        overflow: scroll;
```
saskliutas, line 21: Looks like vertical scrollbar is shown sometimes when there is no need for it (probably it takes into account possible size of horizontal scrollbar). I think this can be set to `auto`

saskliutas (changes requested):

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
simasjar: No, the instance filter dialog would be too big then.

simasjar: changed to other calculation

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
simasjar: removed

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
simasjar: fixed

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
simasjar: fixed

saskliutas (approved):

Issue #158, opened 2023-06-05
Tag documentation builds for publishing


grigasp:
Our documentation builds that we want to be published should be tagged with a specific tag `hasDocs` (see [here](https://github.com/iTwin/auth-clients/blob/17092d6c90899b702c1433e4038ae022157d995f/common/config/azure-pipelines/templates/build-docs.yaml#L67)) to tell the doc site pipeline which ones are ready for release.

PR #159, opened 2023-06-05
Tag docs build after release

saskliutas:
Closes https://github.com/iTwin/presentation/issues/158

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,6 @@ jobs:
         run: |
           git config --local user.email imodeljs-admin@users.noreply.github.com
           git config --local user.name imodeljs-admin
-          pnpm publish-packages -y --branch ${{ github.ref_name }} --message "Version bump [skip actions]"
+          pnpm publish-packages -y --branch ${{ github.ref_name }} --message "Version bump [skip actions][Publish Docs]"
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPMJS_PUBLISH_ITWIN }}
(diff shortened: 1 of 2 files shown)


ben-polinsky:
Looks good, thanks!

.github/workflows/release.yml
```diff
@@ -35,6 +35,6 @@ jobs:
         run: |
           git config --local user.email imodeljs-admin@users.noreply.github.com
           git config --local user.name imodeljs-admin
-          pnpm publish-packages -y --branch ${{ github.ref_name }} --message "Version bump [skip actions]"
+          pnpm publish-packages -y --branch ${{ github.ref_name }} --message "Version bump [skip actions][Publish Docs]"
```
grigasp, line 38: just for consistency I'd prefer `[publish docs]` instead of `[Publish Docs]

grigasp:
In case we need to patch an older release, e.g. let's say we're on `4.1` and want to release `4.0.1`, it looks like we're going to append `[publish docs]` in both cases, which will cause the tag to be added. Should we also check if the release is made from `master`?

saskliutas:
> In case we need to patch an older release, e.g. let's say we're on `4.1` and want to release `4.0.1`, it looks like we're going to append `[publish docs]` in both cases, which will cause the tag to be added. Should we also check if the release is made from `master`?

Docs pipeline should be triggered only by push to master.

.github/workflows/release.yml
saskliutas: Fixed

PR #160, opened 2023-06-13
Add loading spinner when filtering is in progress

simasjar:
Now when filtering is in progress filtering node looks like this: 
image


packages/components/src/presentation-components/tree/controlled/UseHierarchyLevelFiltering.ts
```diff
@@ -63,5 +75,5 @@ function applyHierarchyLevelFilter(nodeLoader: ITreeNodeLoader, modelSource: Tre
   if (updatedNode === undefined || !updatedNode.isExpanded || updatedNode.numChildren !== undefined) {
     return;
   }
-  nodeLoader.loadNode(updatedNode, 0).subscribe();
+  return nodeLoader.loadNode(updatedNode, 0).subscribe({ next: () => onComplete(nodeId) });
```
grigasp, line 78: should we use `complete` and `error` rather than `next`?

grigasp (commented):
The spinner and node label in the screenshot don't seem to be vertically centered

saskliutas (changes requested):

packages/components/src/test/tree/controlled/UseHierarchyLevelFiltering.test.ts
```diff
@@ -208,4 +209,26 @@ describe("useHierarchyLevelFiltering", () => {
     result.current.clearFilter(parentNode.item);
     expect(modelSource.getModel().getNode(childNode.id)).to.be.undefined;
   });
+
+  it("clears filter when filtering is in progress", async () => {
```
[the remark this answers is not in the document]
simasjar: fixed

packages/components/src/test/tree/controlled/UseHierarchyLevelFiltering.test.ts
```diff
@@ -208,4 +209,41 @@ describe("useHierarchyLevelFiltering", () => {
...
+      model.setChildren(undefined, [node], 0);
+    });
+
+    const { result } = renderHook(useHierarchyLevelFiltering, { initialProps: { modelSource, nodeLoader: nodeLoaderMock.object } });
+
+    result.current.applyFilter(node.item, filterInfo);
+    isFilterCalled = true;
+    result.current.clearFilter(node.item);
+
+    expect(applyFilterActionSubject.observers.length).to.be.eq(0);
```
saskliutas, line 230: I think there should be an expect statement after `applyFilter` to check if there are any observers.

packages/components/src/test/tree/controlled/UseHierarchyLevelFiltering.test.ts
```diff
@@ -208,4 +209,41 @@ describe("useHierarchyLevelFiltering", () => {
     result.current.clearFilter(parentNode.item);
     expect(modelSource.getModel().getNode(childNode.id)).to.be.undefined;
   });
+
+  it("clears filter unsubscribes from observable created by `applyFilter`", () => {
+    const applyFilterActionSubject = new Subject<TreeNodeLoadResult>();
+    const clearFilterActionSubject = new Subject<TreeNodeLoadResult>();
+    let isFilterCalled = false;
+    nodeLoaderMock.setup((x) => x.loadNode(moq.It.isAny(), 0)).returns(() => isFilterCalled ? clearFilterActionSubject : applyFilterActionSubject);
```
saskliutas, line 217: I think it should be possible to remove `isFilterCalled` variable:

```suggestion
    nodeLoaderMock.setup((x) => x.loadNode(moq.It.isAny(), 0)).returns(() => applyFilterActionSubject);
    nodeLoaderMock.setup((x) => x.loadNode(moq.It.isAny(), 0)).returns(() => clearFilterActionSubject);
```

saskliutas (changes requested):

packages/components/src/presentation-components/tree/controlled/UseHierarchyLevelFiltering.ts
grigasp: I believe we also need to call `onComplete(nodeId)` in case of an error

simasjar: fixed

packages/components/src/presentation-components/tree/controlled/UseHierarchyLevelFiltering.ts
```diff
@@ -27,25 +27,29 @@ export interface UseHierarchyLevelFilteringProps {
  */
 export function useHierarchyLevelFiltering(props: UseHierarchyLevelFilteringProps) {
   const { nodeLoader, modelSource } = props;
+  const ongoingSubscriptions = useRef(new Map<string, Subscription>());
 
-  const applyFilter = useCallback(
-    (node: TreeNodeItem, info: PresentationInstanceFilterInfo) => {
-      applyHierarchyLevelFilter(nodeLoader, modelSource, node.id, info);
+  const handleFilterAction = useCallback(
```
saskliutas, line 32: This `useCallback` is now useless because this function now returns new `applyFilter` and `clearFilter` functions. Either wrap them in `useCallback` or remove this `useCallback` usage.

saskliutas (changes requested):

packages/components/src/presentation-components/tree/controlled/UseHierarchyLevelFiltering.ts
simasjar: removed

saskliutas (approved):

grigasp:
The regression test failure should get resolved with https://github.com/iTwin/itwinjs-core/pull/5637

Issue #6, opened 2022-11-21
Proof of concept for stateless presentation-driven tree data provider


grigasp:

grigasp:
The load test results have multiple metrics, most important of which, I believe are: `itwin.nodes_request.response_time` and `vusers.session_length`. The stateless implementation makes two type of requests when creating nodes: "query rows" and "schema json", so it's also interesting to see how long they take (represented by `itwin.query_rows.response_time` and `itwin.schema_json.response_time` metrics).

Nodes response time (`itwin.nodes_request.response_time`) shows how long a user has to wait to get the child nodes for a parent. In case of the stateless implementation this is different from just making a request, because creating nodes may involve making more than 1 request and additional post-processing (sorting, grouping, etc.). In case of the the native implementation, everything's done on the backend, so the timings for making the request nearly match the timings of creating the nodes.

Virtual user session length (`vusers.session_length`) shows how long it takes a user to load all the hierarchy.

Ideally, we want those metrics to be affected by adding additional virtual users as little as possible.

grigasp:
## Best case scenario

In this situation there's one backend process (with default config) handling all requests of a single user. This should provide a good sense of the best possible performance we can expect from each implementation.

| Metric                             | Stateless                            | Native (cold cache)               | Native (warm cache)                  |
| ---------------------------------- | ------------------------------------ | --------------------------------- | ------------------------------------ |
| itwin.nodes_request.response_time: |                                      |                                   |                                      |
| ...min                             | 0                                    | 50                                | 23                                   |
| ...max                             | 802                                  | 10809                             | 4609                                 |
| ...median                          | 1                                    | 7709.8                            | 2671                                 |
| ...p95                             | 713.5                                | 10201.2                           | 4583.6                               |
| ...p99                             | 788.5                                | 10617.5                           | 4583.6                               |
| http.requests                      | 911                                  | 2018                              | 2018                                 |
| vusers.session_length:             |                                      |                                   |                                      |
| ...min                             | 1900.6                               | 47225.8                           | 16710.9                              |
| ...max                             | 1900.6                               | 47225.8                           | 16710.9                              |
| ...median                          | 1901.1                               | 47586.7                           | 16819.2                              |
| ...p95                             | 1901.1                               | 47586.7                           | 16819.2                              |
| ...p99                             | 1901.1                               | 47586.7                           | 16819.2                              |
| stats:                             |                                      |                                   |                                      |
[...]

PR #172, opened 2023-07-12
Stateless hierarchy builder

grigasp:
Closes https://github.com/iTwin/presentation/issues/6.

The PR contains:
- The proof of concept for a stateless Models Tree hierarchy builder.
- Substantial load tests' rewrite for testing native and stateless implementations.
- Some dependency updates


packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
```diff
@@ -0,0 +1,515 @@
...
+
+    const directChildren = this.ensureDirectChildren(node);
+    return directChildren
+      .pipe(
+        tap((n) => enableLogging && console.log(`HasNodes: partial node before HideIfNoChildrenReducer: ${JSON.stringify(n)}`)),
+        createHideIfNoChildrenReducer((n) => this.hasNodes(n), true),
+        tap((n) => enableLogging && console.log(`HasNodes: partial node before HideNodesInHierarchyReducer: ${JSON.stringify(n)}`)),
+        createHideNodesInHierarchyReducer(this, this._directNodesCache, true),
+      )
+      .pipe(
```
saskliutas, line 202: Why separate `pipe` calls?

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
```diff
@@ -0,0 +1,515 @@
...
+          },
+          complete() {
+            resolve(nodes);
+          },
+        });
+    });
+  }
+
+  public hasNodes(node: TreeNode): Observable<boolean> {
+    const enableLogging = false;
```
saskliutas, line 189: I see `enableLogging` in majority of functions in this file. Does it usually worth logging just in subset of them? If not then maybe there could be single variable at file level?

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
```diff
@@ -0,0 +1,515 @@
...
+            }),
+          ),
+        ),
+      ),
+    ).pipe(tap((node) => enableLogging && console.log(`DetermineChildrenReducer partial: ${node.label}: ${node.children}`)));
+  };
+}
+
+function createHideNodesInHierarchyReducer(
+  provider: TreeNodesProvider,
```
saskliutas, line 257: Maybe this function could take a callback instead of `provider`? Same as `createHideIfNoChildrenReducer`.

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
```diff
@@ -0,0 +1,515 @@
...
+          // eslint-disable-next-line no-console
+          console.log(`node ${parentNode.label} not extensible for setting children`);
+        }
+      }),
+    );
+    return nodes;
+  };
+}
+
+function createDetermineChildrenReducer(provider: TreeNodesProvider) {
```
saskliutas, line 236: Maybe this function could take a callback instead of `provider`? Same as `createHideIfNoChildrenReducer`

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
```diff
@@ -0,0 +1,515 @@
...
+      (queryObservable) =>
+        this._scheduler.scheduleSubscription(
+          queryObservable.pipe(
+            tap(() => enableLogging && console.log(`[loadDirectNodes] Do query for ${parentNode ? JSON.stringify(parentNode) : "<root>"}`)),
+            mergeMap((query) => from(this._queryReader.read(this._queryExecutor, { ...query, ecsql: applyLimit(query.ecsql, query.ctes) }))),
+          ),
+        ),
+      mergeAll(),
+      map(convertECInstanceIdSuffixToBase36),
+      shareReplay({ refCount: false }),
```
saskliutas, line 131: Looks like `refCount: false` is default value https://rxjs.dev/api/index/function/shareReplay#reference-counting. Maybe this can be removed

saskliutas (approved):
Did you start with `rxjs` v6 or used v7 from beginning? Do you know if there are any significant benefits in v7 and we should  bump version in `@itwin/components-react` too?

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
grigasp: Having it at file level makes it very hard to debug one specific reducer due to all the noise coming from other reducers.

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
grigasp: Cleaned up

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
grigasp: Good point, changed.

packages/hierarchy-builder/src/hierarchy-builder/TreeNodesProvider.ts
grigasp: Good point, changed.

grigasp:
> Did you start with rxjs v6 or used v7 from beginning? Do you know if there are any significant benefits in v7 and we should bump version in @itwin/components-react too?

I don't think there was a specific reason I chose 7 over 6 besides it being the latest version. [Some online material](https://medium.com/volosoft/whats-new-in-rxjs-7-a11cc564c6c0) suggests there're significant performance and memory footprint improvements - not sure that warrants an upgrade in `@itwin/components-react`.

PR #173, opened 2023-07-12
Upgrade @itwin/eslint-plugin to 4.0.0-dev.43 and eslint to 8.x accordingly

anmolshres98:
Change motivated from Nightman build failures owing to eslint-plugin version mismatch: https://github.com/iTwin/nightman/actions/runs/5514259302/jobs/10053301272

Specifically, newer eslint-plugin uses a [new config system](https://eslint.org/blog/2022/08/new-config-system-part-2/) and our packages need to be updated to use this new system


apps/load-tests/backend/eslint.config.js
```diff
@@ -0,0 +1,24 @@
...
+      jsx: true,
+      modules: true,
+    },
+  },
+};
+
+module.exports = [
+  ...eslintBaseConfig,
+  {
+    files: ["**/*.{ts,tsx}"],
```
grigasp, line 20: For backend configs we don't need JSX
```suggestion
      modules: true,
    },
  },
};

module.exports = [
  ...eslintBaseConfig,
  {
    files: ["**/*.ts"],
```

grigasp: This also applies to a number of other configs for packages that don't use React.

apps/test-app/frontend/eslint.config.js
```diff
@@ -0,0 +1,32 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+const eslintBaseConfig = require("../../../eslint.base.config");
+const jsxRuntimeConfig = require("eslint-plugin-react");
```
grigasp, line 3: `package.json` doesn't specify `eslint-plugin-react` as a dependency

packages/components/eslint.config.js
```diff
@@ -0,0 +1,32 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+const eslintBaseConfig = require("../../eslint.base.config");
+const jsxRuntimeConfig = require("eslint-plugin-react");
+
+
```
grigasp, line 5: extra new line

packages/components/eslint.config.js
```diff
@@ -0,0 +1,32 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+const eslintBaseConfig = require("../../eslint.base.config");
+const jsxRuntimeConfig = require("eslint-plugin-react");
```
grigasp, line 3: `package.json` doesn't have `eslint-plugin-react` as a dev dependency

packages/opentelemetry/eslint.config.js
```diff
@@ -0,0 +1,8 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
+
+module.exports = [
```
grigasp, line 3: I believe this config should be similar to other non-ui package configs (e.g. the `testing` one) and extend the base config. I see it wasn't doing that, but it should.

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
```diff
@@ -107,18 +107,18 @@ export function usePresentationInstanceFilteringProps(
...
-    | "properties"
-    | "classes"
-    | "selectedClasses"
-    | "onClassSelected"
-    | "onClassDeselected"
-    | "onClearClasses"
-    | "propertyRenderer"
-    | "onRulePropertySelected"
-    | "isDisabled"
+  InstanceFilterBuilderProps,
```
grigasp, line 110: Do we know why indent got changed here? I like the previous style better.

grigasp (changes requested):
I'm not sure I'm a fan of single-line member delimiter being a comma rather than semi-colon which we have currently. @saskliutas what do you think?

saskliutas:
> I'm not sure I'm a fan of single-line member delimiter being a comma rather than semi-colon which we have currently. @saskliutas what do you think?

I don't have strong preference on this but it looks like prettier uses ';'. It would be interesting to know why these formatting changes are made after upgrading eslint-plugin package. Is there some changes to rules in `itwinjs-recommended` config? Or previous configuration was not working correctly?

grigasp:
> I don't have strong preference on this but it looks like prettier uses ';'. It would be interesting to know why these formatting changes are made after upgrading eslint-plugin package. Is there some changes to rules in itwinjs-recommended config? Or previous configuration was not working correctly?

I assumed `eslint-plugin@dev.38` had a different config, but I just checked and the `@typescript-eslint/member-delimiter-style` rule config is the same in both `dev.38` and `dev.43`, so I'm not sure what caused that anymore.

saskliutas (changes requested):

apps/load-tests/backend/eslint.config.js
```diff
@@ -0,0 +1,24 @@
+const iTwinPlugin = require("@itwin/eslint-plugin");
```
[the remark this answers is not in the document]
anmolshres98: added, good catch. thanks for suggestion

apps/load-tests/backend/eslint.config.js
anmolshres98: removed for all relevant packages

apps/test-app/frontend/eslint.config.js
anmolshres98: added

packages/components/eslint.config.js
anmolshres98: removed

packages/components/eslint.config.js
anmolshres98: added

packages/opentelemetry/eslint.config.js
anmolshres98: done

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
anmolshres98: It is from `@typescript-eslint` that is being used by itwin/eslint-plugin

anmolshres98:
> > I don't have strong preference on this but it looks like prettier uses ';'. It would be interesting to know why these formatting changes are made after upgrading eslint-plugin package. Is there some changes to rules in itwinjs-recommended config? Or previous configuration was not working correctly?
> 
> I assumed `eslint-plugin@dev.38` had a different config, but I just checked and the `@typescript-eslint/member-delimiter-style` rule config is the same in both `dev.38` and `dev.43`, so I'm not sure what caused that anymore.

I am not completely certain about this either but you are right. It doesn't seem like `@typescript-eslint` dep changed in itwin/eslint-plugin between the two versions. My guess is either the linting wasn't working properly on this repo properly or something was messing up in itwin/eslint-plugin side. @Jake-Screen is out right now but he would be able to answer this better.

anmolshres98:
> > > I don't have strong preference on this but it looks like prettier uses ';'. It would be interesting to know why these formatting changes are made after upgrading eslint-plugin package. Is there some changes to rules in itwinjs-recommended config? Or previous configuration was not working correctly?
> > 
> > 
> > I assumed `eslint-plugin@dev.38` had a different config, but I just checked and the `@typescript-eslint/member-delimiter-style` rule config is the same in both `dev.38` and `dev.43`, so I'm not sure what caused that anymore.
> 
> I am not completely certain about this either but you are right. It doesn't seem like `@typescript-eslint` dep changed in itwin/eslint-plugin between the two versions. My guess is either the linting wasn't working properly on this repo properly or something was messing up in itwin/eslint-plugin side. @Jake-Screen is out right now but he would be able to answer this better.
> 
> There might be mismatch between prettier rules and eslint rules on this repo. That needs to be configured so that they match and don't conflict

After some investigation I have determined that this is because the rules from `eslint-config-prettier` are not being imported properly. Currently investigating this

anmolshres98:
> > > > I don't have strong preference on this but it looks like prettier uses ';'. It would be interesting to know why these formatting changes are made after upgrading eslint-plugin package. Is there some changes to rules in itwinjs-recommended config? Or previous configuration was not working correctly?
> > > 
> > > 
> > > I assumed `eslint-plugin@dev.38` had a different config, but I just checked and the `@typescript-eslint/member-delimiter-style` rule config is the same in both `dev.38` and `dev.43`, so I'm not sure what caused that anymore.
> > 
> > 
> > I am not completely certain about this either but you are right. It doesn't seem like `@typescript-eslint` dep changed in itwin/eslint-plugin between the two versions. My guess is either the linting wasn't working properly on this repo properly or something was messing up in itwin/eslint-plugin side. @Jake-Screen is out right now but he would be able to answer this better.
> > There might be mismatch between prettier rules and eslint rules on this repo. That needs to be configured so that they match and don't conflict
> 
> After some investigation I have determined that this is because the rules from `eslint-config-prettier` are not being imported properly. Currently investigating this

Update: figured out the fix, pushed, prettier should be working properly with eslint now. The only issue remaining is this one: 
image

I have set rule to warn for now:
image


Let me know how you want to proceed, thanks!

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
anmolshres98: This is from the prettier issue I have explained below, solved now with no change needed

grigasp:
> Update: figured out the fix, pushed, prettier should be working properly with eslint now. The only issue remaining is this one: image
> 
> I have set rule to warn for now: image
> 
> Let me know how you want to proceed, thanks!

I think that's a valid error, went ahead and fixed all occurrences of it.

grigasp (approved):

PR #176, opened 2023-07-19
Add unique values renderer

### LINKED ISSUE #175: Consider providing a dropdown of unique values rather than having to enter it

simasjar:
closes #175

### DIFF
--- a/packages/components/src/presentation-components/common/Utils.ts
+++ b/packages/components/src/presentation-components/common/Utils.ts
@@ -206,3 +206,16 @@ export function useErrorState() {
   }
   return setErrorState;
 }
+
+/**
+ * Helper function for parsing JSON value.
+ * If provided value is not a valid JSON value, function will return false.
+ * Otherwise function will return parsed JSON value.
+ */
+export function tryParseJSON(value: string) {
+  try {
+    return JSON.parse(value);
+  } catch {
+    return false;
+  }
+}
--- a/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
+++ b/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
@@ -43,7 +43,7 @@ export interface PresentationInstanceFilterBuilderProps {
   ruleGroupDepthLimit?: number;
   /** Initial filter that will be show when component is mounted. */
   initialFilter?: PresentationInstanceFilterInfo;
-  /** Should unique values rendeder be enabled */
+  /** Should unique values renderer be enabled */
   enableUniqueValuesRenderer?: boolean;
 }
 
--- a/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
+++ b/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
@@ -75,7 +75,7 @@ export function PresentationInstanceFilterBuilder(props: PresentationInstanceFil
         ruleGroupDepthLimit={ruleGroupDepthLimit}
         ruleValueRenderer={
           enableUniqueValuesRenderer
-            ? (props: PropertyFilterBuilderRuleValueProps) => <UniqueValuesRenderer {...props} imodel={imodel} descriptor={descriptor} />
+            ? (pr: PropertyFilterBuilderRuleValueProps) => <UniqueValuesRenderer {...pr} imodel={imodel} descriptor={descriptor} />
             : undefined
         }
       />
--- a/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
+++ b/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
@@ -7,13 +7,15 @@
  */
 
 import { useCallback, useState } from "react";
-import { PropertyFilter } from "@itwin/components-react";
+import { PropertyFilter, PropertyFilterBuilderRuleValue, PropertyFilterBuilderRuleValueProps, PropertyFilterRuleOperator } from "@itwin/components-react";
 import { IModelConnection } from "@itwin/core-frontend";
 import { ClassInfo, Descriptor } from "@itwin/presentation-common";
 import { navigationPropertyEditorContext } from "../properties/NavigationPropertyEditor";
+import { UniquePropertyValuesTargetSelector } from "../properties/UniquePropertyValuesTargetSelector";
 import { InstanceFilterBuilder, useFilterBuilderNavigationPropertyEditorContext, usePresentationInstanceFilteringProps } from "./InstanceFilterBuilder";
 import { PresentationInstanceFilter } from "./Types";
 import { convertPresentationFilterToPropertyFilter, createPresentationInstanceFilter } from "./Utils";
+import { StandardTypeNames } from "@itwin/appui-abstract";
 
 /**
  * Data structure that stores information about filter built by [[PresentationInstanceFilterBuilder]].
@@ -41,6 +43,8 @@ export interface PresentationInstanceFilterBuilderProps {
   ruleGroupDepthLimit?: number;
   /** Initial filter that will be show when component is mounted. */
   initialFilter?: PresentationInstanceFilterInfo;
+  /** Should unique values rendeder be enabled */
+  enableUniqueValuesRenderer?: boolean;
 }
 
 /**
@@ -48,7 +52,7 @@ export interface PresentationInstanceFilterBuilderProps {
  * @beta
  */
 export function PresentationInstanceFilterBuilder(props: PresentationInstanceFilterBuilderProps) {
-  const { imodel, descriptor, onInstanceFilterChanged, ruleGroupDepthLimit, initialFilter } = props;
+  const { imodel, descriptor, onInstanceFilterChanged, ruleGroupDepthLimit, initialFilter, enableUniqueValuesRenderer } = props;
   const filteringProps = usePresentationInstanceFilteringProps(descriptor, imodel, initialFilter?.usedClasses);
 
   const onFilterChanged = useCallback(
@@ -69,7 +73,25 @@ export function PresentationInstanceFilterBuilder(props: PresentationInstanceFil
         initialFilter={initialPropertyFilter}
         onFilterChanged={onFilterChanged}
         ruleGroupDepthLimit={ruleGroupDepthLimit}
+        ruleValueRenderer={
+          enableUniqueValuesRenderer
+            ? (props: PropertyFilterBuilderRuleValueProps) => <UniqueValuesRenderer {...props} imodel={imodel} descriptor={descriptor} />
+            : undefined
+        }
       />
     </navigationPropertyEditorContext.Provider>
+  );
+}
+
+function UniqueValuesRenderer(props: PropertyFilterBuilderRuleValueProps & { imodel: IModelConnection; descriptor: Descriptor }) {
+  return (
+    <>
+      {props.property.typename !== StandardTypeNames.Navigation &&
+      (props.operator === PropertyFilterRuleOperator.IsEqual || props.operator === PropertyFilterRuleOperator.IsNotEqual) ? (
+        <UniquePropertyValuesTargetSelector {...props} />
+      ) : (
+        <PropertyFilterBuilderRuleValue {...props} />
+      )}
+    </>
   );
 }
--- a/packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.scss
+++ b/packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.scss
@@ -9,4 +9,7 @@
     outline: none;
     padding: 0;
   }
+  > input::placeholder {
+    opacity: 1; // it is used to override mozilla firefox style for placeholder.
+  }
 }
--- a/packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
+++ b/packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
@@ -8,22 +8,13 @@ import { ActionMeta, MultiValue } from "react-select";
 import { PropertyDescription, PropertyValue, PropertyValueFormat } from "@itwin/appui-abstract";
 import { PropertyFilterBuilderRuleValueProps } from "@itwin/components-react";
 import { IModelConnection } from "@itwin/core-frontend";
-import {
-  ContentFlags,
-  ContentSpecificationTypes,
-  Descriptor,
-  DisplayValueGroup,
-  FieldDescriptor,
-  KeySet,
-  Ruleset,
-  RuleTypes,
-  Value,
-} from "@itwin/presentation-common";
+import { ContentSpecificationTypes, Descriptor, DisplayValueGroup, FieldDescriptor, KeySet, Ruleset, RuleTypes, Value } from "@itwin/presentation-common";
 import { Presentation } from "@itwin/presentation-frontend";
 import { AsyncMultiTagSelect } from "../instance-filter-builder/MultiTagSelect";
-import { NAVIGATION_PROPERTY_TARGETS_BATCH_SIZE } from "./UseNavigationPropertyTargetsLoader";
-import { tryParseJSON } from "../common/Utils";
+import { PRESENTATION_TARGETS_BATCH_SIZE } from "./UseNavigationPropertyTargetsLoader";
+import { translate } from "../common/Utils";
 
+/** @internal */
 export function UniquePropertyValuesTargetSelector(props: PropertyFilterBuilderRuleValueProps & { imodel: IModelConnection; descriptor: Descriptor }) {
   const { imodel, descriptor, property, onChange, value } = props;
   const [selectedTarget, setSelectedTarget] = useState<DisplayValueGroup[] | undefined>(() => getUniqueValueTargetFromProperty(value));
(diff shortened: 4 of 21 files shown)


packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { PropertyFilterBuilderRuleValueProps } from "@itwin/components-react";
+import { IModelConnection } from "@itwin/core-frontend";
+import { ContentSpecificationTypes, Descriptor, DisplayValueGroup, FieldDescriptor, KeySet, Ruleset, RuleTypes, Value } from "@itwin/presentation-common";
+import { Presentation } from "@itwin/presentation-frontend";
+import { AsyncMultiTagSelect } from "../instance-filter-builder/MultiTagSelect";
+import { PRESENTATION_TARGETS_BATCH_SIZE } from "./UseNavigationPropertyTargetsLoader";
+import { translate } from "../common/Utils";
+
+/** @internal */
+export function UniquePropertyValuesTargetSelector(props: PropertyFilterBuilderRuleValueProps & { imodel: IModelConnection; descriptor: Descriptor }) {
```
saskliutas, line 18: I don't think `target` in this this component makes sense. I would suggest removing `target` from functions and variables names.

```suggestion
export function UniquePropertyValuesSelector(props: PropertyFilterBuilderRuleValueProps & { imodel: IModelConnection; descriptor: Descriptor }) {
```

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const { imodel, descriptor, property, onChange, value } = props;
+  const [selectedTarget, setSelectedTarget] = useState<DisplayValueGroup[] | undefined>(() => getUniqueValueTargetFromProperty(value));
+  const [fieldDescriptor, setFieldDescriptor] = useState<FieldDescriptor | undefined>(() =>
```
saskliutas, line 21: Why using `FieldDescriptor` instead of `Field`?

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
```diff
@@ -0,0 +1,154 @@
...
+    descriptor.fields.find((field) => field.label === property.displayLabel)?.getFieldDescriptor(),
```
saskliutas, line 22: Field should be looked up be the name

```suggestion
    descriptor.fields.find((field) => field.name === property.name)?.getFieldDescriptor(),
```

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const ruleset = useUniquePropertyValuesRuleset({ property, fieldDescriptor });
+  const loadTargets = useUniquePropertyValuesLoader({ imodel, ruleset, fieldDescriptor });
+
+  return (
+    <AsyncMultiTagSelect
+      value={selectedTarget ?? null}
+      loadOptions={async (inputValue, options) => loadTargets(inputValue, options.length)}
+      placeholder={translate("unique-values-property-editor.select-values")}
+      onChange={onValueChange}
+      debounceTimeout={500}
```
saskliutas, line 62: I think this does not have an effect when search is disabled.

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
```diff
@@ -0,0 +1,154 @@
...
+    [onChange],
+  );
+
+  const ruleset = useUniquePropertyValuesRuleset({ property, fieldDescriptor });
+  const loadTargets = useUniquePropertyValuesLoader({ imodel, ruleset, fieldDescriptor });
+
+  return (
+    <AsyncMultiTagSelect
+      value={selectedTarget ?? null}
+      loadOptions={async (inputValue, options) => loadTargets(inputValue, options.length)}
```
saskliutas, line 59: You can remove first argument from `loadTargets` as it is not used.

packages/components/src/presentation-components/properties/UseNavigationPropertyTargetsLoader.ts
```diff
@@ -22,7 +22,7 @@ import {
 import { Presentation } from "@itwin/presentation-frontend";
 
 /** @internal */
-export const NAVIGATION_PROPERTY_TARGETS_BATCH_SIZE = 100;
+export const PRESENTATION_TARGETS_BATCH_SIZE = 100;
```
saskliutas, line 25: I would suggest having separate `BATCH_SIZE` constants for navigation property and unique values. Otherwise I would move this to some common place to avoid UniqueValuesSelector having to import from NavigationPropertySelector.

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  useEffect(() => {
+    setFieldDescriptor(descriptor.fields.find((field) => field.label === property.displayLabel)?.getFieldDescriptor());
+  }, [descriptor, property]);
+
+  useEffect(() => {
+    setSelectedTarget(getUniqueValueTargetFromProperty(value));
+  }, [value]);
+
+  const onValueChange = useCallback(
```
saskliutas, line 33: Is this `useCallback` is needed?

packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.scss
```diff
@@ -9,4 +9,7 @@
     outline: none;
     padding: 0;
   }
+  > input::placeholder {
```
saskliutas, line 12: Do we really need this? Can we stick to what browser defaults to?

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
```diff
@@ -69,7 +73,25 @@ export function PresentationInstanceFilterBuilder(props: PresentationInstanceFil
...
       />
     </navigationPropertyEditorContext.Provider>
   );
 }
+
+function UniqueValuesRenderer(props: PropertyFilterBuilderRuleValueProps & { imodel: IModelConnection; descriptor: Descriptor }) {
+  return (
+    <>
+      {props.property.typename !== StandardTypeNames.Navigation &&
+      (props.operator === PropertyFilterRuleOperator.IsEqual || props.operator === PropertyFilterRuleOperator.IsNotEqual) ? (
```
saskliutas, line 90: Looks like this can be refactored to `if` statement instead of ternary operator.

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
saskliutas, line 18: I don't think this component should take `PropertyFilterBuilderRuleValueProps`. I think this could be a generic select component for working with unique content values. I would suggest defining props interface with only props that are used in this component.

packages/components/src/test/properties/UniquePropertyValuesTargetSelector.test.tsx
```diff
@@ -0,0 +1,184 @@
...
+        imodel={testImodel}
+        descriptor={descriptor}
+        value={value}
+      />,
+    );
+
+    expect(queryByText(displayValue[0])).to.not.be.null;
+  });
+
+  it("sets multiple provided values", () => {
```
saskliutas, line 165:
```suggestion
  it("selects multiple provided values", () => {
```

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
```diff
@@ -0,0 +1,154 @@
...
+    [imodel, ruleset, fieldDescriptor],
+  );
+
+  return loadTargets;
+}
+
+function stringifyGroupRawValuesTo(values: MultiValue<DisplayValueGroup>) {
+  const singleArray: Value[] = [];
+  values.forEach((item) => singleArray.push(item.groupedRawValues));
+  return JSON.stringify(singleArray);
```
saskliutas, line 153: I see there are multiple places where values are serialized deserialized. Looks like it would be worth introducing some functions that would serialize/deserialize `MultiValue` and could be reused in all different places

saskliutas (changes requested):
Looks like in some files imports are not sorted correctly. Could you run `imports sorted`?

packages/components/src/presentation-components/properties/NavigationPropertyTargetSelector.scss
simasjar: yes, otherwise there can be UI inconsistencies, for example: 
image

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
simasjar: fixed

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
simasjar: fixed

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
simasjar: fixed

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
simasjar: fixed

packages/components/src/presentation-components/properties/UseNavigationPropertyTargetsLoader.ts
simasjar: fixed

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
simasjar: removed

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterBuilder.tsx
simasjar: fixed

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
simasjar: fixed? not sure if I could create a new interface and extend it from `PropertyFilterBuilderRuleValueProps`

packages/components/src/presentation-components/properties/UniquePropertyValuesTargetSelector.tsx
simasjar: fixed

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
```diff
@@ -0,0 +1,197 @@
...
+        hasMore: content.items.length === UNIQUE_PROPERTY_VALUES_BATCH_SIZE,
+      };
+    },
+    [imodel, ruleset, fieldDescriptor],
+  );
+
+  return loadTargets;
+}
+
+function serializeDisplayValueGroupArray(values: DisplayValueGroup[]) {
```
saskliutas, line 189: I would suggest moving this to `common/Utils.ts` so both `serialize` and `deserialize` function are at the same file.

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
```diff
@@ -0,0 +1,197 @@
...
+        descriptor: {},
+        fieldDescriptor,
+        rulesetOrId: ruleset,
+        paging: { start: loadedOptionsCount, size: UNIQUE_PROPERTY_VALUES_BATCH_SIZE },
+        keys: new KeySet(),
+      });
+      return {
+        options: content.items.filter(
+          (item) =>
+            item.groupedRawValues.filter((value) => value !== undefined && value !== null) && item.displayValue !== undefined && item.displayValue !== "",
```
saskliutas, line 178: I think checking `item.displayValue !== undefined && item.displayValue !== ""` should be enough.

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
```diff
@@ -0,0 +1,197 @@
...
+      hideSelectedOptions={false}
+      isSearchable={false}
+      closeMenuOnSelect={false}
+      getOptionLabel={(option) => option.displayValue?.toString() ?? ""}
+      getOptionValue={(option) => option.groupedRawValues[0]?.toString() ?? ""}
+    />
+  );
+}
+
+function getUniqueValueTargetFromProperty(property: PropertyValue | undefined): DisplayValueGroup[] | undefined {
```
saskliutas, line 95:
```suggestion
function getUniqueValueFromProperty(property: PropertyValue | undefined): DisplayValueGroup[] | undefined {
```

packages/components/src/presentation-components/common/Utils.ts
```diff
@@ -208,14 +208,25 @@ export function useErrorState() {
...
+    }
+  };
+  const displayValues = tryParseJSON(serializedDisplayValues);
+  const groupedRawValues = tryParseJSON(serializedGroupedRawValues);
+
+  if (displayValues === false || groupedRawValues === false) {
+    return { deserializedDisplayValues: undefined, deserializedGroupedRawValues: undefined };
   }
+
+  return { deserializedDisplayValues: displayValues as string[], deserializedGroupedRawValues: groupedRawValues as Value[][] };
```
saskliutas, line 231:
```suggestion
  return { displayValues: displayValues as string[], groupedRawValues: groupedRawValues as Value[][] };
```

packages/components/src/presentation-components/common/Utils.ts
```diff
@@ -208,14 +208,25 @@ export function useErrorState() {
...
+    try {
+      return JSON.parse(value);
+    } catch {
+      return false;
+    }
+  };
+  const displayValues = tryParseJSON(serializedDisplayValues);
+  const groupedRawValues = tryParseJSON(serializedGroupedRawValues);
+
+  if (displayValues === false || groupedRawValues === false) {
```
saskliutas, line 227:
```suggestion
  if (!displayValues || !groupedRawValues) {
```

packages/components/src/test/properties/UniquePropertyValuesTargetSelector.test.tsx
```diff
@@ -168,17 +146,36 @@ describe("UniquePropertyValuesTargetSelector", () => {
...
     const { queryByText } = render(
-      <UniquePropertyValuesTargetSelector
+      <UniquePropertyValuesSelector property={propertyDescription} onChange={() => {}} imodel={testImodel} descriptor={descriptor} value={value} />,
+    );
+
+    expect(queryByText(displayValue[0])).to.not.be.null;
+    expect(queryByText(displayValue[1])).to.not.be.null;
+  });
+
+  it("does not set value when provided value is not parsable", () => {
```
saskliutas:
```suggestion
  it("does not set value when provided value is invalid", () => {
```

saskliutas (changes requested):

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
simasjar: fixed

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
simasjar: fixed

packages/components/src/test/properties/UniquePropertyValuesTargetSelector.test.tsx
simasjar: for some reason can't commit suggestion, fixed

packages/components/src/presentation-components/common/Utils.ts
```diff
@@ -206,3 +215,41 @@ export function useErrorState() {
...
+ * Returns an object, which consists of `serializedDisplayValues` and `serializedGroupedRawValues`.
+ */
+export function serializeDisplayValueGroupArray(values: DisplayValueGroup[]) {
+  const displayValues: DisplayValue[] = [];
+  const groupedRawValues: Value[] = [];
+  values.forEach((item) => {
+    displayValues.push(item.displayValue);
+    groupedRawValues.push(item.groupedRawValues);
+  });
+  return { serializedDisplayValues: JSON.stringify(displayValues), serializedGroupedRawValues: JSON.stringify(groupedRawValues) };
```
saskliutas, line 230:
```suggestion
  return { displayValues: JSON.stringify(displayValues), groupedRawValues: JSON.stringify(groupedRawValues) };
```

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
```diff
@@ -0,0 +1,173 @@
...
+  /** Current IModel */
+  imodel: IModelConnection;
+  /** Current descriptor */
+  descriptor: Descriptor;
+}
+
+/** @internal */
+export function UniquePropertyValuesSelector(props: UniquePropertyValuesSelectorProps) {
+  const { imodel, descriptor, property, onChange, value } = props;
+  const [selectedTarget, setSelectedTarget] = useState<DisplayValueGroup[] | undefined>(() => getUniqueValueFromProperty(value));
```
saskliutas, line 37:
```suggestion
  const [selectedValues, setSelectedValues] = useState<DisplayValueGroup[] | undefined>(() => getUniqueValueFromProperty(value));
```

saskliutas (changes requested):

saskliutas (approved):

saskliutas:
Probably need to update CI pipeline to trigger on `next` branch instead of only on `master`.

PR #206, opened 2023-08-07
Fix filtering when rule contains `Like` operator

### LINKED ISSUE #205: Instance Filter Dialog: filter is not working with `Like` operator

simasjar:
closes #205

### DIFF
--- a/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
+++ b/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
@@ -76,10 +76,14 @@ function convertFilter(filter: PresentationInstanceFilter, ctx: ConvertContext) 
 }
 
 function convertUniqueValuesCondition(filter: PresentationInstanceFilterCondition, ctx: ConvertContext) {
+  // Unique values works only with `IsEqual` and `IsNotEqual` operators.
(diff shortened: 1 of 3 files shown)


packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
```diff
@@ -224,7 +224,11 @@ function handleStringifiedValues(filter: PresentationInstanceFilterCondition, se
   let selectedValueIndex = 0;
 
   const { displayValues, groupedRawValues } = deserializeDisplayValueGroupArray(serializedDisplayValues, serializedGroupedRawValues);
-  if (displayValues === undefined || groupedRawValues === undefined) {
+  if (
+    displayValues === undefined ||
+    groupedRawValues === undefined ||
+    (operator !== PropertyFilterRuleOperator.IsEqual && operator !== PropertyFilterRuleOperator.IsNotEqual)
```
saskliutas, line 230: I think operator should be checked in `convertUniqueValuesCondition` function before trying to deserialize values.

saskliutas (changes requested):

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
simasjar: fixed

simasjar: fixed

saskliutas (approved):

PR #208, opened 2023-08-08
Add filter expression for `Point` type

### LINKED ISSUE #198: Instance Filter Dialog: unique values editor is not working with point and navigation types

simasjar:
closes #198

### DIFF
--- a/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
+++ b/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
@@ -6,7 +6,7 @@
  * @module InstancesFilter
  */
 
-import { Primitives, PrimitiveValue } from "@itwin/appui-abstract";
+import { Primitives, PrimitiveValue, StandardTypeNames } from "@itwin/appui-abstract";
 import { isUnaryPropertyFilterOperator, PropertyFilterRuleGroupOperator, PropertyFilterRuleOperator } from "@itwin/components-react";
 import { IModelConnection } from "@itwin/core-frontend";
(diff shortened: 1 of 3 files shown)


packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
```diff
@@ -122,6 +122,11 @@ function createComparison(propertyName: string, type: string, alias: string, ope
     case "number":
       valueExpression = value.toString();
       break;
+    case "object":
+      if (isPoint2d(value)) {
+        return createPointComparision(value, operatorExpression, propertyAccessor);
+      }
+      break;
```
saskliutas, line 129: I think this should be handled outside `switch` statement like `navigation` and `dateTime` properties.

saskliutas (changes requested):

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
simasjar: fixed

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
```diff
@@ -124,6 +124,12 @@ function createComparison(propertyName: string, type: string, alias: string, ope
       break;
   }
 
+  if (type === StandardTypeNames.Point2d || type === StandardTypeNames.Point3d) {
+    // istanbul ignore else
+    if (typeof value === "object" && isPoint2d(value)) {
```
saskliutas, line 129:
```suggestion
  if (type === StandardTypeNames.Point2d || type === StandardTypeNames.Point3d) {
    assert(isPoint2d(value)
```

saskliutas (approved):

Issue #212, opened 2023-08-10
Commit links in changelogs are incorrect


grigasp:
1. Open https://github.com/iTwin/presentation/blob/master/packages/components/CHANGELOG.md?plain=1
2. Notice all commit links within a single release are the same and seem to point to the last commit at the point of release.

PR #214, opened 2023-08-17
Fix changelogs

saskliutas:
Closes https://github.com/iTwin/presentation/issues/212

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
         run: pnpm install -w
 
       - name: Run check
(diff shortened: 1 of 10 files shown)


grigasp:
Can you please describe what was causing the changelogs to point to the same commit? Not checkouting the branch with full history?

saskliutas:
> Can you please describe what was causing the changelogs to point to the same commit? Not checkouting the branch with full history?

Yes workflow was checking out only top commit and that was used for all changes when generating changelog.

grigasp (approved):

Issue #93, opened 2023-03-09
Update components when active unit system changes


grigasp:
Components that show formatted values should react to active unit system change and re-format the values with new unit system.

PR #218, opened 2023-08-22
Components should react to unit system change

saskliutas:
Closes https://github.com/iTwin/presentation/issues/93

Reload Tree, Table and PropertyGrid data when active unit system is changed to make sure values format matches the new active unit system.

### DIFF
--- a/packages/components/src/test/table/UseRows.test.tsx
+++ b/packages/components/src/test/table/UseRows.test.tsx
@@ -6,12 +6,13 @@
 import { expect } from "chai";
 import sinon from "sinon";
 import * as moq from "typemoq";
-import { IModelConnection } from "@itwin/core-frontend";
+import { BeUiEvent } from "@itwin/core-bentley";
+import { FormattingUnitSystemChangedArgs, IModelApp, IModelConnection, QuantityFormatter } from "@itwin/core-frontend";
 import { Content, DescriptorOverrides, KeySet, SortDirection } from "@itwin/presentation-common";
 import { Presentation, PresentationManager } from "@itwin/presentation-frontend";
 import { render, waitFor } from "@testing-library/react";
 import { renderHook } from "@testing-library/react-hooks";
-import { useRows, UseRowsProps } from "../../presentation-components/table/UseRows";
+import { ROWS_RELOAD_PAGE_SIZE, useRows, UseRowsProps } from "../../presentation-components/table/UseRows";
 import { createTestECInstanceKey, createTestPropertyInfo, TestErrorBoundary } from "../_helpers/Common";
 import {
   createTestCategoryDescription,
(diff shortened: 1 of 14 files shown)


packages/components/src/test/table/UseRows.test.tsx
```diff
@@ -242,4 +248,113 @@ describe("useRows", () => {
...
+    await waitFor(() => {
+      expect(result.current.rows).to.have.lengthOf(itemsCount);
+      presentationManagerMock.verify(
+        async (x) => x.getContentAndSize(moq.It.is(({ paging }) => paging?.start === 0 && paging?.size === ROWS_RELOAD_PAGE_SIZE)),
+        moq.Times.once(),
+      );
+      presentationManagerMock.verify(
+        async (x) => x.getContentAndSize(moq.It.is(({ paging }) => paging?.start === ROWS_RELOAD_PAGE_SIZE && paging?.size === 1)),
+        moq.Times.once(),
+      );
```
grigasp, line 357: maybe use `verifyable` at where you set up these mocks and then just call `presentationManagerMock.verifyAll()` here?

packages/testing/src/test/ContentBuilder.test.ts
```diff
@@ -242,6 +242,9 @@ describe("ContentBuilder", () => {
       presentationManagerMock.setup(async (manager) => manager.getContent(moq.It.isAny())).returns(getEmptyContent);
       presentationManagerMock.setup((x) => x.onIModelContentChanged).returns(() => new BeEvent());
       sinon.stub(Presentation, "presentation").get(() => presentationManagerMock.object);
+      sinon.stub(IModelApp, "quantityFormatter").get(() => ({
+        onActiveFormattingUnitSystemChanged: new BeUiEvent<FormattingUnitSystemChangedArgs>(),
```
grigasp, line 246: this suggests we're subscribing for the event somewhere in these tests - do you know where?

saskliutas: This is where we [create default ContentDataProvider](https://github.com/iTwin/presentation/blob/4416ca4f1fa7d0f6127c7e651ffd23a7460b69af/packages/testing/src/presentation-testing/ContentBuilder.ts#L98) and this is the [test that uses it](https://github.com/iTwin/presentation/blob/4416ca4f1fa7d0f6127c7e651ffd23a7460b69af/packages/testing/src/test/ContentBuilder.test.ts#L252)

grigasp: looks like a useless test that tests the provider rather than the builder - maybe just remove it?..

packages/components/src/test/table/UseRows.test.tsx
saskliutas: Changed to `verifyAll`

packages/testing/src/test/ContentBuilder.test.ts
saskliutas: Refactored that test into two to make sure that test coverage is still at 100%.

grigasp (approved):

PR #224, opened 2023-08-29
Add tests for Helpers.ts to reach 100% coverage

JonasDov:
Helpers.ts was not included into "npm run cover". Included it into "npm run cover", and added tests to reach 100% coverage.

### DIFF
--- a/packages/testing/src/test/Helpers.test.ts
+++ b/packages/testing/src/test/Helpers.test.ts
@@ -2,11 +2,13 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+
 import { expect } from "chai";
 import * as rimraf from "rimraf";
 import * as sinon from "sinon";
+import { IModelHost } from "@itwin/core-backend";
 import { Guid } from "@itwin/core-bentley";
-import { NoRenderApp } from "@itwin/core-frontend";
+import { IModelApp, NoRenderApp } from "@itwin/core-frontend";
 import { Presentation as PresentationBackend, PresentationProps } from "@itwin/presentation-backend";
 import { Presentation as PresentationFrontend } from "@itwin/presentation-frontend";
 import { HierarchyCacheMode, initialize, PresentationTestingInitProps, terminate } from "../presentation-testing/Helpers";
@@ -21,6 +23,9 @@ describe("Helpers", () => {
       frontendInitializationStub = sinon.stub(PresentationFrontend, "initialize");
       sinon.stub(PresentationBackend, "terminate");
       sinon.stub(PresentationFrontend, "terminate");
+      sinon.stub(IModelHost, "startup");
+      sinon.stub(NoRenderApp, "startup");
+      sinon.stub(IModelApp, "localization").get(() => ({ getLanguageList: () => ["en"] }));
     });
 
     afterEach(async () => {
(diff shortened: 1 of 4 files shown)


packages/testing/src/test/Helpers.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    afterEach(async () => {
+      await terminate();
+    });
+
+    it("initializes PresentationBackend and PresentationFrontend on initialize", async () => {
+      const backendInitializationSpy = sinon.spy(PresentationBackend, "initialize");
+      const frontendInitializationSpy = sinon.spy(PresentationFrontend, "initialize");
+
+      await initialize();
+      expect(backendInitializationSpy.calledOnce);
```
grigasp, line 26: Let's use [sinon-chai](https://www.chaijs.com/plugins/sinon-chai/) way of asserting these. E.g. instead of `expect(backendInitializationSpy.calledOnce)` do `expect(backendInitializationSpy).to.be.calledOnce`

packages/testing/src/test/Helpers.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    it("calls rimraf sync when initialized with HybridCacheConfig", async () => {
+      const syncStub = sinon.stub(rimraf, "sync");
+      const props: PresentationTestingInitProps = {
+        backendProps: { caching: { hierarchies: { mode: HierarchyCacheMode.Hybrid, disk: { mode: HierarchyCacheMode.Disk, directory: process.cwd() } } } },
+      };
+      await initialize(props);
+
+      await terminate();
+      expect(syncStub.calledOnce);
+    });
```
grigasp, line 101: Using `rimraf.sync` is an implementation detail of the goal to simply clear the cache directory. I suggest we rename the tests to describe the goal rather than this implementation detail, e.g. "clears cache directory when initialized with DiskHierarchyCacheConfig". 

Also, I don't think we need to pass a real (`process.cwd()`) directory here - simply `"/test/directory/"` should be sufficient. And we should assert that `rimraf.sync` is called with that directory.

packages/testing/src/test/Helpers.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+describe("Helpers", () => {
+  describe("initialize", () => {
+    afterEach(async () => {
+      await terminate();
+    });
+
+    it("initializes PresentationBackend and PresentationFrontend on initialize", async () => {
+      const backendInitializationSpy = sinon.spy(PresentationBackend, "initialize");
+      const frontendInitializationSpy = sinon.spy(PresentationFrontend, "initialize");
```
grigasp, line 23: I suggest we stub **all** these `initialize`, `startup` and `terminate` calls in `beforeEach` - that we the real ones won't be called, thus speeding up the tests.

grigasp (changes requested):

packages/testing/src/test/Helpers.test.ts
```diff
@@ -2,102 +2,112 @@
...
       sinon.resetHistory();
 
       await initialize();
-      expect(backendInitializationSpy.notCalled);
-      expect(frontendInitializationSpy.notCalled);
+      expect(backendInitializationStub).to.not.be.called;
+      expect(frontendInitializationStub).to.not.be.called;
     });
 
     it("initializes PresentationBackend and PresentationFrontend with provided props", async () => {
```
grigasp, line 48: the test name says it checks presentation frontend initialized too, although the test itself doesn't do the check

JonasDov: Added the check to the test, now the name should be correct.

packages/testing/src/test/Helpers.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    beforeEach(() => {
+      backendTerminationStub = sinon.stub(PresentationBackend, "terminate");
+      frontendTerminationStub = sinon.stub(PresentationFrontend, "terminate");
+      rimrafSyncStub = sinon.stub(rimraf, "sync");
+      sinon.stub(PresentationBackend, "initialize");
+      sinon.stub(PresentationFrontend, "initialize");
+      sinon.stub(IModelHost, "startup");
+      sinon.stub(NoRenderApp, "startup");
+      sinon.stub(IModelApp, "localization").get(() => ({ getLanguageList: () => ["en"] }));
```
grigasp, line 81: it seems both `initialize` and `terminate` suites use the same stubs - maybe move them up a level to avoid repeated code?

JonasDov: Moved all except rimrafSyncStub, because it is only used in "terminate" suite.

grigasp (approved):

Issue #204, opened 2023-08-04
Link issues with releases


grigasp:
At the moment, when generating the changelog for a release, we add a commit link for every entry. While this is useful to find a specific PR or issue when looking at the changelog, it's not very helpful when we want to find which releases a specific PR got into.

We could use GH REST API to find all PRs and Issues associated with a commit, but need to somehow link them with particular release. One approach could be to release through a PR similar to this: https://github.com/iTwin/iTwinUI/pull/1453. That way the issues mentioned in release PR description get automatically linked to it.

PR #227, opened 2023-08-31
Add `changesets` for versioning

saskliutas:
Closes https://github.com/iTwin/presentation/issues/204

Added initial config for `changesets` and pipeline for releases using `changesets` github action. Until we test out `changesets` change files should be generated using both tools: `changesets` and `beachball`. This will allow to generate correct CHANGELOG file if we decide to stick with beachball.


.changeset/changelog-github.cjs
```diff
@@ -0,0 +1,43 @@
...
+    // indent everything so it's aligned with the first line
+    if (futureLines.length > 0) {
+      returnVal += `\n${futureLines.map((l) => `  ${l}`).join("\n")}`;
+    }
+
+    return returnVal;
+  },
+
+  getDependencyReleaseLine: async (changesets, dependenciesUpdated) => {
+    if (dependenciesUpdated.length === 0) return "";
```
grigasp, line 37:
```suggestion
    if (dependenciesUpdated.length === 0) {
      return "";
    }
```

grigasp (approved):
let's give this a shot

PR #231, opened 2023-09-04
Extend API of presentation table

ViliusBa:
Expand the implementation of  usePresentationTableWithUnifiedSelection() to additionally return:
- an onSelect() callback which will update the unifiedSelection one level above it (+1).
- an array of selectedRows which are filled everytime unifiedSelection changes in the level of the table component.

These changes to the API are followed by tests and an ExperimentalTableWidget which includes a react-table v8 and implements full functionality of the extended API.


packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,59 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
...
-  return usePresentationTable({ ...props, keys });
-}
+  const [selectedRows, setSelectedRows] = useState<TableRowDefinition[]>();
+
+  const { imodel, ruleset, pageSize, columnMapper, rowMapper } = props;
+  const columns = useColumns({ imodel, ruleset, keys });
+  const { options, sort, filter } = useTableOptions({ columns });
+  const { rows, isLoading, loadMoreRows } = useRows({ imodel, ruleset, keys, pageSize, options });
+
+  const tableGuid: string = useMemo(() => window.crypto.randomUUID(), []);
```
saskliutas, line 109: You can use `Guid` from `@itwin/core-bentley` to generate random guid.

```suggestion
  const tableGuid = useMemo(() => Guid.createValue(), []);
```

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,59 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
...
+
+  useEffect(() => {
+    const disposeListener = Presentation.selection.selectionChange.addListener(({ level, keys: toggledRowKeys, source }) => {
+      if (level !== (unifiedSelection?.selectionLevel ?? 0) + 1 || source === tableGuid) {
```
saskliutas, line 113: I see `(unifiedSelection?.selectionLevel ?? 0) + 1` calculation is done in couple places. I think it could be stored into a variable in `usePresentationTableWithUnifiedSelection`

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,59 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
...
+        return;
+      }
+
+      const rowsToAddToSelection: TableRowDefinition[] = [];
+      toggledRowKeys.forEach((key) => {
+        // should return just one row
+        const selectedRow = rows.filter((row) => row.key === JSON.stringify(key))[0];
```
saskliutas, line 120: This might result in `undefined` being pushed to `rowsToAddToSelection` array if some instance is selected for which table row is not loaded yet.

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,59 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
...
+        rowsToAddToSelection.push(selectedRow);
+      });
 
+      setSelectedRows(rowsToAddToSelection);
+    });
+
+    return disposeListener;
+  }, [rows, tableGuid, unifiedSelection?.selectionLevel]);
+
+  const onSelect = (selectedData: string[]) => {
```
saskliutas, line 130: I think this function should not update selection if valid keys are passed but this table does not have rows for those keys.

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,59 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
  */
 export function usePresentationTableWithUnifiedSelection<TColumn, TRow>(
   props: Omit<UsePresentationTableProps<TColumn, TRow>, "keys">,
-): UsePresentationTableResult<TColumn, TRow> {
+): UsePresentationTableWithUnifiedSelectionResult<TColumn, TRow> {
   const unifiedSelection = useUnifiedSelectionContext();
   const keys = unifiedSelection?.getSelection() ?? emptyKeySet;
-  return usePresentationTable({ ...props, keys });
-}
+  const [selectedRows, setSelectedRows] = useState<TableRowDefinition[]>();
```
saskliutas, line 102: This list of rows should always be in sync with unified selection at `+1` but it gets out of sync when `onSelect` is called. It should be updated when `onSelect` is replacing selection or `selectionChange` event handler should not skip selection changes caused by this hook.

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,59 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
  */
 export function usePresentationTableWithUnifiedSelection<TColumn, TRow>(
```
saskliutas, line 97: Current implementation does not handle selection of lazy loaded rows. For example:
- Table loads 10 rows.
- Other component updates unified selection and puts instance key of instance representated by the row 11 which is not yet loaded.
- Table is scrolled and 11 row is loaded.
- `selectedRows` returns by this hook should include 11th row.

saskliutas (changes requested):

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,74 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
...
   const keys = unifiedSelection?.getSelection() ?? emptyKeySet;
-  return usePresentationTable({ ...props, keys });
+  const [selectedRows, setSelectedRows] = useState<TableRowDefinition[]>();
+
+  const { imodel, ruleset, pageSize, columnMapper, rowMapper } = props;
+  const columns = useColumns({ imodel, ruleset, keys });
+  const { options, sort, filter } = useTableOptions({ columns });
+  const { rows, isLoading, loadMoreRows } = useRows({ imodel, ruleset, keys, pageSize, options });
+
+  const unifiedSelectionLevel = useMemo(() => (unifiedSelection?.selectionLevel ?? 0) + 1, [unifiedSelection?.selectionLevel]);
```
grigasp, line 109: I don't think `useMemo` is necessary here - calculating the value is not expensive.

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -84,10 +96,74 @@ export function usePresentationTable<TColumn, TRow>(props: UsePresentationTableP
...
+        const selectedRow = rows.filter((row) => row.key === JSON.stringify(key));
+
+        if (selectedRow[0] !== undefined) {
+          rowsToAddToSelection.push(selectedRow[0]);
+        }
+      });
+      setSelectedRows(rowsToAddToSelection);
+    };
+
+    const selectedRowKeys = Presentation.selection.getSelection(imodel, unifiedSelectionLevel) ?? emptyKeySet;
```
grigasp, line 125: shouldn't `unifiedSelection` be used instead of `Presentation.selection`?

.changeset/thirty-moons-burn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": minor
+---
+
+Extend API of presentation table with unified selection.
```
grigasp, line 5: This is an API addition - could we explain what exactly was added and add an example for how to use it?

grigasp (changes requested):
Please reply to @saskliutas comments explaining how you resolved them.

packages/components/src/presentation-components/table/UsePresentationTable.ts
ViliusBa: Removed the creation of tableGuid as it isn't needed in the updated code.

packages/components/src/presentation-components/table/UsePresentationTable.ts
ViliusBa: Fixed this by adding an useMemo() hook to save this calculation above the useEffect() hook

packages/components/src/presentation-components/table/UsePresentationTable.ts
ViliusBa: Fixed this by adding an additional check to confirm that the value at the [0] index is not undefined.

packages/components/src/presentation-components/table/UsePresentationTable.ts
ViliusBa: Fixed this by checking if a row for the key exists before adding it to the unifiedSelection.

packages/components/src/presentation-components/table/UsePresentationTable.ts
ViliusBa: Fixed this by removing checking between tableGuid and event source, so when onSelect() is called, selectedRows get updated afterwards, therefore it does not go out of sync.

packages/components/src/presentation-components/table/UsePresentationTable.ts
ViliusBa: This was fixed by additionally updating `selectedRows` when the `rows` variable changes. So now every time new rows get loaded, these new rows get checked for having matching keys that exist in the `unifiedSelection`.

apps/test-app/frontend/src/components/table-widget/ExperimentalTableWidget.tsx
```diff
@@ -0,0 +1,167 @@
...
+    onRowSelectionChange: (updater) => {
+      const newRowSelection = typeof updater === "function" ? updater(rowSelection) : updater;
+
+      // collect selected row ids
+      const newSelectedRows: string[] = [];
+      for (const rowId in newRowSelection) {
+        if (rowId in newRowSelection && newRowSelection[rowId]) {
+          newSelectedRows.push(rowId);
+        }
+      }
```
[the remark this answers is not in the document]
ViliusBa: I am going to trust your experience and change it to your suggestion😃

apps/test-app/frontend/src/components/table-widget/ExperimentalTableWidget.tsx
```diff
@@ -0,0 +1,167 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { useCallback, useEffect, useRef, useState } from "react";
+import { PropertyRecord } from "@itwin/appui-abstract";
+import { IModelConnection } from "@itwin/core-frontend";
+import { ProgressRadial } from "@itwin/itwinui-react";
+import { TableCellRenderer, TableColumnDefinition, TableRowDefinition, usePresentationTableWithUnifiedSelection } from "@itwin/presentation-components";
+import { flexRender, getCoreRowModel, useReactTable } from "@tanstack/react-table";
+
+export interface TableWidgetProps {
```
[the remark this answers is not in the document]
ViliusBa: Good call, missed this somehow.

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -106,38 +106,52 @@ export function usePresentationTableWithUnifiedSelection<TColumn, TRow>(
...
-  }, [rows, tableGuid, unifiedSelection?.selectionLevel]);
+  }, [rows, unifiedSelectionLevel, imodel]);
 
   const onSelect = (selectedData: string[]) => {
     const parsedKeys: Key[] = [];
     for (const passedData of selectedData) {
       try {
         const parsedKey: Key = JSON.parse(passedData);
-        parsedKeys.push(parsedKey);
+        if (rows.some((row) => row.key === JSON.stringify(parsedKey))) {
```
grigasp, line 146: we already have a stringified version of `parsedKey` - no need to stringify it again

apps/test-app/frontend/src/components/table-widget/ExperimentalTableWidget.tsx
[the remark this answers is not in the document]
grigasp: From your response it sounds like and explanation on why I think my approach is more readable is necessary... 
1. Using `Object.entries` allows us to avoid the `if (rowId in newRowSelection)` check.
2. By naming value part of the entry in `filter` we provide meaning to it. Type of `newRowSelection` is unclear, which makes the `if (newRowSelection[rowId])` check very vague.

apps/test-app/frontend/src/components/table-widget/ExperimentalTableWidget.tsx
```diff
@@ -0,0 +1,167 @@
...
+  imodel: IModelConnection;
+  rulesetId?: string;
+}
+
+export function ExperimentalTableWidget(props: TableWidgetProps) {
```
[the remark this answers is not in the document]
grigasp: So the plan is to remove `ExperimentalTableWidget` as soon as iTwinUI releases an updated table component?

apps/test-app/frontend/src/components/table-widget/ExperimentalTableWidget.tsx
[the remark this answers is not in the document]
ViliusBa: Thanks, it is indeed more clear this way

apps/test-app/frontend/src/components/table-widget/ExperimentalTableWidget.tsx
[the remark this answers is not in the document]
ViliusBa: The plan was to showcase that the updated functionality can be implemented without hacking the table component. But to me your suggestion seems valid, we should remove it once the new _iTwinUI_ table is released, after adjusting the old `TableWidget` of course.

apps/test-app/frontend/src/components/table-widget/TableWidget.tsx
```diff
@@ -48,21 +48,20 @@ function PresentationTable(props: PresentationTableProps) {
...
   );
 
-  if (columns === undefined) {
-    return <ProgressRadial indeterminate={true} />;
-  }
-
   return (
     <Table
-      columns={columns}
+      columns={columns ?? []}
```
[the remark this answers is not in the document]
ViliusBa: This removal was unintentional, will add it back in.

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -53,6 +54,17 @@ export interface UsePresentationTableResult<TColumns, TRow> {
...
 
+/**
+ * Return type of [[usePresentationTableWithUnifiedSelection]] hook.
+ * @beta
+ */
+export interface UsePresentationTableWithUnifiedSelectionResult<TColumns, TRow> extends UsePresentationTableResult<TColumns, TRow> {
+  /** Specifies rows that have been selected (toggled) by other components on the appropriate selection level. */
+  selectedRows: TRow[];
+  /** Callback to use when a row is selected. */
+  onSelect: (selectedData: string[]) => void;
```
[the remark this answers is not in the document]
ViliusBa: Agree, will adjust.

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,89 @@ describe("usePresentationTableWithUnifiedSelection", () => {
     expect(result.current.columns).to.have.lengthOf(0);
     expect(result.current.rows).to.have.lengthOf(0);
   });
+
+  it("Adds passed keys to the unified selection with onSelect", async () => {
+    const keys = new KeySet([createTestECInstanceKey()]);
+
+    const { result } = renderHook(() => usePresentationTableWithUnifiedSelection(initialProps), { wrapper: Wrapper });
+
+    assert(!(result.current instanceof Error));
```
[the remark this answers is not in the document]
ViliusBa: Not really, will remove

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,89 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+
+  it("Gets invalid keys and does not pass any to the selection with onSelect", async () => {
+    const keys = ["this is not a valid key", ""];
+    const { result } = renderHook(() => usePresentationTableWithUnifiedSelection(initialProps), { wrapper: Wrapper });
+    assert(!(result.current instanceof Error));
+
+    const onSelectCallback = result.current.onSelect;
+
+    const replaceSpy = sinon.stub(Presentation.selection, "replaceSelection");
+    const onSelectSpy = sinon.spy(onSelectCallback);
```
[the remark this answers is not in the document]
ViliusBa: I agree, the check is not necessary therefore it is not worth spying on the function.

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,105 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+        .returns(async () => descriptor);
+      presentationManagerMock
+        .setup(async (x) => x.getContentAndSize(moq.It.is((options) => options.keys.size === args.keys.size)))
+        .returns(async () => ({ content: new Content(descriptor, [item]), size: 1 }));
+
+      const { result } = renderHook(() => usePresentationTableWithUnifiedSelection(initialProps), { wrapper: Wrapper });
+      const resultBeforeAdding = result.current.selectedRows;
+      expect(resultBeforeAdding.length).to.be.equal(0);
+
+      await waitFor(() => expect(result.current.isLoading).to.be.false);
```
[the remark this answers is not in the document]
ViliusBa: We should, will adjust

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,105 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+    expect(onSelectSpy).to.not.have.thrown();
+    expect(replaceSpy).to.have.been.calledOnceWithExactly("UnifiedSelectionContext", {}, [], 1);
+  });
+
+  [
+    {
+      selectionLevel: 3,
+      keys: new KeySet(),
+      returnLength: 0,
+      testName: "Returns an empty array of selectedRows when keys are passed from the wrong level",
```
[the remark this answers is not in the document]
ViliusBa: Discussed in private, test was rewritten.

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,105 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+      presentationManagerMock
+        .setup(async (x) => x.getContentAndSize(moq.It.is((options) => options.keys.size === args.keys.size)))
+        .returns(async () => ({ content: new Content(descriptor, [item]), size: 1 }));
+
+      const { result } = renderHook(() => usePresentationTableWithUnifiedSelection(initialProps), { wrapper: Wrapper });
+      const resultBeforeAdding = result.current.selectedRows;
+      expect(resultBeforeAdding.length).to.be.equal(0);
+
+      await waitFor(() => expect(result.current.isLoading).to.be.false);
+      Presentation.selection.addToSelection("UnifiedSelectionContext", initialProps.imodel, new KeySet([createTestECInstanceKey()]), args.selectionLevel);
```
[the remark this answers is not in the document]
ViliusBa: Discussed in private, test was rewritten.

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -61,8 +61,8 @@ export interface UsePresentationTableResult<TColumns, TRow> {
 export interface UsePresentationTableWithUnifiedSelectionResult<TColumns, TRow> extends UsePresentationTableResult<TColumns, TRow> {
   /** Specifies rows that have been selected (toggled) by other components on the appropriate selection level. */
   selectedRows: TRow[];
-  /** Callback to use when a row is selected. */
-  onSelect: (selectedData: string[]) => void;
+  /** Callback to use when a row is selected. It needs to be called with keys of selected rows that have been converted to strings. */
+  onSelect: (selectedKeys: string[]) => void;
```
grigasp: I don't think API consumers need to know anything about keys being converted to strings - that's an implementation detail.

```suggestion
  /** 
   * A function that should be called when a table row is selected.
   * @param selectedRowKeys Keys of selected table rows. These should match `TableRowDefinition.key` passed to `UsePresentationTableProps.rowMapper` function when new rows are loaded.
   */
  onSelect: (selectedRowKeys: string[]) => void;
```

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,89 @@ describe("usePresentationTableWithUnifiedSelection", () => {
     expect(result.current.columns).to.have.lengthOf(0);
     expect(result.current.rows).to.have.lengthOf(0);
   });
+
+  it("Adds passed keys to the unified selection with onSelect", async () => {
```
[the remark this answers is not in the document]
grigasp: I see there are still a few tests starting with upper case

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,101 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+
+    await waitFor(() => expect(result.current.isLoading).to.be.false);
+    Presentation.selection.addToSelection("UnifiedSelectionContext", initialProps.imodel, new KeySet([createTestECInstanceKey()]), 3);
+
+    await waitFor(() => expect(result.current.isLoading).to.be.false);
+    const resultAfterAdding = result.current.selectedRows;
+    expect(resultAfterAdding.length).to.be.equal(0);
+  });
+
+  function getMockData(): { descriptor: Descriptor; item: Item } {
```
grigasp, line 257: I believe this function could also set up `presentationManagerMock` to further reduce clutter in the tests

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,101 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+
+  it("adds passed keys to the unified selection with onSelect", async () => {
+    const keys = new KeySet([createTestECInstanceKey()]);
+    const stringifiedKeys: string[] = [];
+    const validKeys: Key[] = [];
+
+    keys.forEach((key) => {
+      stringifiedKeys.push(JSON.stringify(key));
+      validKeys.push(key);
+    });
```
grigasp, line 183: when testing valid situations, could we just take `rows` from `usePresentationTableWithUnifiedSelection` result and use their `key`? that would mimic the real world scenario.

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -171,4 +171,101 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+
+    const { result } = renderHook(() => usePresentationTableWithUnifiedSelection(initialProps), { wrapper: Wrapper });
+
+    await waitFor(() => expect(result.current.isLoading).to.be.false);
+    Presentation.selection.addToSelection("UnifiedSelectionContext", initialProps.imodel, keys, 1);
+
+    await waitFor(() => expect(result.current.isLoading).to.be.false);
+    const resultAfterAdding = result.current.selectedRows;
+    expect(resultAfterAdding.length).to.be.equal(1);
+  });
```
grigasp, line 244: I believe we need multiple variations of this test:
1. Ensure `selectedRows` is correct if unified selection already contains a key of one of the rows when the table is initially loaded.
2. Ensure `selectedRows` is correct when initially unified selection changes after the initial table load. For that, we need to rows in the table. And we have 4 cases:
   - 1 row is selected on initial load, 1 row is added to unified selection using `addToSelection` and we ensure both rows get selected in the table.
   - rows are selected on initial load, 1 row is removed from unified selection using `removeFromSelection` and we ensure only one is selected in the table.
   - row is selected on initial load, we change the selection to the other one using `replaceSelection` and ensure table selection is valid.
   - rows are selected, we clear the selection using `clearSelection` and ensure no rows are selected in the table.

grigasp (changes requested):

packages/components/src/presentation-components/table/UsePresentationTable.ts
```diff
@@ -119,22 +124,19 @@ export function usePresentationTableWithUnifiedSelection<TColumn, TRow>(
...
       setSelectedRows(rowsToAddToSelection);
     };
 
-    const selectedRowKeys = unifiedSelection?.getSelection(unifiedSelectionLevel) ?? emptyKeySet;
-    if (selectedRowKeys !== emptyKeySet) {
-      updateSelectedRows(selectedRowKeys);
-    }
-
-    const disposeListener = Presentation.selection.selectionChange.addListener(({ level, keys: toggledRowKeys }) => {
+    const disposeListener = Presentation.selection.selectionChange.addListener(({ level }) => {
```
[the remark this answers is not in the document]
ViliusBa: Yes, it required an addition of `selectedRows` to the dependencies of the `useEffect()` hook which in turn made our test-app freeze occasionally.

packages/components/src/test/table/UsePresentationTable.test.tsx
grigasp: Could you also check existing tests and see if they could reuse this function? E.g. it looks like `loads columns and rows with keys from unified selection` could...

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -223,24 +218,113 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
     const keys = new KeySet([createTestECInstanceKey()]);
     sinon.stub(Presentation.selection, "getSelection").returns(new KeySet([createTestECInstanceKey()]));
 
-    const { descriptor, item } = getMockData();
-    presentationManagerMock.setup(async (x) => x.getContentDescriptor(moq.It.is((options) => options.keys.size === keys.size))).returns(async () => descriptor);
-    presentationManagerMock
-      .setup(async (x) => x.getContentAndSize(moq.It.is((options) => options.keys.size === keys.size)))
-      .returns(async () => ({ content: new Content(descriptor, [item]), size: 1 }));
+    setupPresentationManager();
+    Presentation.selection.addToSelection("UnifiedSelectionContext", initialProps.imodel, keys, 1);
```
grigasp, line 226: Could we avoid the situation of partially stubbing some object? In this situation, you're stubbing `getSelection` to always return something and then calling the real `addToSelection` implementation. This makes `getSelection` and `addToSelection` behave inconsistently, because no matter what you add, you'll always get the same result. This makes it very hard to understand the test... IMO, for this test you shouldn't stub selection manager.

packages/components/src/test/table/UsePresentationTable.test.tsx
ViliusBa: I did check on this before, but decided to leave them untouched since these tests do explicit checking of the values in the setting up part. I think getting the values that are checked from a separate call would reduce the readability of the test.

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -223,24 +218,113 @@ describe("usePresentationTableWithUnifiedSelection", () => {
...
+  it("returns an array of selectedRows when keys are added on selectionChange event", async () => {
+    const instanceKey1 = createTestECInstanceKey();
+    const instanceKey2 = createTestECInstanceKey({ className: "TestSchema:TestClass", id: "0x2" });
+    const keySet1 = new KeySet([instanceKey1]);
+    const keySet2 = new KeySet([instanceKey2]);
+    const keySet3 = new KeySet([instanceKey1, instanceKey2]);
+
+    setupPresentationManager(2);
+    Presentation.selection.addToSelection("UnifiedSelectionContext", imodel, keySet3, 0);
+    Presentation.selection.addToSelection("UnifiedSelectionContext", imodel, keySet1, 1);
```
grigasp, line 243: - [ ] `id` of `instanceKey1` is not clear here - I recommend specifying it explicitly, even if it matches the default value.
- [ ] For `instanceKey2` you probably don't need to specify a class.
- [ ] I don't think you need to create these `KeySet` objects here. They're all used only once, their name tells nothing and they only make it harder to read the code - whenever this object I used I have to look up the declaration to see what's stored in there. Instead, I recommend just creating those keysets where you use them.
- [ ] It's not clear what instances `setupPresentationManager` creates rows for. I recommend changing it to take a list of instance keys rather than a number.
- [ ] Seeing two calls to `Presentation.selection.addToSelection` is a bit confusing - can you add comments why you call each of them? E.g.
   ```ts
   // select both instances at level 0 to get them displayed in the component
   Presentation.selection.addToSelection("UnifiedSelectionContext", imodel, new KeySet([ instanceKey1, instanceKey2 ]), 0);
   // select instanceKey1 at level 1 to get its row selected 
   Presentation.selection.addToSelection("UnifiedSelectionContext", imodel, new KeySet([ instanceKey1 ]), 1);
   ```

grigasp (changes requested):

packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -338,8 +341,8 @@ describe("usePresentationTableWithUnifiedSelection", () => {
     expect(selectedRowsAfterAdding.length).to.be.equal(0);
   });
 
-  /** Set the itemCount variable to adjust the number of rows added */
-  function setupPresentationManager(itemCount: number = 1) {
+  /** Creates rows for the provided keys */
+  function setupPresentationManager(keys: Key[] = [createTestECInstanceKey()]) {
```
grigasp, line 345:
```suggestion
  function setupPresentationManager(keys: InstanceKey[] = [createTestECInstanceKey()]) {
```

grigasp (approved):

PR #237, opened 2023-09-05
Stateless hierarchies package cleanup

grigasp:
- Some renames and file moves
- Add release tags to exported stuff
- Avoid exporting any `@internal` APIs through barrel file
- Drop `presentation-common` peer dependency
- Stop exposing `rxjs` stuff through our public API


.changeset/three-lamps-tell.md
```diff
@@ -0,0 +1,2 @@
+---
```
saskliutas, line 1: I think we should add `@itwin/presentation-hierarchy-builder` to ignore list in changesets config until we are ready to start versioning this package.

saskliutas (approved):

.changeset/three-lamps-tell.md
grigasp: tried and turns out it's not that trivial - I'll look into that separately, outside of this PR's scope

Issue #239, opened 2023-09-06
API for building select clause


grigasp:
We're going to request consumers to provide us ECSQL queries that we read into hierarchy nodes. Need an API that would help ensure these queries select what we expect to get when parsing results.

PR #258, opened 2023-09-08
Hierarchy builder: Select clause building API

grigasp:
Closes https://github.com/iTwin/presentation/issues/239


grigasp:
Regression tests failing due to a regression in `@itwin/presentation-backend@4.1.4`. Fix here: https://github.com/iTwin/itwinjs-core/pull/5957.

JonasDov (approved):

PR #264, opened 2023-09-15
Fix NumericPropertyEditor commit logic

ViliusBa:
Change _NumericPropertyEditor_ logic to only commit changes on `onBlur` event.

This fixes a bug where _NumericPropertyEditor_ was basically unusable because it committed changes on every input change.


packages/components/src/test/properties/NumericPropertyInput.test.tsx
```diff
@@ -259,4 +259,35 @@ describe("<NumericInput />", () => {
...
+  it("does not fire `onBlur` when value is changed but the input is not blurred", async () => {
+    const user = userEvent.setup();
+    const spy = sinon.spy();
+    const { getByRole } = render(<NumericInput handleBlur={spy} onChange={() => {}} value="2" />);
+    const inputContainer = await waitFor(() => getByRole("textbox"));
+
+    await user.type(inputContainer, "1");
+
+    expect(spy).to.be.not.be.called;
+  });
```
grigasp, line 272: I don't think this test is necessary. 

Generally we test something for doing things, unless those things are commonly done and it's unusual to not do them. Calling a blur callback on value change is not a common thing to do, so we don't need to test this.

packages/components/src/test/properties/NumericPropertyEditor.test.tsx
```diff
@@ -38,7 +38,7 @@ describe("<NumericPropertyEditorBase />", () => {
     await waitFor(() => expect(getByTestId("numeric-input")).to.not.be.null);
   });
 
-  it("invokes `onCommit` when input changes", async () => {
+  it("Does not invoke `onCommit` when input changes but blur event is not triggered", async () => {
```
grigasp, line 41: I don't think this test is necessary anymore (see my other comment regarding unnecessary tests).

To make sure the fix is properly tested we could improve the test below by making the user type more than once before blur. Previously that would've caused as many commits as there were `type` calls. Now it doesn't matter and commit should only happen once - on blur.

grigasp (changes requested):

packages/components/src/presentation-components/properties/NumericPropertyInput.tsx
```diff
@@ -66,11 +69,12 @@ const getInputTargetFromPropertyRecord = (propertyRecord: PropertyRecord) => {
 /** @internal */
 export interface NumericInputProps {
   onChange: (newValue: string) => void;
+  handleBlur?: React.FocusEventHandler;
```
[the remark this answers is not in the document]
ViliusBa: Discussed in private, will change.

grigasp (approved):

PR #268, opened 2023-09-18
Merge master into next

ViliusBa:


.github/workflows/ci.yml
grigasp, line 1: The command `pnpm check --branch origin/next` should be changed to `pnpm check --since origin/next`

grigasp (changes requested):
- I see you added `changeset` files for the old `beachball` changeset files that we had in `next` branch. However, it seems you forgot to remove at least some of the old change files.
- `@itwin/presentation-common` dependencies in `hierarchy-builder` should be removed to match `master`.
- It seems `CHANGELOG.json` in `opentelemetry` and `components` packages are missing some entries compared to what we have on `master`.
- `presentation-components.ts` is missing a `PresentationInstanceFilterBuilder` re-export.

.changeset/config.json
```diff
@@ -5,7 +5,7 @@
   "fixed": [],
   "linked": [],
   "access": "public",
-  "baseBranch": "next",
+  "baseBranch": "master",
```
grigasp, line 8: why not `next`?

ViliusBa: Because it was this way before and I only changed it initially to see if it will help pass the CI build. However, it didn't and since I do not fully understand what impact does this have, I decided to not touch it.

grigasp (approved):

PR #272, opened 2023-09-21
Fix unique value selector

ViliusBa:


.changeset/silent-plums-begin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Fix issue where UniquePropertyValueSelector selects all values when only one of the values is selected.
```
grigasp, line 5: `UniquePropertyValueSelector` is an internal component that consumers will know nothing about. Let's just name it "value selector".

```suggestion
Fix property value selector showing all values as selected when only one of the values is actually selected.
```

grigasp (approved):

PR #273, opened 2023-09-21
Add explicit zIndex for tooltip

ViliusBa:
Added specific `zIndex` property for the tooltip component. 

The `zIndex` value was chosen by checking what is the value of the parent element, which was _99999_, and adding one additional _9_ to it.

### DIFF
--- a/.changeset/big-weeks-refuse.md
+++ b/.changeset/big-weeks-refuse.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Fix bug where property tooltip appears to be covered by propertySelector.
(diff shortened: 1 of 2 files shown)


.changeset/big-weeks-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Fix bug where property tooltip appears to be covered by propertySelector.
```
grigasp, line 5:
```suggestion
Fix property tooltip appearing behind the property selector.
```

grigasp (approved):

PR #277, opened 2023-09-22
Hierarchy builder: Drop peer deps

grigasp:
Peer dependencies have a number of downsides and we want to avoid them as much as possible. In this PR I:
- Removeed all peer dependencies from `@itwin/presentation-hierarchy-builder` (had them on `core-bentley`, `core-common` and `ecschema-metadata`).
- Introduced interfaces for stuff that we needed from those peer dependencies.
- Introduced a new `@itwin/presentation-core-interop` package that implements those interfaces using core packages.

So `@itwin/presentation-hierarchy-builder` now doesn't have any peer deps as they have been moved to `@itwin/presentation-core-interop`. Having this separation provides:
- A clear interface of what `@itwin/presentation-hierarchy-builder` really needs.
- Ability to avoid bumping `@itwin/presentation-hierarchy-builder` by a major version only because of peer dependency update. That could happen only to `@itwin/presentation-core-interop` whose sole purpose is interop.


grigasp:
See some more issues, will re-open when they're fixed.

packages/hierarchy-builder/src/hierarchy-builder/internal/operators/PersistChildren.ts
```diff
@@ -4,8 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { mergeAll, Observable, reduce, tap } from "rxjs";
-import { Logger } from "@itwin/core-bentley";
+
```
JonasDov, line 7:
```suggestion
```

JonasDov (approved):

PR #278, opened 2023-09-25
Support for displaying and selecting empty string in the query builder

ViliusBa:
Introduced additional checking for undefined and empty string values.


.changeset/silly-spoons-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+"@itwin/presentation-components": minor
+---
+
+- `null` values will no longer be represented as empty fields in the query builder. Instead, building a query for `null` values will only be possible with `isNull` or `isNotNull` operators.
+- Empty strings will be selectable and displayed as _Empty Value_ in the query builder.
```
grigasp, line 6:
```suggestion
Property value selector in instance filter builder changes:
- `null` values are no longer represented as empty fields - instead, building a query for `null` values should be done using "Is Null" and "Is Not Null" operators.
- Empty strings are selectable and displayed as _Empty Value_ in the selector.
```

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
```diff
@@ -207,4 +222,41 @@ describe("UniquePropertyValuesSelector", () => {
...
+
+    const { queryByText, container, user } = render(
+      <UniquePropertyValuesSelector property={propertyDescription} onChange={() => {}} imodel={testImodel} descriptor={descriptor} />,
+    );
+
+    // open menu
+    const selector = await waitFor(() => queryByText("unique-values-property-editor.select-values"));
+    await user.click(selector!);
+
+    // assert that tno row is loaded in the dropdown.
```
grigasp, line 240:
```suggestion
    // assert that no row is loaded in the dropdown.
```

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
```diff
@@ -207,4 +222,41 @@ describe("UniquePropertyValuesSelector", () => {
...
+    await waitFor(() => expect(container.querySelectorAll(".iui-menu-item").length).to.be.equal(0));
+  });
+
+  it("loads row with empty string as displayValue and sets it to an 'Empty Value' string", async () => {
+    sinon.stub(Presentation.presentation, "getPagedDistinctValues").resolves({
+      total: 1,
+      items: [{ displayValue: "", groupedRawValues: [""] }],
+    });
```
grigasp, line 248: Let's have one more test with 
```ts
items: [{ displayValue: "", groupedRawValues: [undefined, ""] }],
```

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
```diff
@@ -165,6 +165,21 @@ describe("UniquePropertyValuesSelector", () => {
...
+        onChange={() => {}}
+        imodel={testImodel}
+        descriptor={descriptor}
+        value={{ valueFormat: PropertyValueFormat.Primitive, displayValue: '[""]', value: '[""]' }}
+      />,
+    );
+    await waitFor(() => expect(container.querySelector(".iui-tag-label")?.innerHTML).to.include("Empty Value"));
+  });
+
+  // This test does not represent a real situation, since one would only be able to create a tag-label if it is selected in the filter
```
grigasp, line 181: why do we need the test then?

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
```diff
@@ -165,6 +165,21 @@ describe("UniquePropertyValuesSelector", () => {
     expect(queryByText("unique-values-property-editor.select-values")).to.not.be.null;
   });
 
+  it("sets empty value text if provided value is an empty string", async () => {
```
grigasp, line 168: doesn't `loads row with empty string as displayValue and sets it to an 'Empty Value' string` cover this?

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
```diff
@@ -17,7 +17,10 @@ import { getInstanceFilterFieldName } from "../instance-filter-builder/Utils";
 /** @internal */
 export const UNIQUE_PROPERTY_VALUES_BATCH_SIZE = 100;
 
-/** @interna */
+/** @internal */
```
grigasp, line 20: no need to add a release tag if the variable is not exported

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
```diff
@@ -166,7 +169,7 @@ function useUniquePropertyValuesLoader({ imodel, ruleset, fieldDescriptor }: Use
         keys: new KeySet(),
       });
       return {
-        options: content.items.filter((item) => item.displayValue !== undefined && item.displayValue !== ""),
+        options: content.items.filter((item) => item.displayValue !== undefined && item.displayValue !== nullStringIdentifier),
```
grigasp, line 172: I don't understand this - what should cause `getPagedDistinctValues` to return `nullStringIdentifier` as display value?

grigasp (changes requested):

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
ViliusBa: This test covers the creation of tag-label elements for empty display values, whereas `loads row with empty string as displayValue and sets it to an 'Empty Value' string` covers the empty string value that is displayed in the selector dropdown.

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
ViliusBa: This test was written before, I decided to leave it since, I couldn't find a better way to reach a branch where an option has undefined display value in [this line](https://github.com/iTwin/presentation/pull/278/files/2b15c69cc6dde5c85e34bd7873d58fa533370c62#diff-cd3479e259eb8fe117da9a547677eb07e3a6fdef707a4de33927511785bea27cR85).

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
ViliusBa: Display value is set to `nullStringIdentifier` in [this line](https://github.com/iTwin/presentation/pull/278/files/2b15c69cc6dde5c85e34bd7873d58fa533370c62#diff-cd3479e259eb8fe117da9a547677eb07e3a6fdef707a4de33927511785bea27cR85) on _UniquePropertyValueSelector_ initial load. Thus, `getPagedDistinctValues` returns `content` with updated `displayValues`.

packages/components/src/presentation-components/instance-filter-builder/MultiTagSelect.tsx
```diff
@@ -89,7 +89,7 @@ function TagSelectOption<TOption, IsMulti extends boolean = boolean>({ children:
 
   return (
     <components.Option {...props} className={className}>
-      <span>{optionLabel === "" ? <i>Empty Value</i> : optionLabel}</span>
+      {optionLabel === "" ? <span className="presentation-instance-filter-empty-property-text">Empty Value</span> : <span>optionLabel</span>}
```
grigasp, line 92: - [ ] `Empty Value` needs to be localized.
- [ ] I'd change the class name to `presentation-instance-filter-special-property-value`.
- [ ] Could we make this class part of `className`? E.g.
   ```tsx
   const className = classnames("iui-menu-item", {
     "iui-focused": props.isFocused,
     "iui-active": props.isSelected,
     "presentation-instance-filter-special-property-value": (optionLabel === ""),
   });
   ```

packages/components/src/presentation-components/instance-filter-builder/MultiTagSelect.tsx
```diff
@@ -133,7 +133,13 @@ function TagContainer<TOption, IsMulti extends boolean = boolean>({ children, ..
 function TagLabel<TOption, IsMulti extends boolean = boolean>({ children, ...props }: MultiValueGenericProps<TOption, IsMulti>) {
   return (
     <components.MultiValueLabel {...props} innerProps={{ ...props.innerProps, className: "iui-tag-label" }}>
-      {children === "" ? <i style={{ padding: 1 }}>Empty Value</i> : children}
+      {children === "" ? (
+        <span className="presentation-instance-filter-empty-property-text" style={{ padding: 1 }}>
```
grigasp, line 137: I think the new class could be combined with `iui-tag-label`. Also, padding could be assigned through SCSS:
```css
.iui-tag-label.presentation-instance-filter-empty-property-text {
  padding: 1;
}
```

packages/components/src/presentation-components/properties/UniquePropertyValuesSelector.tsx
ViliusBa: Discussed in private

packages/components/src/presentation-components/instance-filter-builder/MultiTagSelect.tsx
```diff
@@ -29,6 +29,8 @@ import { AsyncPaginate, AsyncPaginateProps } from "react-select-async-paginate";
 import { SvgCaretDown, SvgCaretDownSmall, SvgCheckmarkSmall, SvgCloseSmall } from "@itwin/itwinui-icons-react";
 import { useResizeObserver } from "../common/Utils";
 
+export const EMPTY_PROPERTY_TEXT = "Empty Value";
```
grigasp, line 32: This doesn't localize the value. Please have a look at [this PR](https://github.com/iTwin/presentation/pull/176), specifically search for `unique-values-property-editor` usage

packages/components/public/locales/en/PresentationComponents.json
```diff
@@ -44,6 +44,7 @@
     "loading-target-instances": "Loading target instances"
   },
   "unique-values-property-editor": {
-    "select-values": "Select provided value(s)"
+    "select-values": "Select provided value(s)",
+    "empty-property": "Empty Value"
```
grigasp, line 48:
```suggestion
    "empty-value": "Empty value"
```

grigasp (approved):

Issue #181, opened 2023-07-25
Filtering in stateless hierarchies


grigasp:
Need to find the best way to filter stateless hierarchies. Requirements:
- Ability to find instances by: ECInstanceId, label, complex instance filter
- Ability to create the hierarchy with only those instances and their ancestors
- Non leaf matching nodes should be expandable to see their children, but be collapsed by default

PR #281, opened 2023-09-28
Hierarchy builder: Filtering

grigasp:
Closes https://github.com/iTwin/presentation/issues/181


packages/hierarchy-builder/src/test/internal/FilteringHierarchyLevelDefinitionsFactory.test.ts
```diff
@@ -0,0 +1,608 @@
...
+    const metadataProvider = {} as unknown as IMetadataProvider;
+    let stubClass: TStubClassFunc;
+    beforeEach(() => {
+      stubClass = createGetClassStub(metadataProvider).stubClass;
+    });
+
+    it("returns source definitions when filtered instance paths list is empty", async () => {
+      const sourceDefinitions: HierarchyLevelDefinition = [];
+      const sourceFactory = {
+        defineHierarchyLevel: sinon.stub().resolves(sourceDefinitions),
```
saskliutas, line 221: Looks like `sinon` is not necessary in most of these test and this could be simplified to `defineHierarchyLevel: async () => sourceDefinitions`?

saskliutas (approved):

saskliutas (approved):

packages/hierarchy-builder/src/test/internal/FilteringHierarchyLevelDefinitionsFactory.test.ts
grigasp: agree, will clean this up

grigasp: done

PR #284, opened 2023-09-28
Add fix for incomplete/missing category labels in instance filter builder

ViliusBa:
The fix was achieved by always using the parent category information if parent category is not the default presentation category.


packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -61,6 +61,8 @@ function getPropertySourceClassInfo(field: PropertiesField | NestedContentField)
   return field.pathToPrimaryClass[field.pathToPrimaryClass.length - 1].targetClassInfo;
 }
 
+export const DEFAULT_ROOT_CATEGORY_NAME = "/selected-item/";
```
grigasp, line 64: missing release tag

packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -119,8 +122,16 @@ interface CategoryInfo {
 
 function getCategoryInfo(category: CategoryDescription, categoryInfo: CategoryInfo): CategoryInfo {
   if (!category.parent) {
-    return categoryInfo;
+    if (category.name === DEFAULT_ROOT_CATEGORY_NAME) {
```
saskliutas, line 125: I think you can reduce duplication and nesting in this method:
```typescript
if (category.name === DEFAULT_ROOT_CATEGORY_NAME) {
  return categoryInfo;
}

const newInfo = {
  name: categoryInfo.name ? `${category.name}/${categoryInfo.name}` : `${category.name}`,
  label: categoryInfo.label ? `${category.label} | ${categoryInfo.label}` : `${category.label}`,
}

return category.parent ? getCategoryInfo(category.parent, newInfo) : newInfo;
```

saskliutas (approved):

saskliutas (approved):

PR #292, opened 2023-10-03
Broaden default instance filter builder dialog

### LINKED ISSUE #249: Default size of the filtering dialog is too small
The default size of the instance filtering dialog should at least accommodate placeholders.

image
image

ViliusBa:
closes #249

### DIFF
--- a/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
+++ b/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
@@ -3,6 +3,11 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
+.iui-dialog-default.presentation-instance-filter-dialog-content-container {
+  height: calc(var(--iui-component-height-large) * 7);
+  width: calc(var(--iui-size-2xl) * 9);
+}
+
 .presentation-instance-filter-dialog {
   .presentation-instance-filter-dialog-content-container {
     min-height: calc(var(--iui-component-height-large) * 7);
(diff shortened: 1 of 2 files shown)


packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
```diff
@@ -3,6 +3,11 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
+.iui-dialog-default.presentation-instance-filter-dialog-content-container {
```
saskliutas, line 6: Do not use `iui-*` css classes. It is considered an internal API and might change with new `itwinui-react` versions. Second bullet point https://github.com/iTwin/iTwinUI/wiki/Version-conflicts#general-guidelines

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
```diff
@@ -3,6 +3,11 @@
...
+  height: calc(var(--iui-component-height-large) * 7);
```
saskliutas, line 7: Is dialog resizing still working with `height` and `width` applied?

saskliutas (changes requested):

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.scss
ViliusBa: It does, but I changed the implementation with the latest commit so this is no longer relevant.

saskliutas (approved):

PR #295, opened 2023-10-05
Filter property value selector values based on parent node

### LINKED ISSUE #254: Values in property value selector should be filtered based on parent node
1. Use the "BayTown" sample iModel and "ModelElements" ruleset.
2. Open the filtering dialog for "ProcessPhysicalModel" root node.
    image
3. Select the "Model" property
4. Open the value selector - all Models in the iModel are available:
    image

Since we're filtering children of the "ProcessPhysicalModel", I'd expect only this one model to be available in the Model selector. See [this test](https://github.com/iTwin/itwinjs-core/blob/2831842ba8ac94d68f31f20c1e1e144aa1e1d443/full-stack-tests/presentation/src/frontend/Content.test.ts#L284) for an implementation example.

ViliusBa:
closes #254

### DIFF
--- a/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
+++ b/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
@@ -54,7 +54,7 @@ export interface InstanceFilterBuilderProps extends PropertyFilterBuilderRendere
   /** [Descriptor]($presentation-common) that will be used for getting [[navigationPropertyEditorContext]]. */
   descriptor: Descriptor;
   /** [Keys]($presentation-common) that will be passed through to [[FilterBuilderValueRenderer ]] */
-  filterNodeKeys?: Keys;
+  descriptorInputKeys?: Keys;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


grigasp (changes requested):
I don't like the name of `filterNodeKeys` variable. The keys doesn't necessarily store node keys, it could be instance keys as well. Also, neither variable name, nor its comment explains the meaning of it. My suggestion:
- Rename the variable to `descriptorInputKeys`.
- Add a comment telling that this should be the keys used to create the descriptor (if there were any).

The filtering component may be used outside of tree, e.g. to filter table rows. There we won't have any nodes, so that term should be avoided.

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -99,6 +99,7 @@ function TreeNodeFilterBuilderDialog(props: TreeNodeFilterBuilderDialogProps) {
       imodel={imodel}
       descriptor={filteringInfo.descriptor}
       initialFilter={filteringInfo.active}
+      filterNodeKeys={[filterNode.key]}
```
saskliutas, line 102: Might want to wrap `[filterNode.key]` into `useMemo` to avoid creating new array on each render. Keys are later used as `useEffect` dependency in `UniquePropertyValueSelector` and will cause that `useEffect` to run on each render.

saskliutas (changes requested):

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
```diff
@@ -54,7 +54,7 @@ export interface InstanceFilterBuilderProps extends PropertyFilterBuilderRendere
   /** [Descriptor]($presentation-common) that will be used for getting [[navigationPropertyEditorContext]]. */
   descriptor: Descriptor;
   /** [Keys]($presentation-common) that will be passed through to [[FilterBuilderValueRenderer ]] */
```
grigasp, line 56:
```suggestion
  /** [Keys]($presentation-common) that will be passed through to [[FilterBuilderValueRenderer]] */
```

grigasp (approved):

saskliutas (approved):

PR #296, opened 2023-10-06
Add support for formatting dates in property value selector

### LINKED ISSUE #252: DateTime values are not formatted
1. Use the "BayTown" sample iModel and "ModelElements" ruleset.
2. Open the filtering dialog for "ProcessPhysicalModel" root node.
    image
3. Select the "Create Time" property
4. Open the value selector - DateTime values are not formatted.
    image

ViliusBa:
closes #252 

Besides adding formatting to dates, with this PR the logic of creating labels for the value selector is moved into the `getOptionLabel()` callback,

### DIFF
--- a/packages/components/src/presentation-components/instance-filter-builder/MultiTagSelect.tsx
+++ b/packages/components/src/presentation-components/instance-filter-builder/MultiTagSelect.tsx
@@ -86,12 +86,12 @@ function TagSelectOption<TOption, IsMulti extends boolean = boolean>({ children:
   const className = classnames("iui-menu-item", {
     "iui-focused": props.isFocused,
     "iui-active": props.isSelected,
-    "presentation-instance-filter-special-property-value": optionLabel === "",
+    "presentation-instance-filter-special-property-value": optionLabel === translate("unique-values-property-editor.empty-value"),
   });
 
   return (
     <components.Option {...props} className={className}>
-      <span>{optionLabel === "" ? translate("unique-values-property-editor.empty-value") : optionLabel}</span>
+      <span>{optionLabel}</span>
       {props.isSelected && (
         <span className="iui-icon" style={{ marginLeft: "auto" }}>
           <SvgCheckmarkSmall />
@@ -139,7 +139,7 @@ function TagLabel<TOption, IsMulti extends boolean = boolean>({ children, ...pro
 
   return (
     <components.MultiValueLabel {...props} innerProps={{ ...props.innerProps, className }}>
-      {children === "" ? translate("unique-values-property-editor.empty-value") : children}
+      {children}
     </components.MultiValueLabel>
   );
 }
--- a/packages/components/src/test/instance-filter-builder/MultiTagSelect.test.tsx
+++ b/packages/components/src/test/instance-filter-builder/MultiTagSelect.test.tsx
@@ -4,10 +4,21 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { expect } from "chai";
+import sinon from "sinon";
+import { EmptyLocalization } from "@itwin/core-common";
+import { IModelApp } from "@itwin/core-frontend";
+import { Presentation } from "@itwin/presentation-frontend";
 import { fireEvent, render } from "@testing-library/react";
 import { MultiTagSelect } from "../../presentation-components/instance-filter-builder/MultiTagSelect";
 
 describe("MultiTagSelect", () => {
+  beforeEach(async () => {
+    const localization = new EmptyLocalization();
+    sinon.stub(IModelApp, "initialized").get(() => true);
+    sinon.stub(IModelApp, "localization").get(() => localization);
+    await Presentation.initialize();
+  });
+
   const options = [
     {
       label: "Option1",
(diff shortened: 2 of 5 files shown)


packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
```diff
@@ -294,6 +294,83 @@ describe("UniquePropertyValuesSelector", () => {
...
+        <UniquePropertyValuesSelector property={datePropertyDescription} onChange={() => {}} imodel={testImodel} descriptor={descriptor} />,
+      );
+
+      // open menu
+      const selector = await waitFor(() => queryByText("unique-values-property-editor.select-values"));
+      await user.click(selector!);
+
+      // assert that row is displayed correctly
+      const option = await waitFor(() => container.querySelector(".iui-menu-item"));
+      expect(option?.innerHTML.indexOf(new Date("1410-07-15").toLocaleDateString())).to.not.be.equal(-1);
```
saskliutas, line 320: Instead of looking for menu item element and checking `innerHTML` you should use `queryByText`:
```typescript
await waitFor(() => {
  expect(queryByText(new Date("1410-07-15").toLocaleDateString()).to.not.be.null
});
```

Same in other tests.

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
```diff
@@ -294,6 +294,83 @@ describe("UniquePropertyValuesSelector", () => {
...
+        typename: "shortDate",
+        editor: undefined,
+      };
+
+      const { queryByText, container, user } = render(
+        <UniquePropertyValuesSelector property={datePropertyDescription} onChange={() => {}} imodel={testImodel} descriptor={descriptor} />,
+      );
+
+      // open menu
+      const selector = await waitFor(() => queryByText("unique-values-property-editor.select-values"));
```
saskliutas, line 315: When looking up element for interactions (e.g. clicking, typing) use `getByText` instead of `queryByText`. `getByText` throws if element is not found returns `Element` instead of `Element | null` so there is no need to use `!` when passing that variable to other functions.

packages/components/src/test/instance-filter-builder/MultiTagSelect.test.tsx
```diff
@@ -4,10 +4,21 @@
...
+import { Presentation } from "@itwin/presentation-frontend";
 import { fireEvent, render } from "@testing-library/react";
 import { MultiTagSelect } from "../../presentation-components/instance-filter-builder/MultiTagSelect";
 
 describe("MultiTagSelect", () => {
+  beforeEach(async () => {
+    const localization = new EmptyLocalization();
+    sinon.stub(IModelApp, "initialized").get(() => true);
+    sinon.stub(IModelApp, "localization").get(() => localization);
+    await Presentation.initialize();
```
saskliutas, line 19: Is it necessary to initialize `Presentation`? Maybe it would be enough to stub `Presentation.localization` same as `IModelApp.localization`?

packages/components/src/presentation-components/instance-filter-builder/MultiTagSelect.tsx
```diff
@@ -86,12 +86,12 @@ function TagSelectOption<TOption, IsMulti extends boolean = boolean>({ children:
   const className = classnames("iui-menu-item", {
     "iui-focused": props.isFocused,
     "iui-active": props.isSelected,
-    "presentation-instance-filter-special-property-value": optionLabel === "",
+    "presentation-instance-filter-special-property-value": optionLabel === translate("unique-values-property-editor.empty-value"),
```
saskliutas, line 89: It would be better to compare by option value to avoid localizing string just for comparison. Also for this to work you will need to update `getOptionValue` function that is passed to `AsyncMultiTagSelect` in `UniquePropertyValuesSelector.tsx`. I think it should be the same as `getOptionLabel` was before your changes.

saskliutas (changes requested):

packages/components/src/test/instance-filter-builder/MultiTagSelect.test.tsx
ViliusBa: deleted this beforeEach step in the following commit.

packages/components/src/presentation-components/instance-filter-builder/MultiTagSelect.tsx
ViliusBa: good call

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
```diff
@@ -294,6 +294,86 @@ describe("UniquePropertyValuesSelector", () => {
...
+        <UniquePropertyValuesSelector property={datePropertyDescription} onChange={() => {}} imodel={testImodel} descriptor={descriptor} />,
+      );
+
+      // open menu
+      const selector = await waitFor(() => getByText("unique-values-property-editor.select-values"));
+      await user.click(selector);
+
+      // assert that row is displayed correctly
+      await waitFor(() => {
+        expect(queryByText(new Date("-17655058800").toLocaleString())).to.not.be.null;
```
saskliutas, line 372: Running `new Date("-17655058800")` in browser console returns `'Invalid Date'`. Maybe use [ISO 8601 date format](https://en.wikipedia.org/wiki/ISO_8601) something like `1410-07-15T12:34:00Z`

saskliutas (changes requested):

packages/components/src/test/properties/UniquePropertyValuesSelector.test.tsx
ViliusBa: Interesting how this works on tests but not in browsers, will adjust

saskliutas (approved):

PR #313, opened 2023-10-13
Improved size limit messages

ViliusBa:
Message in tree component when size limit is exceeded:
image
Instance filter dialog when size limit is exceeded:
image
Instance filter dialog when matching item count is within size limit:
image


grigasp:
What do you think about changing the messages to the following?

In the hierarchy:
> Please provide [additional filtering](#) - the hierarchy level exceeds allowed size of 5.

Maybe the one in the dialog could be conditional?
- When the limit is exceeded: 
  > Estimated hierarchy level size: 5+. Please provide additional filters.
- When the limit is not exceeded: 
  > Estimated hierarchy level size: 4.

packages/components/src/presentation-components/tree/controlled/PresentationTreeNodeRenderer.tsx
```diff
@@ -22,6 +22,7 @@ import { isPresentationInfoTreeNodeItem, isPresentationTreeNodeItem, Presentatio
 export interface PresentationTreeNodeRendererProps extends TreeNodeRendererProps {
   onFilterClick: (node: PresentationTreeNodeItem) => void;
   onClearFilterClick: (node: PresentationTreeNodeItem) => void;
+  getTreeModel?: () => TreeModel;
```
saskliutas, line 25: Maybe `onFilterClick` and `onClearFilterClick` can take `nodeId: string` instead of `node: PresentationTreeNodeItem`? That way you won't need `getTreeModel` and can just call `onFilterClick(node.parentId)`

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -116,13 +117,22 @@ function MatchingInstancesCount({ filter, dataProvider, parentKey }: MatchingIns
   const { value, inProgress } = useDebouncedAsyncValue(
     useCallback(async () => {
       const instanceFilter = await convertToInstanceFilterDefinition(filter.filter, dataProvider.imodel);
+      const requestOptions = dataProvider.createRequestOptions(parentKey, instanceFilter);
 
       try {
-        const count = await Presentation.presentation.getNodesCount(dataProvider.createRequestOptions(parentKey, instanceFilter));
-        return `${translate("tree.filter-dialog.results-count")}: ${count}`;
+        const count = await Presentation.presentation.getNodesCount(requestOptions);
+        return <>{`${translate("tree.filter-dialog.results-count")}: ${count}`}</>;
```
saskliutas, line 124: Why different styling for `tree.filter-dialog.results-count` message and result set too large message?

packages/components/src/presentation-components/tree/PresentationTreeNodeItem.ts
```diff
@@ -60,6 +68,8 @@ export interface PresentationInfoTreeNodeItem extends ImmediatelyLoadedTreeNodeI
   isSelectionDisabled: true;
   /** This type of tree item cannot have children. */
   children: undefined;
+  /** Type of item message */
+  type?: InfoTreeNodeItemType;
```
saskliutas, line 72: I think this should be non optional.

saskliutas (changes requested):

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
ViliusBa: Because I wanted the informational messages to match the ones that are displayed in the tree. Also, I thought that it is a good idea to have different styling for messages when everything goes well and another styling for _when something goes wrong_.

saskliutas: I don't think in this case `ResultSetTooLarge` error means that something went wrong. We just want to provide some indication to user if current filter is enough to limit hierarchy level to required size or not. So both messages should have same format in my opinion.

ViliusBa: Okay, I will change it to match

packages/components/src/presentation-components/tree/controlled/PresentationTreeNodeRenderer.tsx
ViliusBa: Good call, will change it

packages/components/src/presentation-components/tree/controlled/PresentationTreeNodeRenderer.tsx
```diff
@@ -42,19 +41,21 @@ export function PresentationTreeNodeRenderer(props: PresentationTreeNodeRenderer
...
-                    if (isTreeModelNode(parentNode) && isPresentationTreeNodeItem(parentNode.item)) {
-                      onFilterClick(parentNode.item);
-                    }
-                  }}
-                >
-                  {`${translate("tree.filtering-needed")}. `}
-                </UnderlinedButton>
+              {nodeItem.type === InfoTreeNodeItemType.ResultSetTooLarge && (
+                <span>
+                  <span>Provide </span>
```
grigasp, line 46: missing localization

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -125,13 +139,13 @@ function MatchingInstancesCount({ filter, dataProvider, parentKey }: MatchingIns
...
-              {`${translate("tree.filtering-needed")}. ${
+            <>
+              {`${
                 requestOptions.sizeLimit === undefined
                   ? translate("tree.filter-dialog.result-limit-exceeded.limit-unknown")
-                  : `${translate("tree.filter-dialog.result-limit-exceeded.limit-known")} ${requestOptions.sizeLimit}.`
-              } `}
-            </Text>
+                  : `${translate("tree.filter-dialog.result-limit-exceeded.limit-known")} ${requestOptions.sizeLimit}`
+              }. Provide ${translate("tree.additional-filtering")}.`}
```
grigasp: missing localization

packages/components/src/presentation-components/tree/controlled/PresentationTreeNodeRenderer.tsx
```diff
@@ -35,7 +35,35 @@ export function PresentationTreeNodeRenderer(props: PresentationTreeNodeRenderer
...
+        isLeaf={true}
+        label={
+          <span>
+            <Text isMuted className="info-tree-node-item">
+              {nodeItem.type === InfoTreeNodeItemType.ResultSetTooLarge && (
+                <span>
+                  <span>Provide </span>
+                  <UnderlinedButton
+                    onClick={() => {
+                      if (nodeItem.parentId !== undefined) {
```
saskliutas, line 49: Probably shouldn't show button if there is no `parentId`

grigasp (changes requested):
- Can we also change "Provide" to "Please provide"? The current sentence sounds like we're ordering user to do that.
- Should start with lower case:
   image

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -36,23 +47,26 @@ export interface PresentationTreeRendererProps extends TreeRendererProps {
...
 
   const filterableNodeRenderer = useCallback(
     (nodeProps: TreeNodeRendererProps) => {
       return (
         <PresentationTreeNodeRenderer
           {...nodeProps}
-          onFilterClick={(node) => {
-            setFilterNode(node);
+          onFilterClick={(nodeId) => {
+            const node = treeModel.getNode(nodeId);
```
saskliutas, line 60: Better use `nodeLoader.modelSource.getModel()`. That way you don't need `useMemo`

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
saskliutas: I guess it could be changed back to return `string` instead of `React.Fragment`

packages/components/src/presentation-components/tree/PresentationTreeNodeItem.ts
```diff
@@ -48,6 +48,14 @@ export interface PresentationTreeNodeItem extends DelayLoadedTreeNodeItem {
   filtering?: PresentationTreeNodeItemFilteringInfo;
 }
 
+export enum InfoTreeNodeItemType {
```
saskliutas, line 51: Missing docs and tag

saskliutas (changes requested):

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
ViliusBa: Missed these somehow, will localize.

packages/components/src/presentation-components/tree/controlled/PresentationTreeNodeRenderer.tsx
ViliusBa: Resolved in private

grigasp (approved):
Need to update API extractions and tests, but otherwise LGTM.

packages/components/src/test/tree/controlled/PresentationTreeRenderer.test.tsx
```diff
@@ -151,6 +153,36 @@ describe("PresentationTreeRenderer", () => {
...
+        undefined,
+        [createTreeModelNodeInput({ id: "A", item: { filtering: { descriptor: createTestContentDescriptor({ fields: [] }), ancestorFilters: [] } } })],
+        0,
+      );
+    });
+
+    // stub getNode method to make it return undefined when onFilterClick() is called.
+    nodeLoader.modelSource.getModel = () => {
+      return { getNode: sinon.stub().returns(undefined) } as unknown as TreeModel;
+    };
```
saskliutas, line 168: I think it should be:
```suggestion
    sinon.stub(nodeLoader.modelSource, "getModel").returns({
      getNode: () => undefined 
    } as unknown as TreeModel};
```

or 

```suggestion
    sinon.stub(nodeLoader.modelSource.getModel(), "getNode").returns(undefined);
```

saskliutas (approved):

packages/components/src/test/tree/controlled/PresentationTreeRenderer.test.tsx
ViliusBa: Will stick to latter one, seems more neat

PR #321, opened 2023-10-19
Add base class grouping and merge groupings into single operator

JonasDov:
Closes #245

PR #323, opened 2023-10-23
Hierarchy builder: API cleanup and tests

grigasp:
**Note:** The plan is to wait for https://github.com/iTwin/presentation/pull/321 to merge, then merge with `master` and publish this PR.

- Added missing test coverage, now it's only somewhat missing for grouping. Will need to be finalized after https://github.com/iTwin/presentation/pull/321 merges.

- Fixed a number of issues found by additional tests

- Made a substantial change in `HierarchyNode` and related APIs. There are now several variations of the type:

  - `ParsedHierarchyNode` is what `HierarchyDefinition` provides us either through `CustomHierarchyNodeDefinition`, or through `InstanceNodesQueryDefinition`, where we run the query to get `ParsedInstanceHierarchyNode`. Specifics about the node:
    - `children` attribute may only be a `boolean`, 
    - `label` may be either a `string` or `ConcatenatedValue`,
    - has no `parentKeys`,
    - it may have processing-related params on it (the type of params depends on whether it's a `ParsedInstanceHierarchyNode` or `ParsedCustomHierarchyNode`).

  - `ProcessedHierarchyNode` is built from `ParsedHierarchyNode` by formatting the label to `string` and assigning it the parent node keys. Specifics about the node:
    - `label` is a `string`,
    - has `parentKeys`,
    - it may have processing-related params on it.

    There are 3 types of `ProcessedHierarchyNode`:
      1. `ProcessedCustomHierarchyNode` has `key: string` and a just subset of processing params (only the ones related to hiding).
      2. `ProcessedInstanceHierarchyNode` has `key: InstancesNodeKey` and full set of processing params, including grouping ones.
      3. `ProcessedGroupingHierarchyNode` has `key: GroupingNodeKey` and `children: Array<ProcessedGroupingHierarchyNode | ProcessedInstanceHierarchyNode>`.

  - `ParentHierarchyNode` is what consumers pass to `HierarchyProvider.getNodes()` as a parent. It's similar to `HierarchyNode` except that it doesn't know anything about its children.

  - `HierarchyDefinitionParentNode` is what's passed to `IHierarchyLevelDefinitionsFactory.defineHierarchyLevel` as a parent node. It's a more concrete `ParentHierarchyNode` with the only possible `key` types being a `string` (for custom parent nodes) and `InstancesNodeKey` (for instance parent nodes).

  - `HierarchyNode` is what consumers receive by calling `HierarchyProvider.getNodes()`. Its specifics:
    - `children` attribute may only be a `boolean`, 
    - `label` is a `string`,
    - has `parentKeys`,
    - it doesn't have any processing-related params on it.


packages/hierarchy-builder/src/hierarchy-builder/internal/Common.ts
```diff
@@ -138,33 +155,45 @@ export function hasChildren<TNode extends { children?: boolean | Array<unknown>
...
 }
 
 /** @internal */
-export function createOperatorLoggingNamespace(operatorName: string) {
-  return `Presentation.HierarchyBuilder.Operators.${operatorName}`;
+// export function omit<T extends {}>(obj: T, attrs: Array<keyof T>) {
+//   const copy = { ...obj };
+//   attrs.forEach((attr) => delete copy[attr]);
+//   return copy;
+// }
```
saskliutas, line 173: Leftover?

saskliutas (approved):

packages/hierarchy-builder/src/hierarchy-builder/internal/Common.ts
grigasp: yes, cleaned up

packages/hierarchy-builder/src/hierarchy-builder/internal/operators/Grouping.ts
```diff
@@ -19,71 +19,107 @@ const OPERATOR_NAME = "Grouping";
...
+      ungrouped: groupings.ungrouped,
+    };
+  }
+  if (curr && curr.grouped.length > 0) {
+    onGroupingNodeCreated && curr.grouped.forEach(onGroupingNodeCreated);
+    return sortNodesByLabel([...curr.grouped, ...curr.ungrouped]);
   }
-  // only need to sort if the nodes list changed (e.g. grouping nodes were created)
-  return originalNodes !== nodes ? sortNodesByLabel(nodes) : nodes;
+  return nodes;
```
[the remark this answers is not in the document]
grigasp: Added a couple of tests. 

However, I couldn't come up with a case when we need to re-sort after hiding all grouping nodes - if no grouping nodes are created, then our original nodes list is what we want to return. And that list is already sorted.

PR #326, opened 2023-10-30
Add autoExpand option

### LINKED ISSUE #270: Auto-expanding grouping nodes
Need to add ability to auto-expand these nodes when they have only 1 child or always.

JonasDov:
closes #270


packages/hierarchy-builder/src/hierarchy-builder/internal/operators/grouping/AutoExpand.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import { BaseGroupingParams, HierarchyNode, HierarchyNodeHandlingParams } from "../../../HierarchyNode";
+import { GroupingHandlerResult, GroupingType } from "../Grouping";
+
+/** @internal */
+export function assignAutoExpand(props: GroupingHandlerResult): GroupingHandlerResult {
+  for (const node of props.grouped) {
+    if (Array.isArray(node.children)) {
+      const autoExpand = getGroupingAutoExpandOptionsFromParentNode(node, props.groupingType);
+      if (autoExpand === "always" || (node.children.length === 1 && !!autoExpand)) {
```
grigasp, line 14: I think the following is more readable:

```suggestion
      if (autoExpand === "always" || (autoExpand === "single-child" && node.children.length === 1)) {
```

packages/hierarchy-builder/src/hierarchy-builder/internal/operators/grouping/AutoExpand.ts
```diff
@@ -0,0 +1,58 @@
...
+        node.autoExpand = true;
```
grigasp, line 15: could `break` here?

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
```diff
@@ -106,6 +106,8 @@ export interface BaseGroupingParams {
   hideIfNoSiblings?: boolean;
   /** Hiding option that determines whether to hide group nodes which have only one node as its children. */
   hideIfOneGroupedNode?: boolean;
+  /** Option which auto expands grouping nodes' children when it has single child or always. */
+  autoExpand?: "single-child" | "always";
```
grigasp, line 110: I think it would be worth creating a `type` for this string union. At the moment we don't have type safety everywhere, e.g.  in `getGroupingAutoExpandOptionsFromParentNode`.

packages/hierarchy-builder/src/hierarchy-builder/internal/operators/grouping/AutoExpand.ts
```diff
@@ -0,0 +1,58 @@
...
+): string | undefined {
+  let autoExpand: string | undefined;
+  for (const node of nodes) {
+    if (autoExpand === "always") {
+      break;
+    }
+    const params = node.params ? autoExpandOptionsAccessor(node.params) : undefined;
+    autoExpand = !!params?.autoExpand ? params.autoExpand : autoExpand;
+  }
+  return autoExpand;
```
grigasp, line 57:
```suggestion
  for (const node of nodes) {
    const params = node.params ? autoExpandOptionsAccessor(node.params) : undefined;
    if (params?.autoExpand) {
      return params.autoExpand;
    }
  }
  return undefined;
```

packages/hierarchy-builder/src/test/internal/operators/grouping/AutoExpand.test.ts
```diff
@@ -0,0 +1,352 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import { HierarchyNode } from "../../../../hierarchy-builder/HierarchyNode";
+import { assignAutoExpand } from "../../../../hierarchy-builder/internal/operators/grouping/AutoExpand";
+import { createTestNode } from "../../../Utils";
+
+describe("AutoExpand", () => {
```
grigasp, line 11: There are 3 sets of nearly identical tests. Maybe we could have something like this to avoid repeating the code?

```ts
[
  [
    "Base class grouping", 
    {
      label: "Base Class",
      key: {
        type: "class-grouping",
        class: {
          name: "TestSchema:BaseClass",
          label: "Base Class",
        },
      },
      children: [
        createTestNode({
          key: { type: "instances", instanceKeys: [{ className: "TestSchema:A", id: "0x1" }] },
          label: "1",
          params: { grouping: { byBaseClasses: { fullClassNames: ["TestSchema:BaseClass"], autoExpand: "single-child" } } },
        }),
      ],
    },
  ],
  [
    "Class grouping",
    { ... },
  ],
  [
    "Label grouping",
    { ... },
  ],
].forEach(([scenario, groupingNode]) => {
  describe(scenario, () => {
    it("sets autoExpand to true when grouping node has one node child and it has autoExpand set to 'single-child'", async () => {
      const nodes: HierarchyNode[] = [groupingNode];
      const result = assignAutoExpand({ grouped: nodes, ungrouped: [], groupingType: "base-class" });
      expect(result.ungrouped).to.deep.eq([]);
      expect(HierarchyNode.isGroupingNode(result.grouped[0])).to.eq(true);
      expect(result.grouped[0].children).to.deep.eq(nodes[0].children);
      expect(result.grouped[0].autoExpand).to.eq(true);
    });
    // ... and other tests here
  });
});
```

packages/full-stack-tests/src/hierarchy-builder/grouping/AutoExpand.test.ts
```diff
@@ -0,0 +1,316 @@
...
+                  SELECT ${await selectClauseFactory.createSelectClause({
+                    ecClassId: { selector: `this.ECClassId` },
+                    ecInstanceId: { selector: `this.ECInstanceId` },
+                    nodeLabel: { selector: `this.UserLabel` },
+                    grouping: specifiedGrouping,
+                  })}
+                  FROM (
+                    SELECT ECClassId, ECInstanceId, UserLabel, Parent
+                    FROM ${subjectClassName}
+                  ) AS this
```
grigasp, line 50:
```suggestion
                  FROM ${subjectClassName} this
```

grigasp (changes requested):

packages/hierarchy-builder/src/hierarchy-builder/internal/operators/grouping/AutoExpand.ts
JonasDov: No, if `break` is put here, then this loop would stop on the first grouping node which has autoExpand set to true. But we need to set autoExpand for each grouping node, not only the first.

packages/hierarchy-builder/src/hierarchy-builder/internal/operators/grouping/AutoExpand.ts
JonasDov: Your suggestion does not work, when we have:
node1: {
...
 byLabel: {
  autoExpand: "single-child"
 }
}
node2: {
...
 byLabel: {
  autoExpand: "always"
 }
}
node3: {
...
 byLabel: {
  autoExpand: "single-child"
 }
}

JonasDov: Keeping initial implementation.

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
JonasDov: Added:
```
export interface BaseGroupingParams {
  /** Hiding option that determines whether to hide group nodes which have no siblings at the same hierarchy level. */
  hideIfNoSiblings?: boolean;
  /** Hiding option that determines whether to hide group nodes which have only one node as its children. */
  hideIfOneGroupedNode?: boolean;
  /** Option which auto expands grouping nodes' children when it has single child or always. */
  autoExpand?: AutoExpand;
}
/**
 * A [[BaseGroupingParams]] autoExpand option used to expand grouping nodes' children.
 * @beta
 */
export type AutoExpand = "single-child" | "always";
```

packages/hierarchy-builder/src/test/internal/operators/grouping/AutoExpand.test.ts
JonasDov: Changed to this:
```
 const baseClassGroupingParams = {
    testName: "Base class grouping",
    label: "Base Class",
    key: {
      type: "class-grouping",
      class: {
        name: "TestSchema:BaseClass",
        label: "Base Class",
      },
    },
    groupingType: "base-class",
  };
  ...
  describe("sets autoExpand to true when grouping node has one node child and it has autoExpand set to 'single-child'", () => {
    [
      {
        testParams: baseClassGroupingParams,
        processingParams: { grouping: { byBaseClasses: { fullClassNames: ["TestSchema:BaseClass"], autoExpand: "single-child" } } },
      },
      ...
    ].forEach(({ testParams, processingParams }) => {
      it(testParams.testName, async () => {
        const nodes: ProcessedInstancesGroupingHierarchyNode[] = [
          createTestProcessedGroupingNode({
            label: testParams.label,
            key: testParams.key as GroupingNodeKey,
            children: [
              createTestProcessedInstanceNode({
                key: { type: "instances", instanceKeys: [{ className: "TestSchema:A", id: "0x1" }] },
                label: "1",
                processingParams: processingParams as InstanceHierarchyNodeProcessingParams,
              }),
            ],
          }),
        ];
        const result = assignAutoExpand({ grouped: nodes, ungrouped: [], groupingType: testParams.groupingType as GroupingType });
        expect(result.ungrouped).to.deep.eq([]);
        expect(HierarchyNode.isGroupingNode(result.grouped[0])).to.eq(true);
        expect(result.grouped[0].children).to.deep.eq(nodes[0].children);
        expect(result.grouped[0].autoExpand).to.eq(true);
      });
    });
  });
// ... and other tests here
```

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
```diff
@@ -206,8 +206,14 @@ export interface BaseGroupingParams {
   /** Hiding option that determines whether to hide group nodes which have only one node as its children. */
   hideIfOneGroupedNode?: boolean;
   /** Option which auto expands grouping nodes' children when it has single child or always. */
-  autoExpand?: "single-child" | "always";
+  autoExpand?: AutoExpand;
 }
+/**
+ * A [[BaseGroupingParams]] autoExpand option used to expand grouping nodes' children.
```
grigasp, line 212:
```suggestion
 * Defines possible values for [[BaseGroupingParams.autoExpand]] attribute:
 * - `single-child` - set the grouping node to auto-expand if it groups a single node.
 * - `always` - always set the grouping node to auto-expand.
```

packages/hierarchy-builder/src/test/internal/operators/grouping/AutoExpand.test.ts
grigasp: 1. I don't think you need to parametrize the grouping node label. Could just use "Test label" or whatever.
2. I'd rename the `key` parameter name to `groupingNodeKey` to make it easier to understand what it means.
3. Could we move the extra `processingParams` parameters to `testParams`? I think each grouping params could contain a function like 
  ```ts
  {
    createGroupedNodeProcessingParams: (autoExpand: AutoExpand | undefined) => InstanceHierarchyNodeProcessingParams;
  }
  ```
  And then each test could call it to create processing params for grouped nodes.
4. `describe` should describe what you're testing (situation or API) and `it` should describe situation and read like "it does something". This is not the case now - I'd like the structure to be like this:
  ```ts
  [baseClassGroupingParams, classGroupingParams, labelGroupingParams].forEach(({ testName }) => {
    describe(testName, () => {
      it("sets autoExpand to true when grouping node has one node child and it has autoExpand set to 'single-child'", () => {
        // test...
      });

      it("sets autoExpand to true when some child nodes have autoExpand set to 'always'", () => {
        // test...
      });
    });
  });
  ```
  Any reason we can't achieve this?

JonasDov: Changed.

packages/hierarchy-builder/src/test/internal/operators/grouping/AutoExpand.test.ts
```diff
@@ -0,0 +1,163 @@
...
+import { GroupingType, ProcessedInstancesGroupingHierarchyNode } from "../../../../hierarchy-builder/internal/operators/Grouping";
+import { assignAutoExpand } from "../../../../hierarchy-builder/internal/operators/grouping/AutoExpand";
+import { createTestProcessedGroupingNode, createTestProcessedInstanceNode } from "../../../Utils";
+
+describe("AutoExpand", () => {
+  [
+    {
+      testName: "Base class grouping",
+      testParams: {
+        label: "Test label",
```
grigasp, line 17: I don't think we need `label` parameter as it's the same for all cases and we don't really care what it is. Instead, just hardcode it in each test.

JonasDov: Changed.

grigasp (approved):

Issue #341, opened 2023-11-17
Support creating filtering dialog from stateless hierarchies


grigasp:
The current implementation allows creating a hierarchy level descriptor, which we can use to build the hierarchy level filter. Need something similar for stateless hierarchies implementation.

PR #342, opened 2023-11-17
Hierarchy builder: Hierarchy level filtering

grigasp:
Closes https://github.com/iTwin/presentation/issues/298.

Follow-up: https://github.com/iTwin/presentation/issues/341.


packages/full-stack-tests/src/hierarchy-builder/HierarchyLevelFiltering.test.ts
```diff
@@ -0,0 +1,371 @@
...
+          },
+        },
+      });
+      validateHierarchyLevel({
+        nodes,
+        expect: [NodeValidators.createForInstanceNode({ instanceKeys: [keys.childSubject1] })],
+      });
+    });
+
+    it("filters by property class", async function () {
```
saskliutas, line 146: It is very difficult to understand what is happening in this test. We insert one `PhysicalPartition` element and then validating that with filter applied we get node for that element. It looks like without filter we also would get one node for that element. 

I guess it would help if hierarchy level without filter would be validated too.

packages/full-stack-tests/src/hierarchy-builder/HierarchyLevelFiltering.test.ts
```diff
@@ -0,0 +1,371 @@
...
+        const childSubject2 = insertSubject({ builder, codeValue: "test subject 2", parentId: rootSubject.id });
+        return { rootSubject, childSubject1, childSubject2 };
+      });
+
+      const selectQueryFactory = new NodeSelectQueryFactory(createMetadataProvider(imodel));
+      const hierarchy: IHierarchyLevelDefinitionsFactory = {
+        async defineHierarchyLevel({ instanceFilter }) {
+          const filterClauses = instanceFilter
+            ? await selectQueryFactory.createFilterClauses(instanceFilter, { fullName: subjectClassName, alias: "this" })
+            : undefined;
```
saskliutas, line 44: This and 
```
FROM ${filterClauses?.from ?? subjectClassName} AS this
${filterClauses?.joins ?? ""}
```

are repeated when defining each hierarchy level. I think repetition could be reduced by doing something like this:
```
// if instance filter is undefined `createFilterClauses` returns { from: subjectClassName, joins: "" }
const filterClauses = await selectQueryFactory.createFilterClauses(instanceFilter, { fullName: subjectClassName, alias: "this" })
// ...
ecsqli: `
  ...
  FROM ${filterClauses.from} AS this
  ${filterClauses.joins}
  ...
```

saskliutas (changes requested):

packages/full-stack-tests/src/hierarchy-builder/HierarchyLevelFiltering.test.ts
grigasp: Made the following changes to `NodeSelectQueryFactory.createFilterClauses`:
- Also accept `undefined` filter definition, in which case it returns `{ from: contentClass.fullName, joins: "", where: "" }`.
- Set the clauses to empty strings rather than `undefined` when they need to be left empty - that removes the necessity for consumers to do empty string fallbacks: `filter?.joins ?? ""`.

saskliutas (approved):

PR #349, opened 2023-11-22
Add localization to core-interop and hierarchy-builder packages

JonasDov:
prerequisite #242


packages/hierarchy-builder/public/locales/en/PresentationHierarchyBuilder.json
grigasp, line 1: I believe this file is not going to be included in the package. Please see `.npmignore` in `presentation-components` for reference.

packages/core-interop/src/core-interop/Localization.ts
```diff
@@ -0,0 +1,14 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { Localization } from "@itwin/core-common";
+
+/**
+ * Create a `Translator` that uses [Localization]($core-common) API to create a localized string.
+ * @beta
+ */
+export function createTranslator(localization: Localization): (input: string) => string {
+  return (input) => localization.getLocalizedString(input);
```
grigasp, line 13: I believe here we could take care of registering the `hierarchy-builder` localization namespace (that would mean `createTranslator` has to become `async`).

packages/hierarchy-builder/src/hierarchy-builder/Localization.ts
```diff
@@ -0,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+/** A translator implementation returns the same input. */
+const NOOP_Translator = (input: string) => input;
```
grigasp, line 8: I think this variable should be all in upper case.

packages/hierarchy-builder/src/hierarchy-builder/Localization.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export const LOCALIZATION_NAMESPACE = "PresentationHierarchyBuilder";
+
+/**
+ * Set translator to use by this package. By default the package uses a no-op translator.
+ * @beta
+ */
+export function setTranslator(translator: Translator) {
+  g_translator = translator;
+}
```
grigasp, line 31: we should allow setting to `undefined`, which would reset back to noop localization function

grigasp (changes requested):
- [ ] Maybe rename all instances of "Translator" to "LocalizationFunction"?
- [ ] Could we also have 1 full stack test that:
  1. initializes `IModelApp`
  2. sets the `hierarchy-builder` localization function
  3. gets some localized string and ensures it's localized.

packages/hierarchy-builder/public/locales/en/PresentationHierarchyBuilder.json
JonasDov: Added .npmignore

packages/core-interop/src/core-interop/Localization.ts
JonasDov: Added

packages/hierarchy-builder/src/hierarchy-builder/Localization.ts
JonasDov: Renamed

packages/hierarchy-builder/src/hierarchy-builder/Localization.ts
JonasDov: Made localizationFunction parameter optional

JonasDov:
> * [ ]  Maybe rename all instances of "Translator" to "LocalizationFunction"?
> * [ ]  Could we also have 1 full stack test that:
>   
>   1. Initializes `IModelApp`;
>   2. Sets the `hierarchy-builder` localization function;
>   3. Gets some localized string and ensures it's localized.

* [x] Renamed
* [x] Added full stack test

packages/core-interop/src/test/Localization.test.ts
```diff
@@ -0,0 +1,24 @@
...
+import { LOCALIZATION_NAMESPACE } from "@itwin/presentation-hierarchy-builder";
+import { createLocalizationFunction } from "../core-interop/Localization";
+
+describe("createTranslator", () => {
+  it("creates a localization function using provided core `Localization` object", async () => {
+    const localization = {
+      getLocalizedString: (input: string) => `${input}_localized`,
+      registerNamespace: async () => {},
+    } as unknown as Localization;
+    const spy = sinon.spy(localization, "registerNamespace");
```
grigasp, line 18: could just set the spy right away:
```ts
    const localization = {
      getLocalizedString: (input: string) => `${input}_localized`,
      registerNamespace: sinon.spy(),
    } as unknown as Localization;
```

packages/full-stack-tests/src/hierarchy-builder/Localization.test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { expect } from "chai";
+import { IModelApp } from "@itwin/core-frontend";
+import { createLocalizationFunction } from "@itwin/presentation-core-interop";
+import { setLocalizationFunction } from "@itwin/presentation-hierarchy-builder";
+import { translate } from "@itwin/presentation-hierarchy-builder/lib/cjs/hierarchy-builder/Localization";
+import { initialize, terminate } from "../IntegrationTests";
+
+describe("Stateless hierarchy builder", () => {
+  describe("Localization", () => {
+    describe("translate", () => {
```
grigasp, line 15: I don't think we need this "translate" describe.

packages/full-stack-tests/src/hierarchy-builder/Localization.test.ts
```diff
@@ -0,0 +1,32 @@
...
+      beforeEach(async () => {
+        await initialize();
+      });
+
+      afterEach(async () => {
+        await terminate();
+      });
+
+      it("translates input based on localization", async function () {
```
grigasp, line 24:
```suggestion
      it("translates strings using `IModelApp.localization`", async function () {
```

packages/hierarchy-builder/src/hierarchy-builder/Localization.ts
```diff
@@ -0,0 +1,38 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+/** A localization function implementation returns the same input. */
+const NOOP_LOCALIZATION_FUNCTION = (input: string) => input;
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+let g_localization_function = NOOP_LOCALIZATION_FUNCTION;
```
grigasp, line 10:
```suggestion
let g_localizationFunction = NOOP_LOCALIZATION_FUNCTION;
```

packages/full-stack-tests/src/hierarchy-builder/Localization.test.ts
JonasDov: Removed

packages/core-interop/src/test/Localization.test.ts
JonasDov: Changed to:
```ts
const registerNamespaceSpy = sinon.spy();
const localization = {
  getLocalizedString: (input: string) => `${input}_localized`,
  registerNamespace: registerNamespaceSpy,
} as unknown as Localization;
const translator = await createLocalizationFunction(localization);
expect(registerNamespaceSpy).to.be.calledWith(LOCALIZATION_NAMESPACE);
```
Because I am checking what the spy was called with.

grigasp (approved):

Issue #329, opened 2023-11-09
Hilite issues after selecting a class grouping node with many grouped instances


grigasp:
**Repro:**
1. Open https://qa-connect-imodelweb.bentley.com/imodeljs/?projectId=684860b3-e9bc-47cb-82fc-c243178000fa&iModelId=c47d8e40-6890-4dfb-b390-27e5bd933c93&ChangeSetId=9f570e69ab6278e61736d15ff91383d191959b44.
2. Open "Search" widget.
3. Enter "cell" in the search, wait for search to complete.
4. Select the "Physical Object" class grouping node.

```[tasklist]
### Issues:
- [x] It takes a long time to hilite the grouped elements. Investigate why.
- [ ] The requests to get the grouped elements are being sent even after the class grouping node is deselected. Could we cancel that if selection changes during the process?
- [x] The class grouping node says it's grouping 42552 elements. However, the `getPagedContent` and `getPagedContentSet` requests think there are 1000341 items in the content set. Needs investigation.
- [x] Hilite fails if the hilite set has more than 65536 items (see image below). Need to avoid this.
```

![Image](https://github.com/iTwin/presentation/assets/35135765/4e7e0b3e-4e08-4eb7-b36b-de7d0c955a05)

saskliutas:
When loading content item for hilite set we traverse `ElementOwnsChildElements` relationship recursively and that results in 1000341 elements when selecting that specific class grouping node.

grigasp:
> It takes a long time to hilite the grouped elements. Investigate why.

Performance improved with https://github.com/iTwin/imodel-native/pull/550.

grigasp:
> Hilite fails if the hilite set has more than 65536 items (see image below). Need to avoid this.

Fixed with https://github.com/iTwin/itwinjs-core/pull/6221

grigasp:
> The requests to get the grouped elements are being sent even after the class grouping node is deselected. Could we cancel that if selection changes during the process?

Part of the fix change is in https://github.com/iTwin/itwinjs-core/pull/6221. When that gets released, a change in `@itwin/presentation-components` will be needed to stop loading hilite set upon unified selection change. That feature will be included in the `5.0` major release.

PR #354, opened 2023-11-27
Cancel hilite set request on selection change

saskliutas:
Closes https://github.com/iTwin/presentation/issues/329

Use new `getHiliteSetIterator` API to highlight elements on unified selection change. This new implementation cancels ongoing hilite set request when unified selection changes again.

### DIFF
--- a/apps/test-app/frontend/src/components/app/App.tsx
+++ b/apps/test-app/frontend/src/components/app/App.tsx
@@ -6,6 +6,8 @@
 import "./App.css";
 import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
 import { useEffect, useRef, useState } from "react";
+import { from, Subject, takeUntil } from "rxjs";
+import { Id64String } from "@itwin/core-bentley";
 import { IModelApp, IModelConnection } from "@itwin/core-frontend";
 import { Geometry } from "@itwin/core-geometry";
 import { UnitSystemKey } from "@itwin/core-quantity";
@@ -65,7 +67,9 @@ export function App() {
   };
 
   useEffect(() => {
-    return Presentation.selection.selectionChange.addListener(async (args: SelectionChangeEventArgs) => {
+    const cancel = new Subject<void>();
+    const removeListener = Presentation.selection.selectionChange.addListener(async (args: SelectionChangeEventArgs) => {
+      cancel.next();
       if (!IModelApp.viewManager.selectedView) {
         // no viewport to zoom in
         return;
@@ -77,13 +81,28 @@ export function App() {
       }
 
       // determine what the viewport is hiliting
-      const hiliteSet = await Presentation.selection.getHiliteSet(args.imodel);
-      if (hiliteSet.elements) {
-        // note: the hilite list may contain models and subcategories as well - we don't
-        // care about them at this moment
-        await IModelApp.viewManager.selectedView.zoomToElements(hiliteSet.elements);
-      }
+      const selectedView = IModelApp.viewManager.selectedView;
+      const elementIds: Id64String[] = [];
+      from(Presentation.selection.getHiliteSetIterator(args.imodel))
+        .pipe(takeUntil(cancel))
+        .subscribe({
+          next: (set) => {
+            if (!set.elements) {
+              // note: the hilite list may contain models and subcategories as well - we don't
+              // care about them at this moment
+              return;
+            }
+
+            elementIds.push(...set.elements);
+            void selectedView.zoomToElements(elementIds);
+          },
+        });
     });
+
+    return () => {
+      cancel.next();
+      removeListener();
+    };
   }, []);
 
   return (
--- a/apps/test-app/frontend/src/components/app/App.tsx
+++ b/apps/test-app/frontend/src/components/app/App.tsx
@@ -6,7 +6,7 @@
 import "./App.css";
 import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
 import { useEffect, useRef, useState } from "react";
-import { from, Subject, takeUntil } from "rxjs";
+import { from, reduce, Subject, takeUntil } from "rxjs";
 import { Id64String } from "@itwin/core-bentley";
 import { IModelApp, IModelConnection } from "@itwin/core-frontend";
 import { Geometry } from "@itwin/core-geometry";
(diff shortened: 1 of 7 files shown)


packages/components/src/presentation-components/viewport/ViewportSelectionHandler.ts
```diff
@@ -69,54 +63,42 @@ export class ViewportSelectionHandler implements IDisposable {
...
-          imodel.hilited.elements.addIds(ids.elements);
-          imodel.selectionSet.replace(ids.elements);
-          shouldClearSelectionSet = false;
-        }
-        if (shouldClearSelectionSet) {
-          imodel.selectionSet.emptyAll();
-        }
-      });
+    using(Presentation.selection.suspendIModelToolSelectionSync(this._imodel), (_) => {
+      imodel.hilited.clear();
```
saskliutas, line 77: @grigasp What do you think about clearing hilited elements before requesting new hilite set? If you don't like this approach I will update `getHiliteSetIterator` implementation to emit empty hilite set if there are node keys.

packages/components/src/presentation-components/viewport/WithUnifiedSelection.tsx
```diff
@@ -34,19 +34,30 @@ export function viewWithUnifiedSelection<P extends ViewportProps>(
   const WithUnifiedSelection = memo<CombinedProps>((props) => {
     const { selectionHandler, ...restProps } = props;
     const imodel = restProps.imodel;
-    const [viewportSelectionHandler] = useState(() => selectionHandler ?? new ViewportSelectionHandler({ imodel }));
+    const [viewportSelectionHandler, setViewportSelectionHandler] = useState<ViewportSelectionHandler>();
 
     // apply currentSelection when 'viewportSelectionHandler' is initialized (set to handler from props or new is created)
     // 'viewportSelectionHandler' should never change because setter is not used.
```
grigasp, line 40: the comment needs update?

apps/test-app/frontend/src/components/app/App.tsx
```diff
@@ -77,13 +81,28 @@ export function App() {
...
+        .subscribe({
+          next: (set) => {
+            if (!set.elements) {
+              // note: the hilite list may contain models and subcategories as well - we don't
+              // care about them at this moment
+              return;
+            }
+
+            elementIds.push(...set.elements);
+            void selectedView.zoomToElements(elementIds);
```
grigasp, line 97: I'm trying to imagine how zooming to elements would look like when we load them in batches. And don't think that's a good idea. Have you tried that?

packages/components/src/presentation-components/viewport/ViewportSelectionHandler.ts
```diff
@@ -69,54 +63,42 @@ export class ViewportSelectionHandler implements IDisposable {
...
-    return this._asyncsTracker.pendingAsyncs;
+  public applyCurrentSelection() {
+    this.applyUnifiedSelection(this._imodel);
   }
 
-  public async applyCurrentSelection() {
-    await this.applyUnifiedSelection(this._imodel, { providerName: "" }, this.selectionHandler.getSelection());
-  }
+  private applyUnifiedSelection(imodel: IModelConnection) {
+    this._cancelOngoingChanges.next();
```
grigasp, line 74: should `dispose` also emit one of these?

packages/components/src/presentation-components/viewport/ViewportSelectionHandler.ts
grigasp: I think we have to consider the situation where user picks an element in the viewport. Here's what happens:
1. User picks an element.
2. The element is added to viewport's selection and hilite sets. The element gets hilited.
3. Unified selection is updated with the picked element (let's ignore selection scopes here).
4. We request a hilite set for the new selection, from now on - in batches.
5. On each batch, we apply it on viewport's selection and hilite sets to get element displayed.

For the user, steps 3-4 are invisible, so all he sees after picking an element is it getting hilited immediately in step 2 and, in cases when we get a larger hilite set, more elements getting hilited in a few moments. 

If we clear viewport's selection/hilite sets between steps 3 and 4, the user would see the picked element hilite, then de-hilite, and then hilite again. I don't think we want that.

I see the previous implementation would first get the full hilite set and then clear before applying the new one. With batched requests this isn't as simple as we want to update the viewport with new batch as soon as we receive it, but it isn't guaranteed that the first batch contains all initially picked elements (e.g. when user uses fence selection there could be more elements that we have in a single batch). However, I think we could say that it's guaranteed for initially picked elements to be a subset of the final hilite set. Maybe we don't have to clear anything at all then?

packages/components/src/test/viewport/WithUnifiedSelection.test.tsx
```diff
@@ -5,161 +5,184 @@
...
-
-    const newImodel = moq.Mock.ofType<IModelConnection>();
-    rerender(<PresentationViewport imodel={newImodel.object} viewDefinitionId={viewDefinitionId} selectionHandler={selectionHandlerMock.object} />);
+    const { rerender } = render(<PresentationViewport imodel={imodel} viewDefinitionId={viewDefinitionId} selectionHandler={selectionHandler} />);
+    expect(selectionHandler.imodel).to.be.eq(imodel);
 
-    selectionHandlerMock.verify((x) => (x.imodel = newImodel.object), moq.Times.atLeastOnce());
-  });
-
-  it("disposes ViewportSelectionHandler when unmounted", () => {
```
grigasp, line 84: do you think this is not worth checking?

packages/components/src/test/viewport/WithUnifiedSelection.test.tsx
```diff
@@ -171,262 +194,295 @@ describe("ViewportSelectionHandler", () => {
...
+        yield {
+          elements: [id],
+        } as HiliteSet;
+      }
+      selectionManager.getHiliteSetIterator.callsFake(() => generator());
 
       // trigger the selection change and wait for event handler to finish
       triggerSelectionChange();
-      await waitForAllAsyncs([handler]);
+      // await waitForAllAsyncs([handler]);
```
grigasp, line 289: needs cleanup? similar leftovers in other tests too

grigasp (changes requested):

packages/components/src/test/viewport/WithUnifiedSelection.test.tsx
saskliutas: I don't think `viewWithUnifiedSelection` should disposed `ViewportSelectionHandler` received through props because that won't work in React 18 `StrictMode`

grigasp: Agree. I guess there's no way to check if we dispose the handler in cases we create it ourselves?

saskliutas: Updated test to check that default handler is disposed [003ef54](https://github.com/iTwin/presentation/pull/354/commits/003ef54dd5ea40d772e61e39421d01195876aec5)

packages/components/src/presentation-components/viewport/ViewportSelectionHandler.ts
saskliutas: As discussed in private. Changed implementation to clear hilite set only when selection change event is of type `Replace` or `Clear`. For `Add` and `Remove` current implementation add/removes only affected elements to/from hilite set.

PR #356, opened 2023-12-04
Filtering dialog buttons improvement

ViliusBa:


packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -54,6 +58,19 @@ export interface PresentationInstanceFilterDialogProps {
...
+  onApplyHandler: () => void;
+  onCloseHandler: () => void;
+  onResetHandler: () => void;
+}
+
+/** Props required for rendering a toolbar */
+export interface ToolbarRendererProps {
+  toolbarHandlers: ToolbarHandlers;
+  getFilterInfo: (options?: BuildFilterOptions | undefined) => PresentationInstanceFilterInfo | undefined;
+}
```
grigasp, line 72: These are used in `@beta` APIs, so should also be `@beta`. Also, since they're tightly bound to the filtering dialog, the name should be changed to tell that.

.changeset/modern-pugs-yell.md
```diff
@@ -0,0 +1,8 @@
+---
+"@itwin/presentation-components": minor
+---
+
+- **Reset** button added which will clear all the filtering rules in the dialog.
+- Ability to press **Apply** button when no filtering rules are selected.
+- Clear hierarchy level filter if **Apply** is pressed when no filtering rules are selected.
+- `toolbarRenderer` added to introduce ability to render a custom toolbar in the filtering dialog.
```
grigasp, line 8:
```suggestion
- `PresentationInstanceFilterDialog` enhancements:
  - Changed the "Apply" button to always be enabled, even when no filtering rules are selected.
  - Added a "Reset" button which clears all the filtering rules in the dialog.
  - Added a `toolbarRenderer` prop to allow rendering a custom buttons' toolbar at the bottom of the dialog.

- Changed `PresentationTreeRenderer` to clear hierarchy level filter if filtering dialog was closed by clicking "Apply" with no filtering rules in the dialog.
```

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -163,11 +197,22 @@ function LoadedFilterDialogContent(props: LoadedFilterDialogContentProps) {
     [buildFilter, descriptor, filteringProps.selectedClasses],
   );
 
+  const onResetHandler = () => {
+    filteringProps.selectedClasses = [];
+    filteringProps.onSelectedClassesChanged([]);
+    rootGroup.items.forEach((item) => {
+      actions.removeItem([item.id]);
+    });
```
saskliutas, line 205: I don't think this is a good approach to clear filter. I think we need to add `clear` actions to `FilterBuilderActions`

ViliusBa: Agree, especially as this action is also going to be needed when class selection changes

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -31,9 +31,13 @@ export interface PresentationInstanceFilterDialogProps {
   /** Specifies whether dialog is open or not. */
   isOpen: boolean;
   /** Callback that is invoked when 'Apply' button is clicked. */
-  onApply: (filter: PresentationInstanceFilterInfo) => void;
+  onApply: (filter?: PresentationInstanceFilterInfo) => void;
   /** Callback that is invoked when 'Close' button is clicked or dialog is closed. */
   onClose: () => void;
+  /** Callback that is invoked when 'Reset' button is clicked. */
+  onReset: () => void;
```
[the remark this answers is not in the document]
ViliusBa: I kept it obligatory for the sake of keeping all of the callbacks the same. Also, it does nothing because of our implementation and it only so happens to be that we use the other two. Maybe, it would be a good idea to make them all optional. What do you think?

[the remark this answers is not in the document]
saskliutas: I think `onClose` can be made optional too. I don't think it makes sense to use this dialog without `onApply` so I would keep it required.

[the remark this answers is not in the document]
ViliusBa: By all, I mean the three button callbacks `onReset`, `onApply` and `onClose`

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -201,18 +226,36 @@ function LoadedFilterDialogContent(props: LoadedFilterDialogContentProps) {
...
+          )}
         </Dialog.ButtonBar>
       </div>
     </>
   );
 }
 
+function ToolbarButtonsRenderer({ handleApply, handleClose, handleReset }: FilteringDialogToolbarHandlers) {
+  return (
+    <Dialog.ButtonBar className="presentation-instance-filter-button-bar">
```
saskliutas, line 242: `Dialog.ButtonBar` can be removed it is added in `LoadedFilterDialogContent` component.

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -54,6 +58,16 @@ export interface PresentationInstanceFilterDialogProps {
   initialFilter?: PresentationInstanceFilterInfo;
 }
 
+/**
+ * Set of custom handlers that will additionally be called on each action respectively.
+ * @beta
+ */
+export interface FilteringDialogToolbarHandlers {
+  handleApply: () => void;
+  handleClose?: () => void;
```
saskliutas, line 67: These don't need to be optional. They are provided by `PresentationInstanceFilterDialog` and are always defined. (Atleast `handleApply` and `handleReset`). In case of `handleClose` I think `PresentationInstanceFilterDialog` could pass something like `() => { props.onClose && props.onClose() }`. This way API for customizing buttons will be cleaner for consumer as they won't need to make undefined checks.

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -54,6 +58,16 @@ export interface PresentationInstanceFilterDialogProps {
   initialFilter?: PresentationInstanceFilterInfo;
 }
 
+/**
+ * Set of custom handlers that will additionally be called on each action respectively.
```
saskliutas, line 62:
```suggestion
 * Set of action handlers that are passed to [[PresentationInstanceFilterDialogProps.toolbarButtonsRenderer]] for rendering custom buttons.
```

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -75,8 +75,9 @@ export function PresentationTreeRenderer(props: PresentationTreeRendererProps) {
         ? createPortal(
             <TreeNodeFilterBuilderDialog
               dataProvider={nodeLoader.dataProvider}
+              onReset={() => {}}
```
saskliutas, line 78: This can be removed.

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -95,20 +96,20 @@ interface TreeNodeFilterBuilderDialogProps {
   dataProvider: IPresentationTreeDataProvider;
   filterNode: FilterablePresentationTreeNodeItem;
   onClose: () => void;
-  onApply: (info: PresentationInstanceFilterInfo) => void;
+  onApply: (info?: PresentationInstanceFilterInfo) => void;
+  onReset: () => void;
```
saskliutas, line 100: This can be removed.

packages/components/src/test/instance-filter-builder/PresentationInstanceFilterDialog.test.tsx
```diff
@@ -74,7 +74,7 @@ describe("PresentationInstanceFilterDialog", () => {
   it("invokes 'onApply' with string property filter rule", async () => {
     const spy = sinon.spy();
     const { container, getByText, queryByDisplayValue, user } = render(
-      <PresentationInstanceFilterDialog imodel={imodel} descriptor={descriptor} onClose={() => {}} onApply={spy} isOpen={true} />,
+      <PresentationInstanceFilterDialog imodel={imodel} descriptor={descriptor} onClose={() => {}} onReset={() => {}} onApply={spy} isOpen={true} />,
```
saskliutas, line 77: `onClose` and `onReset` can be removed from tests where `() => {}` is passed to reduce noise.

saskliutas (changes requested):

saskliutas (approved):

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -201,18 +230,36 @@ function LoadedFilterDialogContent(props: LoadedFilterDialogContentProps) {
...
+        {translate("instance-filter-builder.cancel")}
+      </Button>
+      <Button className="presentation-instance-filter-dialog-reset-button" onClick={handleReset}>
+        {translate("instance-filter-builder.reset")}
+      </Button>
+    </>
+  );
+}
+
+const isFilterNonEmpty = (rootGroup: PropertyFilterBuilderRuleGroup) =>
```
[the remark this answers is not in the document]
ViliusBa: sure

PR #359, opened 2023-12-08
 Integrate stateless hierarchies builder into test app

### LINKED ISSUE #283: Integrate into test app
Need to integrate stateless hierarchies builder into our test app for testing it.

JonasDov:
closes #283


apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -52,14 +87,17 @@ export function TreeWidget(props: Props) {
     );
     setMatchesCount(newMatchesCount);
   }, []);
-
   const { width, height, ref } = useResizeDetector();
```
saskliutas, line 90: It should be enough to track size in `TreeWidget` component and pass it down to rules driven and stateless trees.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -32,16 +33,50 @@ import { Tree } from "./Tree";
...
   rulesetId?: string;
+  height?: number;
 }
 
 export function TreeWidget(props: Props) {
+  const [openTab, setOpenTab] = useState(0);
+  const { height, ref } = useResizeDetector();
+  const tabsClassName = "tree-widget-tabs";
+  const [heightOfTreeWidget, setHeightOfTreeWidget] = useState(0);
+  const tabElements = document.getElementsByClassName(tabsClassName);
```
saskliutas, line 44: Don't use this to access native html element. Use `useRef(null)` instead.

saskliutas (changes requested):

grigasp (changes requested):
Please also update `parseTreeNodeItem` function to assign id like this:
```
id: JSON.stringify([...node.parentKeys, node.key]),
```
This should make sure we don't get any duplicate node ids.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Removed useResizeDetector usage from rules driven and stateless trees.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: `<Tabs>`  and `<Tab>` elements don't have `ref` property, so `useRef` can't be used with them. You can set useRef for each `<div className="filteredTree">`, but then, when trying to resize the right panel, "filteredTree" element height gets updated very slowly.
Changed to:
`ref.current?.getElementsByClassName(tabsClassName);` and moved it to the `useEffect` block.

JonasDov:
> Please also update `parseTreeNodeItem` function to assign id like this:
> 
> ```
> id: JSON.stringify([...node.parentKeys, node.key]),
> ```
> 
> This should make sure we don't get any duplicate node ids.

Changed

saskliutas (approved):

grigasp (approved):

PR #366, opened 2023-12-21
Option to override hierarchy level size limit for each hierarchy level

### LINKED ISSUE #336: Hierarchy level size limit should be overrideable individually for each hierarchy level

JonasDov:
Closes #336


packages/hierarchy-builder/src/hierarchy-builder/HierarchyDefinition.ts
```diff
@@ -110,6 +110,8 @@ export interface DefineHierarchyLevelProps {
   parentNode: HierarchyDefinitionParentNode | undefined;
   /** Optional hierarchy level filter. */
   instanceFilter?: GenericInstanceFilter;
+  /** Optional hierarchy level size limit. Default limit is 1000. */
+  hierarchyLevelSizeLimit?: number;
```
grigasp, line 114: Why do we need this attribute on `DefineHierarchyLevelProps`?

packages/hierarchy-builder/src/hierarchy-builder/ErrorTypeChecker.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Namespace which can be used to determine error type.
+ * @beta
+ */
+export namespace ErrorTypeChecker {
+  /** Checks if the error is a RowsLimitExceededError. */
+  export function isRowsLimitExceededError(err: any): err is RowsLimitExceededError {
+    return err instanceof RowsLimitExceededError;
+  }
+}
```
grigasp, line 17: I'm not sure we need this... Calling `ErrorTypeChecker.isRowsLimitExceededError(error)` doesn't simplify anything compared to `error instanceof RowsLimitExceededError` call. Are there ant other reasons to have it?

packages/hierarchy-builder/src/hierarchy-builder/internal/TreeNodesReader.ts
```diff
@@ -26,17 +25,17 @@ export class TreeQueryResultsReader {
     // istanbul ignore next
     this._props = {
       parser: props?.parser ?? defaultNodesParser,
-      limit: props?.limit ?? DEFAULT_ROWS_LIMIT,
     };
   }
 
-  public async read(executor: IECSqlQueryExecutor, query: Omit<ECSqlQueryDef, "ctes">): Promise<ParsedHierarchyNode[]> {
+  public async read(executor: IECSqlQueryExecutor, query: Omit<ECSqlQueryDef, "ctes">, limit?: number): Promise<ParsedHierarchyNode[]> {
+    const nodeLimit = limit ?? DEFAULT_ROWS_LIMIT;
```
grigasp, line 32: We apply the limit on the query in `applyLimit` which is called by `HierarchyProvider` before sending the result here. Instead, I think `HierarchyProvider` should send the full `ECSqlQueryDef` here and `applyLimit` should be called from there too.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -172,21 +177,32 @@ export function StatelessTreeWidget(props: Omit<Props, "rulesetId">) {
       const parent: HierarchyNode | undefined = node ? (node as any).__internal : undefined;
       try {
         if (modelsTreeHierarchyProvider) {
-          return (await modelsTreeHierarchyProvider.getNodes({ parentNode: parent })).map(parseTreeNodeItem);
+          return (await modelsTreeHierarchyProvider.getNodes({ parentNode: parent, hierarchyLevelSizeLimit })).map(parseTreeNodeItem);
         }
         return [];
       } catch (e) {
         // eslint-disable-next-line no-console
         console.error(e);
```
grigasp, line 185: the error should not be logged if it gets handled

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -273,3 +290,51 @@ function parseTreeNodeItem(node: HierarchyNode): DelayLoadedTreeNodeItem {
...
+        isLeaf={true}
+        label={
+          <span>
+            <Text isMuted>
+              <span>{nodeItem.message}</span>
+              {nodeItem.type === InfoTreeNodeItemType.ResultSetTooLarge && (
+                <span>
+                  <span> - </span>
+                  <UnderlinedButton onClick={() => props.onLimitReset()}>
+                    {` ${IModelApp.localization.getLocalizedString("Sample:controls.reset-hierarchy-level-limit")}.`}
```
grigasp, line 326: why is there a space before the localized string?

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -172,21 +177,32 @@ export function StatelessTreeWidget(props: Omit<Props, "rulesetId">) {
...
     };
-  }, [modelsTreeHierarchyProvider]);
+  }, [modelsTreeHierarchyProvider, hierarchyLevelSizeLimit]);
   const modelSource = useTreeModelSource(dataProvider);
   const nodeLoader = useTreeNodeLoader(dataProvider, modelSource);
   const eventHandler = useMemo(() => new TreeEventHandler({ nodeLoader, modelSource }), [nodeLoader, modelSource]);
   const treeModel = useTreeModel(modelSource);
-
+  const nodeRenderer = useCallback((nodeProps: TreeNodeRendererProps) => {
+    return <StatelessTreeNodeRenderer {...nodeProps} onLimitReset={() => setHierarchyLevelSizeLimit(undefined)} />;
```
grigasp, line 204: I don't think this works as discussed. Here's how it should work:
1. By default we send all hierarchy level requests without limit override (and it defaults to 1000).
2. In case of "result set too large" error, we should render a custom node, which offers **to remove the limit for that particular hierarchy level**. Note that removing the limit means there is not limit rather than the default limit.

grigasp (changes requested):

packages/hierarchy-builder/src/hierarchy-builder/HierarchyDefinition.ts
JonasDov: `DefineHierarchyLevelProps` is the type of props in `setupObservables`. Then it is passed to `createPreProcessedNodesObservable` where `hierarchyLevelSizeLimit` attribute is passed to `applyLimit` and `_queryReader.read` functions. I just assumed that I can change `DefineHierarchyLevelProps`.
Changed so it does not have this attribute. Now `setupObservables` and `createPreProcessedNodesObservable` have props of type: `DefineHierarchyLevelProps & { hierarchyLevelSizeLimit?: number | "unbounded" }`.

packages/hierarchy-builder/src/hierarchy-builder/internal/TreeNodesReader.ts
JonasDov: Moved `applyLimit` call to `read` function.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Removed.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Added option to set `hierarchyLevelSizeLimit` to "unbounded" which allows this functionality.

packages/hierarchy-builder/src/hierarchy-builder/ErrorTypeChecker.ts
JonasDov: Rows limit exceeded error was in the hierarchy-builder/internal folder, and was marked with `@internal` tag. Because of this, we weren't able to import it in test-app and I assumed that I shouldn't change it.
Changed so we have `HierarchyErrors.ts` file, which exports `RowsLimitExceededError`.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Moved.

packages/hierarchy-builder/src/hierarchy-builder/internal/TreeNodesReader.ts
grigasp: A couple of problems with the change:
- At the moment `applyLimit` is responsible for serializing `query.ctes` into the resulting ECSQL. However, the `query` argument of this function omits `ctes`, so I assume they never get included into the ECSQL?
- Mutating input argument is not a great idea (`query.ecsql = applyLimit(...)`), as that may cause unexpected consequences to the callers.

My suggestions:
1. Change `applyLimit` to only take ECSQL without CTEs. Leave serializing CTEs into the ECSQL for the caller.
2. In the `read` function:
  - Change `query` argument to not omit `ctes` attribute.
  - Serialize `ctes` and `ecsql` attributes of the given query into a local `ecsql` variable.
  - Pass the local `ecsql` to `applyLimit` to get the limit applied.

packages/hierarchy-builder/src/hierarchy-builder/HierarchyProvider.ts
```diff
@@ -92,6 +93,8 @@ export interface GetHierarchyNodesProps {
   parentNode: ParentHierarchyNode | undefined;
   /** Optional hierarchy level filter. */
   instanceFilter?: GenericInstanceFilter;
+  /** Optional hierarchy level size limit. Default limit is 1000. */
```
grigasp, line 96:
```suggestion
  /** Optional hierarchy level size limit. Default limit is `1000`. */
```

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -273,3 +336,56 @@ function parseTreeNodeItem(node: HierarchyNode): DelayLoadedTreeNodeItem {
     autoExpand: node.autoExpand,
   } as DelayLoadedTreeNodeItem;
 }
+
+function createInfoNode(parentNode: TreeNodeItem | undefined, message: string, type?: InfoTreeNodeItemType): PresentationInfoTreeNodeItem {
+  const id = parentNode ? `${parentNode.id}/info-node` : `/info-node/${message}`;
```
grigasp, line 341: why aren't we including `message` if there's a `parentNode`?

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -273,3 +336,56 @@ function parseTreeNodeItem(node: HierarchyNode): DelayLoadedTreeNodeItem {
...
+            <Text isMuted>
+              <span>{nodeItem.message}</span>
+              {nodeItem.type === InfoTreeNodeItemType.ResultSetTooLarge && (
+                <span>
+                  <span> - </span>
+                  <UnderlinedButton
+                    onClick={() => {
+                      const parsedParentIds = nodeItem.parentId ? JSON.parse(nodeItem.parentId) : undefined;
+                      const parentId = parsedParentIds ? parsedParentIds[parsedParentIds.length - 1].instanceKeys[0].id : undefined;
+                      props.onLimitReset(parentId);
```
grigasp, line 374: can you explain what's going on here?

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -170,23 +182,73 @@ export function StatelessTreeWidget(props: Omit<Props, "rulesetId">) {
...
+  const nodeRenderer = useCallback((nodeProps: TreeNodeRendererProps) => {
+    return (
+      <StatelessTreeNodeRenderer
+        {...nodeProps}
+        onLimitReset={(parentId?: string) =>
+          setHierarchyLevelSizeLimit((map) => {
+            const newMap = new Map(map);
+            newMap.set(parentId, { limit: "unbounded" });
+            return newMap;
+          })
```
grigasp, line 247: I think we could store those custom limits in a plain object rather than a `Map`, e.g. `useState<{[parentId: string]: number | "unbounded"}>({})`. Updating it here would be simpler too:
```ts
setHierarchyLevelSizeLimit((map) => ({ ...map, [parentId ?? ""]: "unbounded" }))
```

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -170,23 +182,73 @@ export function StatelessTreeWidget(props: Omit<Props, "rulesetId">) {
...
+          } else {
+            lastNonGroupingNodeId = nodeKey.instanceKeys[0].id;
+          }
+        }
+      }
+      // Get limit that is set for lastNonGroupingNodeId
+      const mapKeyValue = hierarchyLevelSizeLimit.get(lastNonGroupingNodeId);
+      let limit: undefined | number | "unbounded";
+      if (mapKeyValue === undefined) {
+        hierarchyLevelSizeLimit.set(lastNonGroupingNodeId, { limit });
```
grigasp, line 209: Do we need to set it here? I don't think it's a good idea to change state from `useMemo` which has a dependency on that state variable.

grigasp (commented):
@saskliutas, can you please have a look at this too? In `TreeWidget.tsx` I'm not sure about re-creating the data provider on each hierarchy level size limit change. Maybe the limits should be stored in a ref rather than state?

packages/hierarchy-builder/src/hierarchy-builder/internal/TreeNodesReader.ts
JonasDov: Changed to your suggestions with a small difference:
Your suggestion: Serialize `ctes` and `ecsql` attributes of the given query into a local `ecsql` variable -> pass it to `applyLimit`.
My change: Pass provided `ecsql` variable to `applyLimit` -> combine serialized `ctes` and string from `applyLimit` and pass the value to `queryReader`
Made a different change than your suggestion because:
applyLimit currently wraps the ecsql inside another select clause:
image
And doing this with ctes added to the ecsql doesn't work. If you want your suggestion to work, then applyLimit shouldn't wrap the query inside another select clause, and should be implemented like: 
image
Which approach should I use?

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Including `message` in both cases.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: `nodeItem.parentId` contains values like:
```json
"[{\"type\":\"instances\",\"instanceKeys\":[{\"className\":\"BisCore.Subject\",\"id\":\"0x1\"}]},{\"type\":\"instances\",\"instanceKeys\":[{\"className\":\"BisCore.PhysicalModel\",\"id\":\"0x1c\"}]},{\"type\":\"instances\",\"instanceKeys\":[{\"className\":\"BisCore.SpatialCategory\",\"id\":\"0x17\"}]}]"
```
I was parsing it and extracting only the last parent Id (would be "0x17" in the example above). Then we were changing map of key(parent id): value (level limit) pairs. This map was later used before running the `getNodes(...)` to extract limit of specific node.

JonasDov: Changed the implementation to not parse parent id, but instead use the exact `parentId` in `[parentId: string]: number | "unbounded" | undefined}`. Now `ChildNodesCache` doesn't include limit into the `key` if `parentNode` is a grouping node.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Changed to your suggestion.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: It's a mistake, should not be here at all.
Removed.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Should I document that `getNodes(...)` function doesn't use provided limit when `parentNode` is a `GroupingNode`?

grigasp: Looks good. As discussed, let's also change `GetHierarchyNodesProps` to not have `instanceFilter` and `hierarchyLevelSizeLimit` attributes for grouping nodes.

grigasp (changes requested):
Filed an issue for not loosing tree state when changing size limit for specific hierarchy levels: https://github.com/iTwin/presentation/issues/376.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -170,23 +175,45 @@ export function StatelessTreeWidget(props: Omit<Props, "rulesetId">) {
...
       }
     };
-  }, [modelsTreeHierarchyProvider]);
+  }, [modelsTreeHierarchyProvider, hierarchyLevelSizeLimit]);
   const modelSource = useTreeModelSource(dataProvider);
   const nodeLoader = useTreeNodeLoader(dataProvider, modelSource);
   const eventHandler = useMemo(() => new TreeEventHandler({ nodeLoader, modelSource }), [nodeLoader, modelSource]);
   const treeModel = useTreeModel(modelSource);
-
+  const nodeRenderer = useCallback((nodeProps: TreeNodeRendererProps) => {
```
saskliutas, line 209: No need for `useCallback`.

saskliutas (approved):
`TreeWidget.tsx` looks good now. `StatelessTreeWidget` code grown quite and I think it could be moved to separate file.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Tried doing that, but as discussed keeping as is. Added a comment for `instanceFilter` and `hierarchyLevelSizeLimit`: 
```
Isn't applied if `parentNode` is a [[GroupingNode]].
```

packages/hierarchy-builder/src/hierarchy-builder/internal/TreeNodesReader.ts
```diff
@@ -107,15 +108,12 @@ const DEFAULT_ROWS_LIMIT = 1000;
...
   limit?: number | "unbounded";
 }
 
 /** @internal */
 export function applyLimit(props: ApplyLimitProps) {
-  const ctesPrefix = props.ctes && props.ctes.length ? `WITH RECURSIVE ${props.ctes.join(", ")}` : ``;
   return `
-    ${ctesPrefix}
     SELECT *
     FROM (${props.ecsql})
```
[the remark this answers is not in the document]
JonasDov: Changed.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
```diff
@@ -182,33 +175,11 @@ export function StatelessTreeWidget(props: Omit<Props, "rulesetId">) {
...
-      // Get limit that is set for lastNonGroupingNodeId
-      const mapKeyValue = hierarchyLevelSizeLimit.get(lastNonGroupingNodeId);
+      const parentId = node?.id;
+      const levelLimit = hierarchyLevelSizeLimit[parentId ?? ""];
       let limit: undefined | number | "unbounded";
-      if (mapKeyValue === undefined) {
-        hierarchyLevelSizeLimit.set(lastNonGroupingNodeId, { limit });
-      } else {
-        limit = mapKeyValue.limit;
+      if (levelLimit !== undefined) {
```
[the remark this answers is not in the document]
JonasDov: Changed.

apps/test-app/frontend/src/components/tree-widget/TreeWidget.tsx
JonasDov: Removed `useCallback`.

packages/hierarchy-builder/src/hierarchy-builder/HierarchyProvider.ts
```diff
@@ -91,9 +91,9 @@ export interface HierarchyProviderProps {
 export interface GetHierarchyNodesProps {
   /** Parent node to get children for. Pass `undefined` to get root nodes. */
   parentNode: ParentHierarchyNode | undefined;
-  /** Optional hierarchy level filter. */
+  /** Optional hierarchy level filter. Isn't applied if `parentNode` is a [[GroupingNode]]. */
```
grigasp, line 94:
```suggestion
  /** Optional hierarchy level filter. Has no effect if `parentNode` is a [[GroupingNode]]. */
```

packages/hierarchy-builder/src/hierarchy-builder/HierarchyProvider.ts
```diff
@@ -91,9 +91,9 @@ export interface HierarchyProviderProps {
...
   instanceFilter?: GenericInstanceFilter;
-  /** Optional hierarchy level size limit. Default limit is `1000`. */
+  /** Optional hierarchy level size limit. Default limit is `1000`. Isn't applied if `parentNode` is a [[GroupingNode]]. */
```
grigasp, line 96:
```suggestion
  /** Optional hierarchy level size limit. Default limit is `1000`. Has no effect if `parentNode` is a [[GroupingNode]]. */
```

grigasp (approved):

PR #371, opened 2023-12-22
Clear all filtering options when class selection is edited

ViliusBa:
Clear all filtering options when selected class list is changed.


packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
```diff
@@ -118,9 +124,17 @@ export function usePresentationInstanceFilteringProps(
     [propertyInfos, filterClassesByProperty],
   );
 
+  const onSelectedClassesChanged = useCallback(
+    (classIds: string[]) => {
+      changeActiveClasses(classIds);
+      actions?.removeAllItems();
```
saskliutas, line 130: I think this should be done in `LoadedFilterDialogContent` where `usePresentationInstanceFilteringProps` is called.

saskliutas (changes requested):

ViliusBa:
One use case scenario that concerns me: 
- User selects classes.
- User adds property filtering rules.
- User adds an additional class that supports all the properties that are currently in the filter.
- Filtering rules are cleared.

It seems to me that clearing all filtering property rules every time selected class list changes might not be the behavior that an user expects.

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
ViliusBa: Changed the implementation based on your suggestion

grigasp:
> One use case scenario that concerns me:
> 
> * User selects classes.
> * User adds property filtering rules.
> * User adds an additional class that supports all the properties that are currently in the filter.
> * Filtering rules are cleared.
> 
> It seems to me that clearing all filtering property rules every time selected class list changes might not be the behavior that an user expects.

I believe we discussed that a warning / confirmation from user is needed before clearing the properties filter on a class list change. Maybe we could also clear the property filter only when a class is removed, but keep it if a class is added? @saskliutas, do you see anything wrong with that?

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
```diff
@@ -47,34 +48,45 @@ export interface InstanceFilterBuilderProps extends PropertyFilterBuilderRendere
 export function InstanceFilterBuilder(props: InstanceFilterBuilderProps) {
   const { selectedClasses, classes, onSelectedClassesChanged, imodel, descriptor, descriptorInputKeys, ...restProps } = props;
 
+  const [showClearAlert, setShowClearAlert] = useState(false);
   const options = useMemo(() => classes.map(createOption), [classes]);
   const selectedOptions = useMemo(() => selectedClasses.map((classInfo) => classInfo.id), [selectedClasses]);
 
+  // style={{ minWidth: "min(100%, 350px)", marginBottom: "8px" }}
```
grigasp: needs cleanup

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.tsx
```diff
@@ -47,34 +48,45 @@ export interface InstanceFilterBuilderProps extends PropertyFilterBuilderRendere
 export function InstanceFilterBuilder(props: InstanceFilterBuilderProps) {
   const { selectedClasses, classes, onSelectedClassesChanged, imodel, descriptor, descriptorInputKeys, ...restProps } = props;
 
+  const [showClearAlert, setShowClearAlert] = useState(false);
```
grigasp, line 51: maybe rename to `showClassSelectionWarning`?

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.scss
```diff
@@ -14,3 +14,8 @@
     overflow: auto;
   }
 }
+
+.class-selection-warning{
```
grigasp, line 18:
```suggestion
.class-selection-warning {
```

packages/components/public/locales/en/PresentationComponents.json
```diff
@@ -29,7 +29,8 @@
     "error-messages": {
       "not-a-number": "Value must be a number",
       "invalid": "Value is invalid"
-    }
+    },
+    "class-selection-warning": "All filtering rules will be lost if class list is edited."
```
grigasp, line 33:
```suggestion
    "class-selection-warning": "Changing the class list clears all property filtering rules."
```

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -256,7 +263,7 @@ function ToolbarButtonsRenderer({ handleApply, handleClose, handleReset }: Filte
   );
 }
 
-function isFilterNonEmpty(rootGroup: PropertyFilterBuilderRuleGroup) {
+export function isFilterNonEmpty(rootGroup: PropertyFilterBuilderRuleGroup) {
```
grigasp, line 266: needs `@internal` tag?

grigasp (changes requested):

grigasp (approved):

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -257,7 +263,8 @@ function ToolbarButtonsRenderer({ handleApply, handleClose, handleReset }: Filte
   );
 }
 
-function isFilterNonEmpty(rootGroup: PropertyFilterBuilderRuleGroup) {
+/** @internal */
+export function isFilterNonEmpty(rootGroup: PropertyFilterBuilderRuleGroup) {
```
saskliutas, line 267: I would suggest moving this to `Utils.ts` file to avoid exporting it through barrel file.

packages/components/src/presentation-components/instance-filter-builder/InstanceFilterBuilder.scss
```diff
@@ -14,3 +14,8 @@
     overflow: auto;
   }
 }
+
+.class-selection-warning {
+  min-width: 100%;
+  margin-bottom: 8px;
```
saskliutas, line 20: Can we use ITwinUI variable for this? Maybe `iui-size-m` or `iui-size-s`?

saskliutas (approved):

PR #374, opened 2024-01-02
Filter Dialog: Fix for faulty error messages in number selector

### LINKED ISSUE #367: Instance Filtering Dialog: false-positive error messages on number selector
Selecting one of the provided values produces a false-positive error message

Image

ViliusBa:
Closes #367

Additionally check if we are able to parse a number by deserializing the value that is being validated.


packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -221,5 +221,24 @@ function isPropertyNumeric(typename: string) {
...
+  let isInvalid = false;
+  deserialized.groupedRawValues &&
+    deserialized.groupedRawValues.forEach((element: Value[]) => {
+      if (isNaN(Number(element))) {
+        isInvalid = true;
+        return;
+      }
+    });
+
+  return deserialized.groupedRawValues && isInvalid;
```
grigasp, line 243: This is handling our hack for implementing the `IN` operator without introducing braking changes into the `components-react` package by setting the value to a serialized JSON object. Please add a comment here explaining that.

packages/components/src/test/instance-filter-builder/Utils.test.ts
```diff
@@ -155,7 +155,7 @@ describe("filterRuleValidator", () => {
     sinon.restore();
   });
 
-  it("returns error message for invalid numeric rule", () => {
+  it(`returns error message for undefined`, () => {
```
grigasp, line 158: It seems to me that the changed test actually tests a different test case than what the previous test did. I think we should keep the old one and add a new one.

grigasp: Same applies for the changed tests below

packages/components/src/test/instance-filter-builder/Utils.test.ts
```diff
@@ -165,42 +165,67 @@ describe("filterRuleValidator", () => {
...
+    ).to.be.eq("instance-filter-builder.error-messages.invalid");
+  });
+
+  [
+    { val: 10, testCase: "number" },
+    { val: [10], testCase: "number array" },
+    { val: [[10]], testCase: "nested number array" },
+    { val: "[10]", testCase: "number array as string" },
+    { val: "[[10]]", testCase: "nested number array as string" },
+    { val: "[[10], [10]]", testCase: "nested number arrays as string" },
```
grigasp, line 212: Could the test names be more specific about what `PropertyValue` type they test? E.g. I'm not sure which `PropertyValue` type case is covered with the `number array as string` test.

grigasp (changes requested):

packages/components/src/test/instance-filter-builder/Utils.test.ts
ViliusBa: Not really sure if I understood this correctly, please check the latest commit and let me know if it still does not look right.

grigasp: The change did not solve my concern.

`filterRuleValidator` validates a `PropertyFilterBuilderRule`, which contains a `PropertyValue`. Here we're testing if the function correctly handles different types of its subtype - `PrimitiveValue`. The `value` part of `PrimitiveValue`, which varies in this test, is defined as `Primitives.Value`, which is defined as `Text | String | ShortDate | Boolean | Numeric | Enum | Point | Composite | InstanceKey`. I think the test cases should be based on those sub-types. 

For example, I'm not sure which of those sub-types is represented by `{ val: [10], testCase: "number array" }` test case. Maybe `Point`? In that case I'd like the test to be named "Point" or something like that.

packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -221,5 +221,26 @@ function isPropertyNumeric(typename: string) {
...
+    return true;
+  }
+
+  if (typeof value.value !== "string") {
+    return isNaN(Number(value.value));
+  }
+
+  // This is handling our implementation of the IN operator without introducing breaking changes into the "@itwin/components-react".
+  // That is why we need to check for values in a serialized JSON object.
+  const { groupedRawValues } = deserializeDisplayValueGroupArray(value.displayValue, value.value);
```
saskliutas, line 234: I think this should be done in `numericPropertyValidator` and only when operator is `Equals` or `NotEquals`

packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -221,5 +221,26 @@ function isPropertyNumeric(typename: string) {
...
+
+  let isInvalid = false;
+  groupedRawValues &&
+    groupedRawValues.forEach((element: Value[]) => {
```
saskliutas, line 238: I think `for of` loop is more fitting in this case as it can handle elements only up to invalid instead of always iterating all elements.

packages/components/src/test/instance-filter-builder/Utils.test.ts
```diff
@@ -171,36 +171,63 @@ describe("filterRuleValidator", () => {
...
+
+  [
+    { val: 10, testCase: "number" },
+    { val: 10.003, testCase: "float" },
+    { val: [10], testCase: "number array" },
+    { val: [[10]], testCase: "nested number array" },
+    { val: "[10]", testCase: "serialized array of numbers" },
+    { val: "{10}", testCase: "serialized struct" },
+    { val: "[10.003]", testCase: "serialized array of float numbers" },
+    { val: "[[10], [10]]", testCase: "serialized array of arrays of numbers" },
```
saskliutas, line 214: I don't think we need all these cases. We need ones that check if numbers are handled correctly and one that handles serialize grouped values produces by `UniquValuesSelector`.

saskliutas (changes requested):

packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -202,24 +202,49 @@ function quantityPropertyValidator({ property, value }: ValidatorContext) {
...
 }
 
-function numericPropertyValidator({ property, value }: ValidatorContext) {
+function numericPropertyValidator({ property, value, operator }: ValidatorContext) {
   if (!isPropertyNumeric(property.typename) || value === undefined) {
     return undefined;
   }
 
-  if (isInvalidNumericValue(value)) {
+  if (isInvalidNumericValue(value) && (!isOperatorValidForSerialization(operator) || isInvalidSerializedNumericValue(value))) {
```
saskliutas, line 210: Looking at this condition I think `numericPropertyValidator` should skip `Equal` and `NotEqual` operators. For those operators we show unique values selector. I don't think it makes sense to show not a number error message for unique values selector because user cannot edit values shown and if some of those values are not valid numbers than there is some problems in our code that makes us load invalid values for numeric properties.

saskliutas (changes requested):

packages/components/src/presentation-components/instance-filter-builder/Utils.ts
ViliusBa: That makes a lot of sense, will adjust

packages/components/src/test/instance-filter-builder/Utils.test.ts
```diff
@@ -187,13 +203,13 @@ describe("filterRuleValidator", () => {
     ).to.be.eq("instance-filter-builder.error-messages.invalid");
   });
 
-  it("does not return error message for valid numeric rule", () => {
+  it(`does not return error message for valid numeric rule value set as number`, () => {
```
saskliutas, line 206:
```suggestion
  it("does not return error message for valid numeric rule", () => {
```

packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -202,24 +202,30 @@ function quantityPropertyValidator({ property, value }: ValidatorContext) {
...
+  // don't validate values that we supply ourselves
+  if (isInvalidNumericValue(value) && !isValuePickedFromSelector(operator)) {
     return translate("instance-filter-builder.error-messages.not-a-number");
   }
 
   return undefined;
 }
 
+/** If `IsEqual` or `IsNotEqual` operators are set the value will be supplied from the `UniquePropertyValuesSelector`. */
+function isValuePickedFromSelector(operator: PropertyFilterRuleOperator) {
```
saskliutas, line 219:
```suggestion
function isEqualityOperator(operator: PropertyFilterRuleOperator) {
```

packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -202,24 +202,30 @@ function quantityPropertyValidator({ property, value }: ValidatorContext) {
...
 
-function numericPropertyValidator({ property, value }: ValidatorContext) {
+function numericPropertyValidator({ property, value, operator }: ValidatorContext) {
   if (!isPropertyNumeric(property.typename) || value === undefined) {
     return undefined;
   }
 
-  if (isInvalidNumericValue(value)) {
+  // don't validate values that we supply ourselves
+  if (isInvalidNumericValue(value) && !isValuePickedFromSelector(operator)) {
```
saskliutas, line 211: `!isValuePickedFromSelector` condition can be moved to `if` on line 206.

.changeset/tiny-otters-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Fix for false negative error messages in filter dialog number selector
```
saskliutas, line 5:
```suggestion
`PresentationInstanceFilterDialog`: Fix false positive error message shown when selecting numeric values in unique values selector.
```

saskliutas (changes requested):

saskliutas (approved):

packages/components/src/test/instance-filter-builder/Utils.test.ts
ViliusBa: Yeah, now I understand what you meant about `PropertyValue` types. Initially, I thought you were talking about the types  `PrimitiveValue | StructValue | ArrayValue`. 

But now that we've simplified tests and implementation, I don't think these extra test cases are needed.

.changeset/tiny-otters-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+`PresentationInstanceFilterDialog`: Fix false positive error messagesshown when selecting numeric values in unique values selector.
```
grigasp, line 5:
```suggestion
`PresentationInstanceFilterDialog`: Fix false positive error messages shown when selecting numeric values in unique values selector.
```

grigasp (approved):

PR #381, opened 2024-01-19
Hierarchy builder: Unified selection support

grigasp:
Closes https://github.com/iTwin/presentation/issues/360


apps/test-app/frontend/src/components/tree-widget/stateless/UnifiedSelectionTreeEventHandler.ts
```diff
@@ -0,0 +1,258 @@
...
+  imodel: IModelConnection;
+
+  /** Node loader used to load children when node is expanded. */
+  nodeLoader: AbstractTreeNodeLoader;
+
+  /** Specifies whether children should be disposed when parent node is collapsed or not. */
+  collapsedChildrenDisposalEnabled?: boolean;
+}
+
+export class UnifiedSelectionTreeEventHandler extends TreeEventHandler implements IDisposable {
```
saskliutas, line 37: It looks like it should be enough to extend `[UnifiedSelectionTreeEventHandler](https://github.com/iTwin/presentation/blob/9d0bad0c74e37d07401b418779abcd0c4c63c5ac/packages/components/src/presentation-components/tree/controlled/UseUnifiedSelection.ts#L68)` and override `createKeysForSelection`, `shouldSelectNode` methods instead of copying whole implementation.

grigasp: The primary reason I didn't reuse [UnifiedSelectionTreeEventHandler](https://github.com/iTwin/presentation/blob/9d0bad0c74e37d07401b418779abcd0c4c63c5ac/packages/components/src/presentation-components/tree/controlled/UseUnifiedSelection.ts#L68) here is that it wants an `AbstractTreeNodeLoaderWithProvider<IPresentationTreeDataProvider>`, which we don't have. We could extract common behavior into a base class and share it, I don't think that's worth the effort, having in mind that all tree-related APIs for stateless will need to written from scratch in components-agnostic way.

Issue #382, opened 2024-01-19
Limiting the amount of hierarchy filter matches


grigasp:
Similar to limiting the number of nodes we show in a single hierarchy level, we should also limit the number of search results.

PR #383, opened 2024-01-22
Hierarchy builder: Add results count limiting to hierarchy filtering

grigasp:
Closes https://github.com/iTwin/presentation/issues/382


packages/hierarchy-builder/src/test/HierarchyProvider.test.ts
```diff
@@ -147,28 +149,20 @@ describe("HierarchyProvider", () => {
       });
 
       const queryTimeout1 = new ResolvablePromise();
-      queryExecutor.createQueryReader.onFirstCall().returns({
-        async *[Symbol.asyncIterator](): AsyncIterableIterator<ECSqlQueryRow> {
-          await queryTimeout1;
-        },
-      });
+      queryExecutor.createQueryReader.onFirstCall().returns(eachValueFrom<ECSqlQueryRow>(from(queryTimeout1).pipe(mergeMap(() => EMPTY))));
```
saskliutas, line 152: I think it should be possible to replace `from(queryTimeout1).pipe(mergeMap(() => EMPTY)))` with Subject<ECSqlQueryRow>. Maybe it would make this test a little bit easier to understand?

packages/hierarchy-builder/src/test/Utils.ts
```diff
@@ -241,11 +250,5 @@ export async function waitFor(check: () => void, timeout?: number) {
...
 
 export function createFakeQueryReader(rows: object[]): ECSqlQueryReader {
-  return {
-    async *[Symbol.asyncIterator](): AsyncIterableIterator<ECSqlQueryRow> {
-      for (const row of rows) {
-        yield row as ECSqlQueryRow;
-      }
-    },
-  };
+  return eachValueFrom(from(rows));
```
saskliutas, line 253: Should our tests depend on `rxjs` + `rxjs-for-await`. Maybe it would be better to have simple functions `toAsyncIterableIterator` similar to `toArray`. I guess `eachValueFrom(from(rows))` as one liner looks clean but I feel it adds unnecessary diffuculty by converting `Array` -> `Observable` -> `AsyncIterableIterator`.

packages/full-stack-tests/src/hierarchy-builder/models-tree/ModelsTreeFiltering.test.ts
```diff
@@ -798,7 +805,11 @@ describe("Stateless hierarchy builder", () => {
       const actualInstanceKeyPaths = (
         await ModelsTreeDefinition.createInstanceKeyPaths({
           metadataProvider,
-          queryExecutor: createECSqlQueryExecutor(imodel),
+          queryExecutor: {
```
saskliutas, line 808: Why can't this test use `createModelsTreeProvider` and `limitingQueryExecutor`?

saskliutas (changes requested):

packages/hierarchy-builder/src/test/HierarchyProvider.test.ts
grigasp: made a change to stop using `rxjs` for this altogether

packages/hierarchy-builder/src/test/Utils.ts
grigasp: made a change to stop using `rxjs` for this altogether

saskliutas (approved):

PR #397, opened 2024-02-05
Change child nodes cache to use Dictionary instead of Map

JonasDov:
`ChildNodesCache` was using Map with key of type string. We were using JSON.stringify() to create keys from needed properties, this would take up a lot of memory. Changed to use `HierarchyNodeKey` array as the key for the Dictionary.


packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
```diff
@@ -228,6 +228,101 @@ export namespace HierarchyNodeKey {
...
+          const lhsInstanceKey = lhs.instanceKeys[i];
+          const rhsInstanceKey = rhs.instanceKeys[i];
+          const classNameComparResult = compareStrings(lhsInstanceKey.className, rhsInstanceKey.className);
+          if (classNameComparResult !== 0) {
+            return classNameComparResult;
+          }
+          const idCompareResult = compareStrings(lhsInstanceKey.id, rhsInstanceKey.id);
+          if (idCompareResult !== 0) {
+            return idCompareResult;
+          }
```
grigasp, line 274: Let's move this to [InstanceKey namespace](https://github.com/iTwin/presentation/blob/53893a975dce944a356ff7597103466c0fa8ff25/packages/hierarchy-builder/src/hierarchy-builder/values/Values.ts#L28) as a `compare` function.

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
```diff
@@ -228,6 +228,101 @@ export namespace HierarchyNodeKey {
...
+        assert(isLabelGrouping(rhs));
+        const labelCompareResult = compareStrings(lhs.label, rhs.label);
+        if (labelCompareResult !== 0) {
+          return labelCompareResult;
+        }
+        // Have to do this check in case one of them is `undefined` and the other is ""
+        if (typeof lhs.groupId !== typeof rhs.groupId) {
+          return lhs.groupId !== undefined ? 1 : -1;
+        }
+        return compareStrings(lhs.groupId ?? "", rhs.groupId ?? "");
```
grigasp, line 292: could use `compareStringsOrUndefined` from `@itwin/core-bentley`

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
```diff
@@ -228,6 +228,101 @@ export namespace HierarchyNodeKey {
...
+   * @returns
+   *- `0` if they are equal
+   *- `negative value` if lhs key is less than rhs key
+   *- `positive value` if lhs key is more than rhs key
+   */
+  export function compare(lhs: HierarchyNodeKey, rhs: HierarchyNodeKey): number {
+    // This compare is quicker than String.localeCompare
+    function compareStrings(lhsStr: string, rhsStr: string): number {
+      return lhsStr < rhsStr ? -1 : lhsStr > rhsStr ? 1 : 0;
+    }
```
grigasp, line 242: could use `compareStrings` from `@itwin/core-bentley`

packages/hierarchy-builder/src/test/HierarchyProvider.test.ts
```diff
@@ -849,6 +849,7 @@ describe("HierarchyProvider", () => {
                 fullClassName: "x.y",
                 query: { ecsql: "QUERY" },
               },
+              { node: { key: "root", label: "root" } },
```
grigasp, line 852: why was this necessary?

packages/hierarchy-builder/src/test/HierarchyNode.test.ts
```diff
@@ -156,6 +160,334 @@ describe("HierarchyNodeKey", () => {
...
+      const propertyGroupingRangeNodeKey: PropertyValueRangeGroupingNodeKey = {
+        type: "property-grouping:range",
+        propertyClassName: "",
+        propertyName: "",
+        fromValue: 1,
+        toValue: 2,
+        groupedInstanceKeys: [],
+      };
+
+      it("returns correct results when one of the keys is an instance node key", () => {
```
grigasp, line 187: Lots of compares here, a bit hard to read :) What do you think about the following approach:
1. define a single array with keys of different types
2. ```ts
    keys.forEach((lhs, lhsIndex) => {
      keys.forEach((rhs, rhsIndex) => {
        const lhsToRhs = HierarchyNodeKey.compare(lhs, rhs);
        const rhsToLhs = HierarchyNodeKey.compare(rhs, lhs);
        if (lhsIndex === rhsIndex) {
          expect(lhsToRhs).to.eq(0);
          expect(rhsToLhs).to.eq(0);
        } else {
          expect(lhsToRhs).to.not.eq(0);
          expect(rhsToLhs).to.not.eq(0);
          expect(rhsToLhs).to.eq(-1 * lhsToRhs);
        }
      });
    });
    ```

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
JonasDov: Moved

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
JonasDov: Changed

packages/hierarchy-builder/src/hierarchy-builder/HierarchyNode.ts
JonasDov: Changed

packages/hierarchy-builder/src/test/HierarchyProvider.test.ts
JonasDov: It wasn't necessary for this test specifically, but it now does exactly what needs to be done to get coverage for the `else` part [https://github.com/iTwin/presentation/blob/cbdf76ad832214d2fb2ef59d713c257bc3600668/packages/hierarchy-builder/src/hierarchy-builder/HierarchyProvider.ts#L436C7-L436C47](url).

packages/hierarchy-builder/src/test/HierarchyNode.test.ts
JonasDov: Yeah I agree, though about doing this, but for some reason decided that the original implementation was simpler to understand. 
Changed to your suggestion.

packages/hierarchy-builder/src/test/HierarchyProvider.test.ts
grigasp: > It wasn't necessary for this test specifically

That's exactly what's wrong with it. The change makes the line covered, but doesn't check if the covered line actually did anything or affected the result in any way, which it should. 

Options in the order of preference:
1. Write a test that tests the situation specifically
2. Restructure the library code to not require a specific test for the line to be covered
3. `// istanbul ignore next`

Unless you have an idea for a test, I think here the second option is fine - doing a `return lhs.length - rhs.length` should do the trick.

grigasp: > That line of code is needed for the Dictionary to work correctly, and I don't think that we should be testing something that we use from other packages.

I agree with both statements, but that piece of code is ours, not of some other package. Which means it should be tested for correctness on our end. Having said that, sometimes having a test for a very trivial piece of code is just not worth it.

grigasp (approved):

PR #411, opened 2024-02-12
Remove typemoq

saskliutas:
Replaced `typemoq` usage with `sinon`.


grigasp (approved):
tremendous effort, thanks!

PR #416, opened 2024-02-15
Filter hierarchy level only by class

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/773

### DIFF
--- a/.changeset/proud-readers-rest.md
+++ b/.changeset/proud-readers-rest.md
@@ -1,0 +1,6 @@
+---
+"@itwin/presentation-components": minor
+---
+
+`PresentationInstanceFilterDialog` now invokes `onApply` callback with only selected classes.
+Added `createInstanceFilterDefinition` that creates `InstanceFilterDefinition` from `PresentationInstanceFilterInfo`. Created definition can be passes to `PresentationManager` in order to filter results.
--- a/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
+++ b/packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
@@ -20,6 +20,7 @@ import {
 import { IModelConnection } from "@itwin/core-frontend";
 import { ClassInfo } from "@itwin/presentation-common";
 import { getIModelMetadataProvider } from "./ECMetadataProvider";
+import { PresentationInstanceFilter } from "./PresentationInstanceFilter";
 
 /** @internal */
 export async function findBaseExpressionClassName(imodel: IModelConnection, propertyClassNames: string[]) {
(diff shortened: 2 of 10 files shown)


packages/components/src/presentation-components/instance-filter-builder/InstanceFilterConverter.ts
```diff
@@ -65,8 +86,9 @@ function createExpressionFromGroup(group: GenericInstanceFilterRuleGroup): strin
   return `(${convertedConditions.join(` ${getGroupOperatorString(group.operator)} `)})`;
 }
 
-function createClassExpression(usedClasses: ClassInfo[]) {
-  return usedClasses.reduce((queryExpression, classInfo) => `${queryExpression}${queryExpression ? " OR " : ""}this.IsOfClass(${classInfo.id})`, "");
+/** @internal */
+export function createFilterClassExpression(usedClasses: ClassInfo[]) {
+  return `(${usedClasses.reduce((queryExpression, classInfo) => `${queryExpression}${queryExpression ? " OR " : ""}this.IsOfClass(${classInfo.id})`, "")})`;
```
grigasp, line 91: This should give the same result in an easier-to-read way
```suggestion
  return `(${usedClasses.map((classInfo) => `this.IsOfClass(${classInfo.id})`).join(" OR ")})`;
```

packages/components/src/presentation-components/instance-filter-builder/PresentationFilterBuilder.tsx
```diff
@@ -192,3 +193,32 @@ export function PresentationInstanceFilterBuilder(props: PresentationInstanceFil
...
 }
+
+/**
+ * Creates [InstanceFilterDefinition]($presentation-common) from [[PresentationInstanceFilterInfo]]. Created definition
+ * can be passed to [PresentationManager]($presentation-frontend) through request options in order to filter results.
+ * @beta
+ */
+export async function createInstanceFilterDefinition(info: PresentationInstanceFilterInfo, imodel: IModelConnection): Promise<InstanceFilterDefinition> {
+  if (!info.filter) {
+    assert(info.usedClasses.length > 0);
```
grigasp, line 204: Do we need this assert? Maybe just return empty `expression` if there's no filter or classes?

.changeset/proud-readers-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+"@itwin/presentation-components": minor
+---
+
+`PresentationInstanceFilterDialog` now invokes `onApply` callback with only selected classes.
+Added `createInstanceFilterDefinition` that creates `InstanceFilterDefinition` from `PresentationInstanceFilterInfo`. Created definition can be passes to `PresentationManager` in order to filter results.
```
grigasp, line 6:
```suggestion
`PresentationInstanceFilterDialog` now allows applying filter when only classes are selected.

Added `createInstanceFilterDefinition` that creates `InstanceFilterDefinition` from `PresentationInstanceFilterInfo`. Created definition can be passed to `PresentationManager` to filter results when creating content or hierarchies.
```

grigasp (changes requested):

grigasp (approved):

PR #418, opened 2024-02-15
Update filtering dialog results exceed count message

saskliutas:
Closes https://github.com/iTwin/viewer-components-react/issues/774

### DIFF
--- a/packages/components/public/locales/en/PresentationComponents.json
+++ b/packages/components/public/locales/en/PresentationComponents.json
@@ -43,7 +43,7 @@
     "clear-hierarchy-level-filter": "Clear active filter",
     "filter-dialog": {
       "results-count": "Matching items",
-      "result-limit-exceeded": "Exceeded matching item limit of"
+      "result-limit-exceeded": "There are too many results to display. Please adjust your filters to limit the results."
     }
   },
   "navigation-property-editor": {
--- a/packages/components/public/locales/en/PresentationComponents.json
+++ b/packages/components/public/locales/en/PresentationComponents.json
@@ -43,7 +43,7 @@
     "clear-hierarchy-level-filter": "Clear active filter",
     "filter-dialog": {
       "results-count": "Matching items",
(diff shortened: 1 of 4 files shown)


packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -155,9 +155,7 @@ function MatchingInstancesCount({ filter, dataProvider, parentKey }: MatchingIns
       } catch (e) {
         if (e instanceof PresentationError && e.errorNumber === PresentationStatus.ResultSetTooLarge) {
           // ResultSetTooLarge error can't occur if sizeLimit is undefined.
-          return `${translate("tree.filter-dialog.result-limit-exceeded")} ${requestOptions.sizeLimit!}. ${translate("tree.please-provide")} ${translate(
-            "tree.additional-filtering",
-          )}.`;
+          return translate("tree.filter-dialog.result-limit-exceeded");
```
grigasp, line 158: Should clean up other, now unused, strings from the localization file. Also, I personally would want to know what the limit is, so maybe change the message to something like this?

> There are too many results to display. Please adjust your filters to limit the results count to X items.

grigasp (changes requested):

packages/components/public/locales/en/PresentationComponents.json
```diff
@@ -43,7 +39,7 @@
     "clear-hierarchy-level-filter": "Clear active filter",
     "filter-dialog": {
       "results-count": "Matching items",
-      "result-limit-exceeded": "Exceeded matching item limit of"
+      "result-limit-exceeded": "There are too many results to display. Please adjust your filters to limit the results count to {{itemCount}} items"
```
grigasp, line 42:
```suggestion
      "result-limit-exceeded": "There are too many results to display. Please adjust your filters to limit the results count to {{itemCount}} items."
```

grigasp (approved):

PR #433, opened 2024-02-23
Copy all assets from @itwin dependencies

saskliutas:


apps/test-app/frontend/vite.config.mts
```diff
@@ -6,27 +6,17 @@ import { defineConfig } from "vite";
...
 export default defineConfig({
   plugins: [
     react(),
     viteStaticCopy({
       targets: [
-        ...localeDirs.map((dir) => ({
-          src: dir,
+        {
+          // copy assets from `@itwin` dependencies
+          src: "./node_modules/@itwin/**/**/public/*",
```
grigasp, line 17: shouldn't this be `./node_modules/@itwin/**/public/**/*`?

saskliutas: Changed `/**/**/` -> `/**/`. I guess I duplicated it accidentally while testing.

There should not be `public/**/*` at the end because it would copy public assets from subfolders into destination root e.g files from `public/locale/en/` would be copied to the root. `public/*`  allows to keep folder hierarchy.

grigasp (approved):

PR #437, opened 2024-02-26
Struct property sorting

jasdom:
Changed struct properties to be sorted.

### DIFF
--- a/.changeset/gorgeous-plants-nail.md
+++ b/.changeset/gorgeous-plants-nail.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-components": patch
 ---
 
-Added struct member sorting.
+Sort struct property members by label when creating content for property grid, table, and other content components.
--- a/.changeset/gorgeous-plants-nail.md
+++ b/.changeset/gorgeous-plants-nail.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Added struct member sorting.
--- a/packages/components/src/presentation-components/common/ContentBuilder.ts
+++ b/packages/components/src/presentation-components/common/ContentBuilder.ts
@@ -37,6 +37,7 @@ import {
 } from "@itwin/presentation-common";
 import { NumericEditorName } from "../properties/editors/NumericPropertyEditor";
 import { QuantityEditorName } from "../properties/editors/QuantityPropertyEditor";
+import { inPlaceSort } from "fast-sort";
 
 /** @internal */
 export interface FieldInfo {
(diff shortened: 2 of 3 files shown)


packages/components/src/presentation-components/common/ContentBuilder.ts
```diff
@@ -139,9 +139,10 @@ class StructMembersAppender implements INestedPropertiesAppender {
     this._members[record.fieldHierarchy.field.name] = record.record;
   }
   public finish(): void {
+    const sortedMembers = Object.entries(this._members).sort();
     const value: StructValue = {
       valueFormat: UiPropertyValueFormat.Struct,
-      members: this._members,
+      members: Object.fromEntries(sortedMembers),
```
saskliutas, line 145: Can you add test to check properties order?

saskliutas (changes requested):

packages/components/src/presentation-components/common/ContentBuilder.ts
jasdom: Added.

packages/components/src/presentation-components/common/ContentBuilder.ts
```diff
@@ -139,9 +139,10 @@ class StructMembersAppender implements INestedPropertiesAppender {
     this._members[record.fieldHierarchy.field.name] = record.record;
   }
   public finish(): void {
+    const sortedMembers = Object.entries(this._members).sort();
```
grigasp, line 142: In what order is this going to sort? `Object.entries` returns an array of tuples and `sort` orders items in ascending ASCII order 🤔 I think in the best case scenario it's going to be by the first tuple item - the property name, which is not necessarily correct. We want to sort by property label instead.

grigasp (changes requested):

grigasp:
Please also add a changeset.

packages/components/src/presentation-components/common/ContentBuilder.ts
jasdom: It was sorting based on the property name, changed to sort by label instead.

.changeset/gorgeous-plants-nail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Added struct member sorting.
```
grigasp, line 5: Should try to be more specific.

```suggestion
Sort struct property members by label when creating content for property grid, table, and other content components.
```

jasdom: Updated the description.

saskliutas (approved):

grigasp (approved):

PR #438, opened 2024-02-27
`PresentationInstanceFilterDialog`: Make API more convienent

saskliutas:
Updated `PresentationInstanceFilterDialogProps` to make `PresentationInstanceFilterDialog` more convenient to use in case input keys are not now before loading `Descriptore`.

### DIFF
--- a/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
+++ b/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
@@ -24,7 +24,7 @@ import { filterRuleValidator, isFilterNonEmpty } from "./Utils";
  * Data structure that describes source to gather properties from.
  * @beta
  */
-export interface PropertiesSource {
+export interface PresentationInstanceFilterPropertiesSource {
   /**
    * [Descriptor]($presentation-common) that will be used to get properties.
    */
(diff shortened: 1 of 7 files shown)


packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
```diff
@@ -20,14 +20,33 @@ import { PresentationInstanceFilterInfo } from "./PresentationFilterBuilder";
 import { PresentationInstanceFilter } from "./PresentationInstanceFilter";
 import { filterRuleValidator, isFilterNonEmpty } from "./Utils";
 
+/**
+ * Data structure that describes source to gather properties from.
+ * @beta
+ */
+export interface PropertiesSource {
```
grigasp, line 27: Maybe `PresentationInstanceFilterPropertiesSource`? `PropertiesSource` sounds very generic and in a package, containing many different components, makes it unclear what component it's related to.

grigasp (approved):

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilterDialog.tsx
saskliutas: Renamed

PR #440, opened 2024-02-27
[core-interop]: Returns input keys with hierarchy level descriptor

saskliutas:
Input keys are needed when loading distinct values using `Hierarchy level descriptor ruleset`

### DIFF
--- a/packages/core-interop/api/presentation-core-interop.exports.csv
+++ b/packages/core-interop/api/presentation-core-interop.exports.csv
@@ -1,7 +1,9 @@
 sep=;
 Release Tag;API Item
 beta;createECSqlQueryExecutor(imodel: ICoreECSqlReaderFactory): IECSqlQueryExecutor
+beta;createHierarchyLevelDescriptor
 beta;CreateHierarchyLevelDescriptorProps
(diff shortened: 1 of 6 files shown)


packages/core-interop/api/presentation-core-interop.exports.csv
```diff
@@ -1,7 +1,6 @@
 sep=;
 Release Tag;API Item
 beta;createECSqlQueryExecutor(imodel: ICoreECSqlReaderFactory): IECSqlQueryExecutor
-beta;createHierarchyLevelDescriptor
```
grigasp, line 4: any idea why this disappeared?

grigasp (approved):

packages/core-interop/api/presentation-core-interop.exports.csv
saskliutas: Not sure. I thought it might have something to do with mixed dependencies in `node_modules` after switching branches but looks like CI pipeline is ok with this change.

saskliutas: Adding return type instead of using plain object fixed this.

Issue #439, opened 2024-02-27
`UniqueValueSelector` dropdown menu should open according available space


saskliutas:
`UniqueValueSelector` dropdown menu always opens at the bottom. This can cause problems selector is rendered at the bottom of viewport and there are not enough space for dropdown menu. In this case dropdown menu should be open above selector.

Additionally, need to investigate a better way to portal dropdown menu. `.iui-root` class is not available in iTwinUI v3 https://github.com/iTwin/presentation/blob/8791d60cc2f16f50bcec996654a30e3392514c6d/packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx#L148

PR #444, opened 2024-02-29
Change unique value selector dropdown menu to be space aware

jasdom:
Closes [439](https://github.com/iTwin/presentation/issues/439)

- Modified `UniqueValueSelector` dropdown menu to open upwards when there is not enough space below.
- Changed the dropdown menu portal target to the filtering dialog instead of the body.

### DIFF
--- a/packages/components/src/presentation-components/common/PortalTargetContext.tsx
+++ b/packages/components/src/presentation-components/common/PortalTargetContext.tsx
@@ -1,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Core
+ */
+
+import { createContext, PropsWithChildren, useContext } from "react";
+
+/**
+ * Context that stores a portal target.
+ */
+export interface PortalTargetContext {
+  portalTarget: HTMLElement | null;
+}
+
+const portalTargetContext = createContext<PortalTargetContext>({} as PortalTargetContext);
+
+/**
+ * Props for [[PortalTargetContextProvider]]
+ */
+export interface PortalTargetContextProviderProps {
+  portalTarget: HTMLElement | null;
+}
+
+/**
+ * Provides a portal target for components.
+ */
+export function PortalTargetContextProvider({ portalTarget, children }: PropsWithChildren<PortalTargetContextProviderProps>) {
+  return <portalTargetContext.Provider value={{ portalTarget }}>{children}</portalTargetContext.Provider>;
+}
+
+/**
+ * Returns context provided by [[PortalTargetContextProvider]].
+ */
+export function usePortalTargetContext() {
+  return useContext(portalTargetContext);
+}
--- a/packages/components/src/presentation-components/common/PortalTargetContext.tsx
+++ b/packages/components/src/presentation-components/common/PortalTargetContext.tsx
@@ -9,7 +9,8 @@
 import { createContext, PropsWithChildren, useContext } from "react";
 
 /**
- * Context that stores a portal target.
+ * Context that stores a portal target. It will be used to portal popovers opened by presentation components.
+ * @beta
  */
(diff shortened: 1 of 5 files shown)


packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
```diff
@@ -114,17 +115,44 @@ function ClearIndicator<TOption, IsMulti extends boolean = boolean>({ children:
...
 export function AsyncSelect<OptionType, Group extends GroupBase<OptionType>, Additional>(props: AsyncPaginateProps<OptionType, Group, Additional, true>) {
+  const [dropdownUp, setDropdownUp] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+  const menuObserver = useRef<IntersectionObserver>();
+  const selectRef = useRef<SelectInstance<OptionType, true, Group>>(null);
   const divRef = useRef<HTMLDivElement>(null);
   const { ref: resizeRef, width } = useResizeObserver();
+
+  const onMenuOpen = () => {
+    const observeOnscreen: IntersectionObserverCallback = (entries: IntersectionObserverEntry[]) => {
```
saskliutas, line 126: I think we can take a simpler approach here. I think menu max-height is set to 300px so it should be enough to check if there are enough space below select for max heigh menu.

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
```diff
@@ -144,8 +172,12 @@ export function AsyncSelect<OptionType, Group extends GroupBase<OptionType>, Add
           SelectContainer,
           NoOptionsMessage,
         }}
+        selectRef={selectRef}
+        onMenuOpen={onMenuOpen}
+        onMenuClose={onMenuClose}
+        menuPortalTarget={divRef.current?.ownerDocument.body.querySelector(".presentation-instance-filter-dialog")}
```
saskliutas, line 178: I think we need to allow specifying different portal target instead of hard coding selector, because it can be used in property-grid too. Also in other spaces where editors from `@itwin/components-react` are used. Maybe we could add React context for specifying portal target.

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
jasdom: Switched to only checking for max-height.

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
jasdom: Added a context for providing portal targets.

packages/components/src/presentation-components/common/PortalTargetContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+/** @packageDocumentation
+ * @module Core
+ */
+
+import { createContext, PropsWithChildren, useContext } from "react";
+
+/**
+ * Context that stores a portal target.
```
saskliutas, line 12:
```suggestion
 * Context that stores a portal target. It will be used to portal popovers opened by presentation components.
```

packages/components/src/presentation-components/common/PortalTargetContext.tsx
```diff
@@ -0,0 +1,39 @@
...
+ */
+export interface PortalTargetContext {
```
saskliutas, line 14: Need to mark exported API with release tag. I think in this case we can mark them as `@beta`

saskliutas (changes requested):

packages/components/src/presentation-components/common/PortalTargetContext.tsx
jasdom: Changed.

packages/components/src/presentation-components/common/PortalTargetContext.tsx
jasdom: Marked with `@beta`.

saskliutas (approved):

Issue #442, opened 2024-02-28
`UniqueValuesSelector` loads only first page of values


saskliutas:
`UniqueValuesSelector` should load more values when bottom is reached if there are more than one page.

PR #448, opened 2024-03-01
Load more than one page in UniqueValueSelector

jasdom:
Closes [442](https://github.com/iTwin/presentation/issues/442)

Fixed an issue with `UniqueValueSelector` only loading the first page of values.

### DIFF
--- a/.changeset/large-hotels-fly.md
+++ b/.changeset/large-hotels-fly.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-components": patch
 ---
 
-Fixed UniqueValueSelector from loading only the first page of values.
+Fixed UniqueValueSelector loading only the first page of values.
--- a/.changeset/large-hotels-fly.md
+++ b/.changeset/large-hotels-fly.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-components": patch
 ---
 
-Fixed UniqueValueSelector loading only the first page of values.
+Fixed `UniqueValueSelector` loading only the first page of values.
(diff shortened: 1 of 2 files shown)


packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
```diff
@@ -122,6 +122,9 @@ export function AsyncSelect<OptionType, Group extends GroupBase<OptionType>, Add
   const { portalTarget } = usePortalTargetContext();
   const divRef = useRef<HTMLDivElement>(null);
 
+  // Wrap custom menu as a workaround for some internal bugs of react-select
+  const MenuList = wrapMenuList(CustomMenuList);
```
saskliutas, line 126: Can you move with outside AsyncSelect to avoid wrapping on each render?

.changeset/large-hotels-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Fixed UniqueValueSelector from loading only the first page of values.
```
saskliutas, line 5:
```suggestion
Fixed UniqueValueSelector loading only the first page of values.
```

saskliutas (approved):

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
jasdom: Moved.

.changeset/large-hotels-fly.md
jasdom: Changed.

PR #466, opened 2024-03-08
Unified selection storage

jasdom:
Selection storage for unified-selection package:
- Stores ECInstances and custom selectable objects.
- API for adding, deleting, replacing and clearing the storage.
- Event for listening for selection changes.


packages/unified-selection/src/test/SelectionStorage.test.ts
```diff
@@ -0,0 +1,343 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import * as sinon from "sinon";
+import * as moq from "typemoq";
```
saskliutas, line 8: `typemoq` is not actively maintained. Can you remove it? It looks like it is used for mocking `imodeConnection` but it seems that something like `const imodel = {} as IModelConnection` should do the trick.

packages/unified-selection/src/test/_helpers/SelectableObjectCreator.ts
```diff
@@ -0,0 +1,35 @@
+import faker from "@faker-js/faker";
```
saskliutas, line 1: I don't think it's worth having random data for tests. It also adds additional dependency. Can you remove `faker` dependency?

saskliutas (changes requested):

packages/unified-selection/src/unified-selection/SelectionStorage.ts
```diff
@@ -0,0 +1,268 @@
...
+  public readonly selectionChange: StorageSelectionChangeEvent;
+
+  /**
+   * Creates an instance of `SelectionStorage`.
+   */
+  constructor() {
+    this.selectionChange = new StorageSelectionChangeEvent();
+    IModelConnection.onClose.addListener((imodel: IModelConnection) => {
+      this.onConnectionClose(imodel);
+    });
```
grigasp, line 85: Without a dependency on `IModelConnection` we won't be able to add this listener. Need to come up with a plan for handling cleanup.

packages/unified-selection/src/unified-selection/StorageSelectionChangeEvent.ts
```diff
@@ -0,0 +1,61 @@
...
+  /** The selection change type. */
+  changeType: StorageSelectionChangeType;
+  /** Set of selectable objects affected by this selection change event. */
+  selectableObjects: Readonly<SelectableObjectSet>;
+  /** iModel connection with which the selection is associated with. */
+  imodel: IModelConnection;
+  /** The timestamp of when the selection change happened */
+  timestamp: Date;
+  /** Id of the ruleset associated with the selection change. */
+  rulesetId?: string;
```
grigasp, line 60: not needed

```suggestion
  /** Id of the ruleset associated with the selection change. */
  rulesetId?: string;
```

grigasp (changes requested):

packages/unified-selection/src/unified-selection/SelectionStorage.ts
```diff
@@ -0,0 +1,237 @@
...
+export enum StorageSelectionChangeType {
+  /** Added to selection. */
+  Add,
+  /** Removed from selection. */
+  Remove,
+  /** Selection was replaced. */
+  Replace,
+  /** Selection was cleared. */
+  Clear,
+}
```
grigasp, line 31: I think we should use string union here - more flexible.

packages/unified-selection/src/unified-selection/SelectionStorage.ts
```diff
@@ -0,0 +1,237 @@
...
+  timestamp: Date;
+}
+
+/**
+ * Selection storage interface which provides main selection and sub-selection.
+ * @beta
+ */
+export interface SelectionStorageInterface {
+  /** A callback that is called when selection changes */
+  onSelectionChange: (event: StorageSelectionChangeEventArgs, storage: SelectionStorageInterface) => void;
```
grigasp, line 58: This won't support multiple event listeners. Instead, please define an `Event` interface and use that. See here: https://github.com/iTwin/presentation/blob/0ab2c5c4a51620c6bd2af3e39ef55cbaca9dcb79/packages/core-interop/src/core-interop/Transactions.ts#L6C1-L15C2.

packages/unified-selection/src/unified-selection/SelectionStorage.ts
```diff
@@ -0,0 +1,237 @@
...
+  iModelKey: string;
+  /** The timestamp of when the selection change happened */
+  timestamp: Date;
+}
+
+/**
+ * Selection storage interface which provides main selection and sub-selection.
+ * @beta
+ */
+export interface SelectionStorageInterface {
```
grigasp, line 56: Not a good practice to have `Interface` in an `interface`. Suggested renames:
- `SelectionStorageInterface` -> `SelectionStorage`
- `SelectionStorage` -> `SelectionStorageImpl`

packages/unified-selection/src/unified-selection/SelectionStorage.ts
grigasp: Now that I think of it, maybe it's enough to add some docs for `createStorage` explaining that the consumer needs to call `clearStorage` when iModel is closed.

packages/unified-selection/src/unified-selection/SelectionChangeEvent.ts
```diff
@@ -0,0 +1,109 @@
...
+ */
+export type StorageSelectionChangeType =
+  /** Added to selection. */
+  | "Add"
+  /** Removed from selection. */
+  | "Remove"
+  /** Selection was replaced. */
+  | "Replace"
+  /** Selection was cleared. */
+  | "Clear";
```
grigasp, line 25:
```suggestion
export type StorageSelectionChangeType =
  /** Added to selection. */
  | "add"
  /** Removed from selection. */
  | "remove"
  /** Selection was replaced. */
  | "replace"
  /** Selection was cleared. */
  | "clear";
```

packages/unified-selection/src/test/SelectionChangeEvent.test.ts
```diff
@@ -0,0 +1,42 @@
+import { expect } from "chai";
```
grigasp, line 1: missing file header

saskliutas (approved):

Issue #465, opened 2024-03-07
`@itwin/presentation-hierarchies-react`


saskliutas:
Need to set up a package for working with stateless hierarchies builder in react.
```[tasklist]
### Tasks
- [ ] https://github.com/iTwin/presentation/pull/474
- [ ] https://github.com/iTwin/presentation/pull/478
- [ ] https://github.com/iTwin/presentation/pull/511
```

PR #474, opened 2024-03-13
Initial `@itwin/presentation-hierarchies-react` setup

saskliutas:
Moved `useTree` and related code that provides React state management when working with stateless hierarchies to the new package. This is the initial commit that setup package. https://github.com/iTwin/presentation/issues/465 was updated to document next steps.


packages/hierarchies-react/src/presentation-hierarchies-react/UnifiedSelectionContext.tsx
```diff
@@ -0,0 +1,62 @@
+import { createContext, PropsWithChildren, useContext, useEffect, useState } from "react";
```
grigasp, line 1: missing file header

packages/hierarchies-react/src/presentation-hierarchies-react/UnifiedSelectionContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface Props {
+  store: UnifiedSelectionStore;
+}
+
+/** @beta */
+export function UnifiedSelectionContextProvider({ store, children }: PropsWithChildren<Props>) {
+  const [value, setValue] = useState<UnifiedSelectionContext>({
+    has: (keys) => store.container.has(keys),
+    add: (keys) => store.container.add(keys),
```
grigasp, line 42: I know integration with the new `@itwin/unified-selection` is still a TODO, but just noting - we'll want to support "replace", too.

grigasp (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/UnifiedSelectionContext.tsx
saskliutas: Yes. I think I will need to reiterate unified selection handling in tree when integrating the new package.

packages/hierarchies-react/src/presentation-hierarchies-react/UnifiedSelectionContext.tsx
saskliutas: Added. Also updated script to handle `.tsx` files.

PR #475, opened 2024-03-14
Add an additional test iModel to the performance tests

### LINKED ISSUE #424: Performance tests for hierarchy builder
Need more performance tests for the hierarchy builder. 

What to measure:
- Overall performance of a test case.
- Time the library was blocking the main thread (see https://github.com/tj/node-blocked/tree/master).

Test cases:
- Large flat hierarchy level
- Each grouping type individually
- All grouping types 

For the dataset, ideally we should create one. Consider using [pipeline caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action) to avoid creating it on each run.

We want the tests to be part of our performance benchmark. ~The tests currently using `artillery` should probably be removed or refactored to not use `artillery`.~ Currently the benchmark is using `artillery` load tests, which we want to keep, but stop using for benchmarking.

Yato333:
Resolves #424 
- Move test iModel creation code from the `presentation-full-stack-tests` project to a new project that will not have unnecessary dependencies like presentation-frontend, AppUI or React.
- Add a large flat iModel to performance test dataset.
- Implement caching of datasets in the Github workflow.


apps/performance-tests/src/Datasets.ts
```diff
@@ -4,8 +4,54 @@
...
+const BAYTOWN_DOWNLOAD_URL = "https://github.com/imodeljs/desktop-starter/raw/master/assets/Baytown.bim";
+
+export class Datasets {
+  private static readonly _iModels: { baytown?: string; largeFlat?: string } = {};
+
+  public static get bayTown(): string {
+    return this.verifyInitialized(this._iModels.baytown);
+  }
+
+  public static get largeFlat(): string {
```
grigasp, line 20: I'd rename "large flat" (wherever it's currently used) to "50k elements" to make it clear what to expect.

apps/performance-tests/src/Datasets.ts
```diff
@@ -4,8 +4,54 @@
...
-async function downloadDataset(name: string, downloadUrl: string, localPath: string): Promise<void> {
+const LARGE_FLAT_IMODEL_SIZE = 50_000;
+const BAYTOWN_DOWNLOAD_URL = "https://github.com/imodeljs/desktop-starter/raw/master/assets/Baytown.bim";
+
+export class Datasets {
+  private static readonly _iModels: { baytown?: string; largeFlat?: string } = {};
+
+  public static get bayTown(): string {
+    return this.verifyInitialized(this._iModels.baytown);
+  }
```
grigasp, line 18: I'm not sure we want to add this getter for each dataset. Maybe we should have something like this?
```ts
type IModelNames = "baytown" | "50k elements";
type IModelPathsMap = {[name in IModelNames]: string};
class Datasets {
  private static _map: IModelPathsMap;
  public static getPath(name: IModelNames) {
    return this.verifyInitialized(this._map[name]);
  }
}
```

Yato333: Change it to a typed map

apps/performance-tests/src/Datasets.ts
Yato333: Renamed the module to "50k elements" and refactored for easier addition of flat iModels in the future

.github/workflows/benchmark-pr.yml
```diff
@@ -40,9 +40,25 @@ jobs:
...
         run: pnpm lage build --to presentation-performance-tests
 
+      - name: Restore dataset from cache
+        uses: actions/cache/restore@v3
+        id: cache-restore
+        env:
+          cache-name: cache-dataset
+        with:
+          path: "./apps/performance-tests/datasets/*.bim"
+          key: performance-test-dataset-${{ hashFiles('./apps/performance-tests/src/**.ts') }}
```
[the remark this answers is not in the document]
Yato333: Changed it to only use `Datasets.ts`

apps/performance-tests/src/stateless.test.ts
```diff
@@ -3,28 +3,36 @@
...
   });
 
   run("loads full hierarchy", async () => {
-    const provider = new StatelessHierarchyProvider(iModel);
+    const provider = new StatelessHierarchyProvider({ iModel: baytown });
+    await provider.loadFullHierarchy();
+  });
+
+  run("loads large flat hierarchy", async () => {
+    const provider = new StatelessHierarchyProvider({ iModel: largeFlat, rowLimit: "unbounded" });
```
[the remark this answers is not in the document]
Yato333: Wrote a custom hierarchy definition

apps/performance-tests/src/stateless.test.ts
```diff
@@ -3,28 +3,36 @@
...
     await provider.loadInitialHierarchy();
   });
 
   run("loads full hierarchy", async () => {
-    const provider = new StatelessHierarchyProvider(iModel);
+    const provider = new StatelessHierarchyProvider({ iModel: baytown });
+    await provider.loadFullHierarchy();
+  });
+
+  run("loads large flat hierarchy", async () => {
```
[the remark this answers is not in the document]
Yato333: Improved test output

packages/full-stack-tests/src/IModelUtils.ts
```diff
@@ -10,28 +10,14 @@ import { tmpdir } from "os";
...
-  SubCategoryProps,
-  SubjectProps,
-} from "@itwin/core-common";
+import { LocalFileName } from "@itwin/core-common";
 import { IModelConnection } from "@itwin/core-frontend";
 import { ECSqlBinding, parseFullClassName, Point2d, Point3d, PrimitiveValue } from "@itwin/presentation-hierarchy-builder";
 import { buildTestIModel, TestIModelBuilder } from "@itwin/presentation-testing";
 import { createFileNameFromString } from "@itwin/presentation-testing/lib/cjs/presentation-testing/InternalUtils";
 
+export * from "presentation-test-utilities/lib/IModelUtils";
```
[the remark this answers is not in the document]
Yato333: Fixed

apps/performance-tests/src/util/TestUtilities.ts
```diff
@@ -3,10 +3,40 @@
...
+      await callbackOrProps();
+      return;
+    }
+
+    let value: T;
+    try {
+      value = await callbackOrProps.setup();
+    } finally {
+      this.test!.ctx!.reporter.onTestStart();
+    }
```
grigasp, line 33: I don't see a reason to place this into `try/finally` - if the setup fails we're going to throw, so why even call `onTestStart`?

packages/test-utilities/src/presentation-test-utilities.ts
```diff
@@ -0,0 +1 @@
+export * from "./test-utilities/IModelUtils";
```
grigasp, line 1: missing file header

apps/performance-tests/src/util/TestUtilities.ts
Yato333: If `setup()` throws an error we will only be able to see the test failure at the end:
```
 models tree
    ✔ initial (Baytown) (58.55 ms)
    ✔ full (Baytown) (1503.11 ms)

Status | Test                          | Duration   | Blocks                                           
-------------------------------------------------------------------------------------------------------
PASS   | models tree initial (Baytown) | 58.55 ms   | count: 0                                         
PASS   | models tree full (Baytown)    | 1503.11 ms | count: 3, max: 130.00, p95: 130.00, median: 71.00

flat 50k elements list:
Error
    at Object.setup (src/stateless.test.ts:36:11)
    at Context.<anonymous> (src/util/TestUtilities.ts:28:41)
    at process.processImmediate (node:internal/timers:476:21)
```

grigasp: makes sense, thanks

grigasp (approved):

PR #478, opened 2024-03-15
Use `@itwin/unified-selection` to handle selection in `@itwin/presentation-hierarchies-react`

saskliutas:
Initial implementation of selection handling using `@itwin/unified-selection`.


packages/hierarchies-react/src/presentation-hierarchies-react/internal/UseUnifiedSelection.ts
```diff
@@ -18,97 +17,105 @@ export interface TreeSelectionOptions {
...
         return false;
       }
 
-      return has(getNodeKeys(node.nodeData));
+      const selectables = storage.getSelection(key, 0);
+
+      const hierarchyNode = node.nodeData;
+      if (HierarchyNode.isClassGroupingNode(hierarchyNode)) {
+        return Selectables.has(selectables, { identifier: node.id });
+      }
```
grigasp, line 80: I think here we should handle all grouping and custom nodes

packages/hierarchies-react/src/presentation-hierarchies-react/internal/UseUnifiedSelection.ts
```diff
@@ -18,97 +17,105 @@ export interface TreeSelectionOptions {
...
       const node = getNode(nodeId);
       if (!node || !isTreeModelHierarchyNode(node)) {
         return;
       }
 
-      const action = isSelected ? add : remove;
-      action(getNodeKeys(node.nodeData));
+      const action = isSelected ? storage.addToSelection.bind(storage) : storage.removeFromSelection.bind(storage);
+      action(source, key, createSelectables(node.id, node.nodeData), 0);
     },
```
grigasp, line 97: need to ensure our current uses cases are supported, e.g. the extended selection mode: 
- click to replace, 
- shift+click to invert/select multiple nodes, 
- ctrl+click to (de)select a node.

saskliutas: Filled separate issue to avoid expanding scope of this PR. https://github.com/iTwin/presentation/issues/481

PR #479, opened 2024-03-15
Add performance tests for all grouping types

### LINKED ISSUE #424: Performance tests for hierarchy builder

Yato333:
Closes #424

- Add a custom schema to the 50k element iModel.
- Make elements of the 50k iModel groupable by user label, class and properties.
- Add support for assertions.
- Add `only` and `skip` parameters for the `run` function.
- Add logging of iTwin.js errors.

### DIFF
--- a/apps/performance-tests/src/Datasets.ts
+++ b/apps/performance-tests/src/Datasets.ts
@@ -6,7 +6,6 @@
 import fs from "fs";
 import path from "path";
 import { insertPhysicalElement, insertPhysicalModelWithPartition, insertSpatialCategory } from "presentation-test-utilities";
-import { Logger, LogLevel } from "@itwin/core-bentley";
 import { createIModel } from "./util/IModelUtilities";
 
 export const IMODEL_NAMES = ["baytown", "50k elements"] as const;
@@ -23,9 +22,6 @@ export class Datasets {
   }
 
   public static async initialize(datasetsDirPath: string) {
-    Logger.initializeToConsole();
-    Logger.setLevelDefault(LogLevel.Error);
-
     await fs.promises.mkdir(datasetsDirPath, { recursive: true });
     const promises = IMODEL_NAMES.map(async (key) => {
       if (key === "baytown") {
@@ -40,7 +36,7 @@ export class Datasets {
         key,
         datasetsDirPath,
         async (name: string, localPath: string) => createFlatIModel(name, localPath, count),
-        !process.env.REUSE_DATASET,
+        !!process.env.RECREATE,
       );
       this._iModels[key] = iModelPath;
     });
@@ -93,10 +89,13 @@ async function createFlatIModel(name: string, localPath: string, numElements: nu
   // prettier-ignore
   const schema = `
     <ECSchemaReference name="BisCore" version="01.00.16" alias="bis" />
+    <ECEntityClass typeName="Base_${defaultClassName}">
+      <BaseClass>bis:PhysicalElement</BaseClass>
+      <ECProperty propertyName="${customPropName}" typeName="string" />
+    </ECEntityClass>
     ${[...Array(numGroups).keys()].map((i) => `
       <ECEntityClass typeName="${defaultClassName}_${i}">
-        <BaseClass>bis:PhysicalElement</BaseClass>
-        <ECProperty propertyName="${customPropName}" typeName="string" />
+        <BaseClass>Base_${defaultClassName}</BaseClass>
       </ECEntityClass>
     `).join("")}
   `;
--- a/apps/performance-tests/src/Datasets.ts
+++ b/apps/performance-tests/src/Datasets.ts
@@ -119,7 +119,7 @@ export class Datasets {
             userLabel: `${defaultUserLabel}_${groupIdx}`,
             modelId,
             categoryId,
-            propX: `${defaultPropertyValue}_${groupIdx}`,
+            [customPropName]: `${defaultPropertyValue}_${groupIdx}`,
           });
         }
       }
--- a/apps/performance-tests/src/Datasets.ts
+++ b/apps/performance-tests/src/Datasets.ts
@@ -6,6 +6,7 @@
 import fs from "fs";
 import path from "path";
 import { insertPhysicalElement, insertPhysicalModelWithPartition, insertSpatialCategory } from "presentation-test-utilities";
+import { Logger, LogLevel } from "@itwin/core-bentley";
 import { createIModel } from "./util/IModelUtilities";
 
 export const IMODEL_NAMES = ["baytown", "50k elements"] as const;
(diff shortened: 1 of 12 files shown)


apps/performance-tests/src/stateless.test.ts
```diff
@@ -31,42 +32,127 @@ describe("models tree", () => {
...
+    expectedNodeCount: baseClassQueryLimit,
+    nodeSelectProps: {
+      grouping: {
+        byBaseClasses: { fullClassNames },
+      },
+    },
+  });
+
+  runQueryTest({
+    testName: "by everything",
```
grigasp, line 86:
```suggestion
    testName: "by multiple attributes",
```

apps/performance-tests/src/stateless.test.ts
```diff
@@ -31,42 +32,127 @@ describe("models tree", () => {
...
+  });
 });
+
+function runQueryTest(testProps: {
+  testName: string;
+  iModelName: IModelName;
+  fullClassName?: string;
+  nodeSelectProps?: Partial<NodeSelectClauseProps>;
+  expectedNodeCount?: number;
+  limit?: number;
```
grigasp, line 109: looks like `limit` prop is not used

apps/performance-tests/src/util/IModelUtilities.ts
```diff
@@ -3,11 +3,13 @@
...
 
-import { TestIModelBuilder } from "presentation-test-utilities";
+import fs from "fs";
+import { getFullSchemaXml, TestIModelBuilder } from "presentation-test-utilities";
 import { IModelDb, SnapshotDb } from "@itwin/core-backend";
 import { BisCodeSpec, Code, ElementAspectProps, ElementProps, ModelProps } from "@itwin/core-common";
 
-export async function createIModel(name: string, localPath: string, cb: (builder: TestIModelBuilder) => void | Promise<void>) {
+export async function createIModel(name: string, localPath: string, cb: (builder: BackendTestIModelBuilder) => void | Promise<void>) {
+  await fs.promises.rm(localPath, { force: true });
```
grigasp, line 12: Is there a point to use promise api vs sync?

```suggestion
  fs.rmSync(localPath, { force: true });
```

apps/performance-tests/src/util/IModelUtilities.ts
```diff
@@ -37,4 +39,9 @@ class BackendTestIModelBuilder implements TestIModelBuilder {
     const spec = this._iModel.codeSpecs.getByName(codeSpecName).id;
     return new Code({ scope: scopeModelId, spec, value: codeValue });
   }
+
+  public async importSchema(schemaContentXml: string, schemaName: string): Promise<void> {
```
grigasp, line 43: maybe schema name should go before its content?

apps/performance-tests/src/util/TestUtilities.ts
```diff
@@ -12,19 +12,31 @@ interface AdvancedRunProps<TContext> {
 
   /** Callback that cleans up the context produced by the "before" callback. */
   cleanup?: (x: TContext) => void | Promise<void>;
+
+  /** Whether or not to run exclusively this test. */
+  only?: boolean;
```
grigasp, line 17: with these changes, how do I run only `initial (Baytown)` or `grouping by label` tests?

apps/performance-tests/src/StatelessHierarchyProvider.ts
```diff
@@ -37,22 +37,31 @@ export class StatelessHierarchyProvider {
...
 
-  private async loadNodes(nodeHasChildren: (node: HierarchyNode) => boolean) {
-    await new Promise<void>((resolve, reject) => {
+  private async loadNodes(nodeHasChildren: (node: HierarchyNode) => boolean): Promise<number> {
+    let nodeRequestLimit: number | undefined;
+    if (this._props.nodeRequestLimit === undefined) {
+      nodeRequestLimit = DEFAULT_NODE_REQUEST_LIMIT;
+    } else if (this._props.nodeRequestLimit !== "unbounded") {
+      nodeRequestLimit = this._props.nodeRequestLimit;
+    }
```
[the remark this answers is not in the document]
Yato333: If `this._props.nodeRequestLimit` equals `"unbounded"`, we want to pass `undefined` to RXJS.

apps/performance-tests/src/Datasets.ts
```diff
@@ -44,44 +59,72 @@ export class Datasets {
...
+      return path.resolve(localPath);
+    }
+
+    try {
+      await fs.promises.access(localPath, fs.constants.F_OK);
+    } catch {
+      await provider(name, localPath);
+    }
+    return path.resolve(localPath);
   }
```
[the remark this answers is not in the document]
Yato333: I figured, I could use `fs.existsSync` instead, which even more simplifies the code

apps/performance-tests/src/stateless.test.ts
```diff
@@ -31,42 +32,127 @@ describe("models tree", () => {
...
+        byProperties: {
+          propertiesClassName: baseFullClassName,
+          propertyGroups: [{ propertyName: customPropName, propertyClassAlias: "this" }],
+        },
+      },
+    },
+  });
 });
+
+function runQueryTest(testProps: {
```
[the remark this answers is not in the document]
Yato333: Renamed

apps/performance-tests/src/stateless.test.ts
```diff
@@ -31,42 +32,127 @@ describe("models tree", () => {
...
+  const baseClassQueryLimit = 10;
+  const fullClassNames = [...Array(baseClassQueryLimit).keys()].map((i) => `${schemaName}.${defaultClassName}_${i}`);
+  runQueryTest({
+    testName: `by base class (${baseClassQueryLimit} classes)`,
+    iModelName: "50k elements",
+    fullClassName: baseFullClassName,
+    expectedNodeCount: baseClassQueryLimit,
+    nodeSelectProps: {
+      grouping: {
+        byBaseClasses: { fullClassNames },
```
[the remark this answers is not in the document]
Yato333: Added `bis.PhysicalElement` and `baseFullClassName` to `fullClassNames`

apps/performance-tests/src/stateless.test.ts
Yato333: Removed limit prop

apps/performance-tests/src/util/IModelUtilities.ts
Yato333: I wanted to have this method compatible with `TestIModelBuilder` in `presentation-testing` because `BackendTestIModelBuilder` base class `TestIModelBuilder` has to be compatible with that one of the `presentation-testing`. But since this method is already incompatible with in `presentation-testing`'s `TestIModelBuilder`, I can swap the arguments. Done.

apps/performance-tests/src/util/TestUtilities.ts
Yato333: Fixed

grigasp (approved):

PR #485, opened 2024-03-21
Base class grouping performance optimizations

### LINKED ISSUE #480: Investigate why base class grouping is so slow
Performance benchmark shows that base class grouping is twice as slow as other grouping types:

| Benchmark suite | Current: b6162e1199d4ef4385a1f5b2da52e323128bd1a5 | 
|-|-|
| `grouping by label` | `8416.82` ms | 
| `grouping by label (P95 of main thread blocks)` | `0` ms | 
| `grouping by class` | `8576.47` ms | 
| `grouping by class (P95 of main thread blocks)` | `0` ms | 
| `grouping by property` | `9114.43` ms | 
| `grouping by property (P95 of main thread blocks)` | `447` ms | 
| **`grouping by base class (10 classes)`** | **`23407.32` ms** | 
| **`grouping by base class (10 classes) (P95 of main thread blocks)`** | **`5948` ms** | 

Need to investigate why it's much slower and why the main thread is getting blocked so much.

Yato333:
Closes #480
- `BaseClassChecker` instance will now be constructed in the `HierarchyBuilder`'s constructor. This will help to prevent unnecessary base class checks when calling `getNodes` the next time. Due to this, `BaseClassChecker` will now have an LRU cache to limit the consumed amount of memory.
- Made it so `isECClassOfBaseECClass` returns a Promise only when the result is not cached.
- Removed some instances of unnecessary traversals of all instance nodes.


packages/hierarchies/src/test/internal/operators/grouping/BaseClassGrouping.test.ts
```diff
@@ -105,7 +105,7 @@ describe("BaseClassGrouping", () => {
         }),
       ];
       const ecClass = { fullName: "TestSchema.ParentClass", label: "ParentClass" } as unknown as ECClass;
-      expect(await baseClassGrouping.createBaseClassGroupsForSingleBaseClass(nodes, ecClass, new BaseClassChecker(metadataProvider))).to.deep.eq({
+      expect(await baseClassGrouping.createBaseClassGroupsForSingleBaseClass(nodes, ecClass, new BaseClassChecker(metadataProvider, 1000))).to.deep.eq({
```
grigasp, line 108: Maybe change `BaseClassChecker` to default the size to `0`? Majority of tests don't care about the cache.

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
```diff
@@ -33,18 +34,18 @@ export function createGroupingOperator(
       log({ category: LOGGING_NAMESPACE, message: /* istanbul ignore next */ (n) => `in: ${createNodeIdentifierForLogging(n)}` }),
       toArray(),
       concatMap((resolvedNodes) => {
+        const { instanceNodes, restNodes } = partitionInstanceNodes(resolvedNodes);
```
grigasp, line 37: Could we pass `instanceNodes` to `createGroupingHandlers` to avoid the need to filter `resolvedNodes` there again?

packages/hierarchies/src/hierarchies/internal/operators/grouping/BaseClassGrouping.ts
```diff
@@ -29,17 +29,20 @@ export async function createBaseClassGroupsForSingleBaseClass(
...
     ) {
       ungroupedNodes.push(node);
       continue;
     }
     const fullCurrentNodeClassName = node.key.instanceKeys[0].className;
 
-    const isCurrentNodeClassOfBase = await baseClassChecker.isECClassOfBaseECClass(fullCurrentNodeClassName, baseECClass);
+    const isCurrentNodeClassOfBase =
+      baseClassChecker.isECClassOfBaseECClassSync(fullCurrentNodeClassName, baseClassFullName) ??
+      (await baseClassChecker.isECClassOfBaseECClass(fullCurrentNodeClassName, baseECClass));
```
grigasp, line 45: ugh, this is ugly, I need to think more about this

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
Yato333: That's right. Done

packages/hierarchies/src/test/internal/operators/grouping/BaseClassGrouping.test.ts
Yato333: Done

grigasp: Sorry, I wasn't clear. My suggestion was to make `cacheSize` argument in the constructor of `BaseClassChecker` option and default to 0. That way tests don't need to pass anything.

Yato333: Ok. Done

packages/hierarchies/src/test/internal/operators/grouping/BaseClassGrouping.test.ts
```diff
@@ -179,7 +179,7 @@ describe("BaseClassGrouping", () => {
       classStubs.stubEntityClass({ schemaName: "TestSchema", className: "A", classLabel: "Class A", is: async () => true });
       const ecClass = { fullName: "TestSchema.ParentClass", label: "ParentClass" } as unknown as ECClass;
 
-      await baseClassGrouping.createBaseClassGroupsForSingleBaseClass(nodes, ecClass, new BaseClassChecker(metadataProvider));
+      await baseClassGrouping.createBaseClassGroupsForSingleBaseClass(nodes, ecClass, new BaseClassChecker(metadataProvider, 1));
```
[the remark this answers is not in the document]
Yato333: We need to have some sort of a cache so that the test would pass:
`it doesn't call getClass when it was called before with the same className`

packages/hierarchies/src/hierarchies/internal/operators/grouping/BaseClassGrouping.ts
Yato333: I don't see the point in caching the promise if it would be awaited as soon as it's returned.
Also, I wouldn't say this looks prettier from the consumer's side:
```js
    const baseCheckerResult = baseClassChecker.isECClassOfBaseECClass(fullCurrentNodeClassName, baseECClass);
    const isCurrentNodeClassOfBase = baseCheckerResult instanceof Promise ? await baseCheckerResult : baseCheckerResult;

    if (isCurrentNodeClassOfBase) {
      groupedNodes.push(node);
    } else {
      ungroupedNodes.push(node);
    }
```

grigasp: > I don't see the point in caching the promise if it would be awaited as soon as it's returned.

In case of multiple simultaneous checks for the same class (without `await`), storing a promise while its result is being resolved helps us reuse the same computation instead of creating a new promise for every call.

> Also, I wouldn't say this looks prettier from the consumer's side.

I see a couple of advantages:
1. Because there's only 1 function call to get the result, we have to create the cache key only once.
2. Having two methods for basically the same purpose, but one returning `undefined` in case the value is not cached exposes internal details of `BaseClassChecker` (the fact that it's caching the result). The suggested solution has only one method that always returns a valid result (either value or a promise to it), so consumer doesn't need to worry about making two calls to get it.

packages/hierarchies/src/test/internal/operators/grouping/BaseClassGrouping.test.ts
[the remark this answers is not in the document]
grigasp: Oh, I see. Thanks.

packages/hierarchies/src/hierarchies/internal/operators/grouping/BaseClassGrouping.ts
Yato333: I see your point now. I've made the suggested changes

grigasp (approved):

PR #488, opened 2024-03-22
Unified selection hilite set API

jasdom:
Added API to `unified-selection` package for getting hilite sets for `Selectables` using ECSql.


packages/unified-selection/src/unified-selection/queries/ECMetadata.ts
```diff
@@ -0,0 +1,147 @@
...
+ * @module UnifiedSelection
+ */
+
+/**
+ * An interface of an iModel metadata provider used to retrieve information about ECSchemas,
+ * ECClasses, ECProperties, etc.
+ *
+ * @beta
+ */
+export interface IMetadataProvider {
```
grigasp, line 16: let's have an internal discussion on how to avoid duplicating all these types

packages/unified-selection/src/unified-selection/SelectionStorage.ts
```diff
@@ -83,6 +86,16 @@ export interface SelectionStorage {
     /** Key of the iModel to change selection for. */
     iModelKey: string;
   }): void;
+
+  /** Get the current hilite set for the specified imodel */
+  getHiliteSet(props: {
```
grigasp, line 91: I'm hesitant on whether we want to have `getHiliteSet` on a `SelectionStorage` interface... I know it's convenient, but maybe we can come up with a way to keep the benefit of "one hilite set request per selection" without `SelectionStorage` knowing about hilite sets?

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+    return new HiliteSetProvider(props);
+  }
+
+  /**
+   * Get hilite set for supplied `Selectables`.
+   */
+  public async getHiliteSet(selection: Selectables): Promise<HiliteSet> {
+    // Map between element ID types and EC instance IDs
+    const keysByType = new Map<InstanceIdType, string[]>();
+    if (!this._hasFunctionalSchema) {
```
grigasp, line 64: shouldn't this check for `undefined` instead?

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+      return cachedType;
+    }
+
+    const keyClass = await this.getClass(key.className);
+    if (!keyClass) {
+      return "unknown";
+    }
+
+    if (this._hasFunctionalSchema) {
+      const functionalType = await this.checkType(key, keyClass, "Functional", "FunctionalElement", "functionalElement");
```
grigasp, line 114: What would `checkType` return for `Functional.FunctionalElement` if there was no Functional schema in the iModel? Maybe we don't need the `_hasFunctionalSchema` check?

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+    map.set(keyType, array);
+  }
+
+  private async getType(key: SelectableInstanceKey): Promise<InstanceIdType> {
+    const cachedType = this._classRelationCache.get(key.className);
+    if (cachedType) {
+      return cachedType;
+    }
+
+    const keyClass = await this.getClass(key.className);
```
grigasp, line 108: I think this could be moved to `checkType`

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+    const bindings: ECSqlBinding[] = [
+      { type: "idset", value: subjectKeys },
+      { type: "idset", value: modelKeys },
+    ];
+    const query = `WITH RECURSIVE
+                    ChildSubjects (ECInstanceId) AS (
+                      SELECT ECInstanceId FROM BisCore.Subject WHERE InVirtualSet(?, ECInstanceId)
+                      UNION ALL
+                      SELECT r.TargetECInstanceId FROM ChildSubjects s
+                        JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 157: it would be more efficient to join `bis.Subject` using its `Parent` navigation property

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+                    ),
+                    Models (ECInstanceId) AS (
+                      SELECT s.ECInstanceId as \`ECInstanceId\` FROM BisCore.Model s
+                      WHERE InVirtualSet(?, ECInstanceId)
+                    )
+                    SELECT r.TargetECInstanceId as \`ECInstanceId\` FROM ChildSubjects s
+                      JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
+                      WHERE r.TargetECClassId IS (BisCore.PhysicalPartition)
+                    UNION
+                    SELECT DISTINCT ECInstanceId FROM Models;`;
```
grigasp, line 168: why do we need `distinct` here?

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+                        JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
+                      WHERE r.TargetECClassId IS (BisCore.Subject)
+                    ),
+                    Models (ECInstanceId) AS (
+                      SELECT s.ECInstanceId as \`ECInstanceId\` FROM BisCore.Model s
+                      WHERE InVirtualSet(?, ECInstanceId)
+                    )
+                    SELECT r.TargetECInstanceId as \`ECInstanceId\` FROM ChildSubjects s
+                      JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
+                      WHERE r.TargetECClassId IS (BisCore.PhysicalPartition)
```
grigasp, line 166: would be more efficient to join `bis.PhysicalPartition` using its `Parent` navigation property

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+                        JOIN BisCore.CategoryOwnsSubCategories r ON r.SourceECInstanceId = s.ECInstanceId
+                      WHERE r.TargetECClassId IS (BisCore.SubCategory) AND InVirtualSet(?, s.ECInstanceId)
+                    ),
+                    SubCategories (ECInstanceId) AS (
+                      SELECT s.ECInstanceId as \`ECInstanceId\` FROM BisCore.SubCategory s
+                      WHERE InVirtualSet(?, s.ECInstanceId)
+                    )
+                   SELECT DISTINCT ECInstanceId FROM CategorySubCategories
+                   UNION
+                   SELECT DISTINCT ECInstanceId FROM SubCategories;`;
```
grigasp, line 197: I don't think we need `distinct` here

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+
+    return this.executeQuery(query, bindings);
+  }
+
+  private async getHilitedElements(map: Map<InstanceIdType, string[]>): Promise<string[]> {
+    const groupInformationElementKeys = map.get("groupInformationElement") ?? [];
+    let geometricElementKeySet = new Set(map.get("geometricElement") ?? []);
```
grigasp, line 204: why this one is a `Set`?

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+
+    if (groupInformationElementKeys.length === 0 && geometricElementKeys.length === 0 && elementKeys.length === 0) {
+      return [];
+    }
+
+    const bindings: ECSqlBinding[] = [
+      { type: "idset", value: groupInformationElementKeys },
+      { type: "idset", value: elementKeys },
+      { type: "idset", value: geometricElementKeys },
+    ];
```
grigasp, line 222: `InVirtualSet` is inefficient to use on a table that is likely to have many rows. Elements' table is one such table - we should prefer `IN` operator unless there are too many IDs (e.g. less than 1000 IDs - use `IN`, otherwise fallback to `InVirtualSet`).

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+    const query = `WITH RECURSIVE
+                    GroupMembers (ECInstanceId, ECClassId) AS (
+                      SELECT r.TargetECInstanceId, r.TargetECClassId FROM BisCore.GroupInformationElement s
+                        JOIN BisCore.ElementGroupsMembers r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 226: looks like selecting from `bis.ElementGroupsMembers` would be enough - no need to join `bis.GroupInformationElement`

packages/test-utilities/src/test-utilities/IModelUtils.ts
```diff
@@ -304,6 +307,65 @@ export function insertExternalSourceAspect(
...
+  const modelId = builder.insertModel({
+    classFullName: className,
+    modeledElement: { id: modeledElementId },
+    ...modelProps,
+  });
+  return { className, id: modelId };
+}
+
+export function insertFunctionalElement(
+  props: BaseInstanceInsertProps & { modelId: Id64String; elementId: Id64String; relationship: string } & Partial<
```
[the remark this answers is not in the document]
jasdom: Functional elements represent either a `DrawingGraphic` or a `PhysicalElement` so the id would refer to one of those. The relationship prop refers to the relationship used between the functional element and the represented element. Updated the names and allowed values to be more clear.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,338 @@
...
+                      SELECT ECInstanceId, ECClassId FROM GroupGeometricElements WHERE ECClassId IS (BisCore.GeometricElement)
+                      UNION
+                      SELECT ECInstanceId, ECClassId FROM ElementGeometricElements WHERE ECClassId IS (BisCore.GeometricElement)
+                    ),
+                    GeometricElementGeometricElements (ECInstanceId, ECClassId) AS (
+                      SELECT ECInstanceId, ECClassId FROM MergedElements
+                      UNION ALL
+                      SELECT r.TargetECInstanceId, r.TargetECClassId FROM GeometricElementGeometricElements s
+                        JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
+                    )
```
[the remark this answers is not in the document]
jasdom: It was needed to recurse through passed geometric element children, however I split that into a separate table to avoid traversing the same elements again.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
jasdom: You are right, fixed.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
jasdom: After few additional changes the `_hasFunctionalSchema` was no longer needed so removed it.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
jasdom: That would cause the `getClass` function to be called every time the type is checked. Not sure how expensive the call is but I think it is better as it is currently.

packages/unified-selection/src/unified-selection/SelectionStorage.ts
```diff
@@ -129,6 +153,34 @@ class SelectionStorageImpl implements SelectionStorage {
   public clearStorage({ iModelKey }: { iModelKey: string }): void {
     this.clearSelection({ source: "Clear iModel storage", iModelKey });
     this._storage.delete(iModelKey);
+    this._hiliteSetProviders.delete(iModelKey);
+  }
+
+  public async getHiliteSet(props: { iModelKey: string; queryExecutor: IECSqlQueryExecutor; metadataProvider: IMetadataProvider }): Promise<HiliteSet> {
```
saskliutas, line 159: I dont' think we need to have API not using iterator.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,341 @@
...
+      }
+    }
+
+    for (const entry of selection.custom) {
+      for await (const key of entry[1].loadInstanceKeys()) {
+        await this.addKeyByType(key, keysByType);
+      }
+    }
+
+    yield {
```
saskliutas, line 99: This does not exactly work as iterator as in previous implementation it loads all ids and then emits them in one go.

saskliutas (changes requested):

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -178,53 +162,41 @@ export class HiliteSetProvider {
...
 
-    const bindings: ECSqlBinding[] = [
-      { type: "idset", value: categoryKeys },
-      { type: "idset", value: subCategoryKeys },
-    ];
+    const bindings: ECSqlBinding[] = [];
     const query = `WITH
                     CategorySubCategories (ECInstanceId) AS (
                       SELECT r.TargetECInstanceId as \`ECInstanceId\` FROM BisCore.Category s
                         JOIN BisCore.CategoryOwnsSubCategories r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 169: Could use `Parent` navigation property to join `SubCategory` instead

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -178,53 +162,41 @@ export class HiliteSetProvider {
...
-    ];
+    const bindings: ECSqlBinding[] = [];
     const query = `WITH RECURSIVE
+                    ${hasFunctionalElements ? this.getHilitedFunctionalElementsQuery(functionalElements, bindings) : ""}
                     GroupMembers (ECInstanceId, ECClassId) AS (
-                      SELECT r.TargetECInstanceId, r.TargetECClassId FROM BisCore.GroupInformationElement s
-                        JOIN BisCore.ElementGroupsMembers r ON r.SourceECInstanceId = s.ECInstanceId
-                      WHERE InVirtualSet(?, s.ECInstanceId)
+                      SELECT TargetECInstanceId, TargetECClassId FROM BisCore.ElementGroupsMembers
+                      WHERE ${this.formBindings("ECInstanceId", groupInformationElementKeys, bindings)}
```
grigasp, line 199: I believe we should be using `SourceECInstanceId` here. If I'm right, it suggests we're missing a full stack tests for this case.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -233,73 +205,80 @@ export class HiliteSetProvider {
                         JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 205: Better to join `bis.Element` using the `Parent` navigation property

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -233,73 +205,80 @@ export class HiliteSetProvider {
...
                     ),
                     ElementGeometricElements (ECInstanceId, ECClassId) AS (
-                      SELECT ECInstanceId, ECClassId FROM BisCore.Element WHERE InVirtualSet(?, ECInstanceId)
+                      SELECT ECInstanceId, ECClassId FROM BisCore.Element WHERE ${this.formBindings("ECInstanceId", elementKeys, bindings)}
                       UNION ALL
                       SELECT r.TargetECInstanceId, r.TargetECClassId FROM ElementGeometricElements s
                         JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 211: Better to join `bis.Element` using the `Parent` navigation property

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -233,73 +205,80 @@ export class HiliteSetProvider {
...
-                      SELECT ECInstanceId, ECClassId FROM GroupGeometricElements WHERE ECClassId IS (BisCore.GeometricElement)
-                      UNION
-                      SELECT ECInstanceId, ECClassId FROM ElementGeometricElements WHERE ECClassId IS (BisCore.GeometricElement)
-                    ),
                     GeometricElementGeometricElements (ECInstanceId, ECClassId) AS (
-                      SELECT ECInstanceId, ECClassId FROM MergedElements
+                      SELECT ECInstanceId, ECClassId FROM BisCore.GeometricElement WHERE ${this.formBindings("ECInstanceId", geometricElementKeys, bindings)}
                       UNION ALL
                       SELECT r.TargetECInstanceId, r.TargetECClassId FROM GeometricElementGeometricElements s
                         JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 217: Better to join `bis.Element` using the `Parent` navigation property

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -233,73 +205,80 @@ export class HiliteSetProvider {
...
+            ),
+            DrawingGraphicElements (ECInstanceId, ECClassId) AS (
+              SELECT r.SourceECInstanceId, r.SourceECClassId FROM ChildFunctionalElements s
+                JOIN Functional.DrawingGraphicRepresentsFunctionalElement r ON r.TargetECInstanceId = s.ECInstanceId
+            ),
+            PhysicalElementGeometricElements (ECInstanceId, ECClassId) AS (
+              SELECT ECInstanceId, ECClassId FROM PhysicalElements
+              UNION ALL
+              SELECT r.TargetECInstanceId, r.TargetECClassId FROM PhysicalElementGeometricElements s
+                JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 255: Better to join `bis.Element` using the `Parent` navigation property

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -233,73 +205,80 @@ export class HiliteSetProvider {
...
+            ),
+            DrawingGraphicElementGeometricElements (ECInstanceId, ECClassId) AS (
+              SELECT ECInstanceId, ECClassId FROM DrawingGraphicElements
+              UNION ALL
+              SELECT r.TargetECInstanceId, r.TargetECClassId FROM DrawingGraphicElementGeometricElements s
+                JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
```
grigasp, line 261: Better to join `bis.Element` using the `Parent` navigation property

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -233,73 +205,80 @@ export class HiliteSetProvider {
...
-                      UNION ALL
-                      SELECT r.TargetECInstanceId FROM DrawingGraphicElementGeometricElements s
-                        JOIN BisCore.ElementOwnsChildElements r ON r.SourceECInstanceId = s.ECInstanceId
-                      WHERE r.TargetECClassId IS (BisCore.GeometricElement)
-                    )
-                   SELECT DISTINCT ECInstanceId FROM PhysicalElementGeometricElements
-                   UNION
-                   SELECT DISTINCT ECInstanceId FROM DrawingGraphicElementGeometricElements;`;
+    if (ids.length === 0) {
+      return `${property} IN (-1)`;
```
grigasp, line 277:
```suggestion
      return `FALSE`;
```

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
grigasp: I think this would be an ideal solution:
- `getHilitedModels`, `getHilitedSubCategories` and `getHilitedElements` return `Observable<Id64String>`.
- we subscribe to all of them and accumulate the values for fixed time intervals - on each interval we emit an accumulated hilite set if it's not empty.

What do you think?

packages/unified-selection/src/unified-selection/SelectionStorage.ts
grigasp: > Could you elaborate on the "one hilite set request per selection" part? 

The current implementation in `@itwin/presentation-frontend`, as well as yours here, cache hilite set per iModel. In addition, `HiliteSetProvider` in `@itwin/presentation-frontend` caches the result based on input keys - requesting hilite set for the same keys on the same `HiliteSetProvider` doesn't require re-running the requests/queries. This is an important piece of functionality when retrieving hilite sets for selection, because a hilite set may be requested multiple times by multiple different components on a selection change, for the same selected items.

On the other hand, I'd like selection storage to not know anything about hilite sets. Ideally, hilite sets should also not know anything about selection storage and just know about `Selectable`. Maybe we need another concept that takes care of caching hilite sets and invalidating them on selection change?..

packages/unified-selection/src/unified-selection/queries/ECMetadata.ts
grigasp: Summary of the discussion:
- For the scope of this PR, we keep these types as duplicates here, but make sure they're not exported and don't have release tags.
- In the long term, we'll need to come up with a way to share some code between packages - either through a shared package or shared code that gets transpiled into each package that needs it.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
saskliutas: I agree. We could also look into streaming instance keys from selectables instead of loading them all before running any query as `CustomSelectable.loadInstanceKeys` might produce a lot of keys.

So we can leave this as is for now and comeback to it later.

grigasp: Added a backlog item: https://github.com/iTwin/presentation/issues/501.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
```diff
@@ -0,0 +1,98 @@
...
+      this._cache.set(iModelKey, hiliteSet);
+    }
+
+    return eachValueFrom(hiliteSet);
+  }
+
+  public clearCache({ iModelKey }: { iModelKey: string }): void {
+    this._hiliteSetProviders.delete(iModelKey);
+    this._cache.delete(iModelKey);
+  }
```
grigasp, line 82: I don't think we need a `clearCache` function - on iModel close, whoever created the `SelectionStorage`, should call `clearStorage`, which will cause a selection clear for that iModel, at which point we're going to get a selection change event and clear the cache.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /** Clears hilite set cache for a given iModel */
+  clearCache(props: {
+    /** Key of the iModel to change selection for */
+    iModelKey: string;
+  }): void;
+
+  /** Disposes the cache. */
+  dispose(): void;
+}
```
grigasp, line 42: I'd like the API of this to be similar to `HiliteSetProvider`:
```ts
interface CachingHiliteSetProvider extends HiliteSetProvider {
    dispose(): void;
}
```

I understand we need query executor and metadata provider, which we won't get on selection changes, but I think we could request a getter in the "create" function, e.g.:
```ts
function createCachingSelectionHiliteSetProvider(props: {
    selectionStorage: SelectionStorage;
    iModelProvider: (iModelKey: string): { queryExecutor: IECSqlQueryExecutor; metadataProvider: IMetadataProvider };
}): CachingHiliteSetProvider {
    // return slightly modified HiliteSetCacheImpl
}
```

grigasp: I missed one part of my suggestion:
```ts
interface HiliteSetProvider {
  getHiliteSet(props: { iModelKey: string }): AsyncIterableIterator<HiliteSet>;
}
```

- Existing `HiliteSetProvider` class becomes `HiliteSetProviderImpl`.
- `HiliteSetProvider.create` becomes a root level `createHiliteSetProvider`.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
jasdom: The event would cause the hilite set cache to be cleared, however it would not clear the `_hiliteSetProviders` map, which is why I added this function.

grigasp: Do we need to cache providers?

jasdom: They cache mappings between class names and the types they resolve to that are needed for the queries. Other than that - no.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
jasdom: The existing `getHiliteSet` in `HiliteSetProvider` requires `Selectables` as argument rather than `iModelKey`, while the `CachingHiliteSetProvider` requires `iModelKey`.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
grigasp: I think those class / type mappings justify caching providers. But I would still like to avoid requiring a `clearCache` call from our consumers... What do you think about using the `StorageSelectionChangeEventArgs.source` attribute to find out whether it's a regular "clear" event or a "clear because of iModel close" event?

jasdom: That was an option I considered at first but decided against it, because the source attribute comes from what is provided to the public `SelectionStorage` methods so it seemed like a hacky and fragile solution.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
grigasp: Oh, okay - I missed that piece... yeah, sending selectables to this - based on selection - one doesn't make sense. 

What I'd like to avoid is consumers having to pass those 3 arguments to get a hilite set - that's going to be very inconvenient... `iModelKey` is okay, but the other two aren't.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
grigasp: The string comes from the same package where we handle it, so we can use some internal variable to make sure changing it in one place doesn't break the other. If we make it enough unique, we shouldn't have issues with the same string being used from elsewhere. Example: `Unified selection storage: clear`.

jasdom: Removed `clearCache` and switched to clearing based on event source.

packages/unified-selection/src/unified-selection/HiliteSetCache.ts
jasdom: Refactored to not need to pass the `queryExecutor` and `metadataProvider` every call.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,324 @@
...
+
+/**
+ * Creates a hilite set provider.
+ * @beta
+ */
+export function createHiliteSetProvider(props: HiliteSetProviderProps): HiliteSetProvider {
+  return new HiliteSetProviderImpl(props);
+}
+
+class HiliteSetProviderImpl implements HiliteSetProvider {
```
saskliutas, line 51: Looks like this is used only internally. In that case I think we don't need `HiliteSetProvider` interface and `createHiliteSetProvider` function. We can use `HiliteSetProviderImpl` class directly in `CachingHiliteSetProvider`.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -0,0 +1,324 @@
...
+
+  private async addKeyByType(key: SelectableInstanceKey, map: Map<string, string[]>) {
+    const keyType = await this.getType(key);
+    let array = map.get(keyType);
+    if (!array) {
+      array = [];
+    }
+
+    array.push(key.id);
+    map.set(keyType, array);
```
saskliutas, line 97: I think `map.set()` can be moved inside `if` statement.

packages/unified-selection/src/test/HiliteSetProvider.test.ts
```diff
@@ -0,0 +1,339 @@
...
+        subCategories,
+        elements,
+      };
+    }
+
+    beforeEach(() => {
+      queryExecutor.createQueryReader.reset();
+      metadataProvider.getSchema.reset();
+
+      schemaMock = {
```
saskliutas, line 94: Why not `schemaMock.getClass.reset()`?

packages/unified-selection/src/test/HiliteSetProvider.test.ts
```diff
@@ -0,0 +1,339 @@
...
+
+      it("caches class type", async () => {
+        const classMock = {
+          is: (baseClass: string, baseSchema: string) => {
+            return baseClass === "Element" && baseSchema === "BisCore";
+          },
+        };
+
+        schemaMock.getClass.onCall(0).returns(Promise.resolve(classMock as unknown as ECClass));
+        schemaMock.getClass.onCall(1).throws();
```
saskliutas, line 304: I think it would make test cleaner to assert `getClass` was called 1 time instead of mocking second call as error.

packages/unified-selection/src/test/CachingHiliteSetProvider.test.ts
```diff
@@ -0,0 +1,152 @@
...
+    selectionStorage = createStorage();
+
+    hiliteSetCache = createCachingHiliteSetProvider({
+      selectionStorage,
+      iModelProvider,
+    });
+    iModelProvider.returns({ queryExecutor: {} as IECSqlQueryExecutor, metadataProvider: {} as IMetadataProvider });
+    selectionStorage.addToSelection({ iModelKey, source: "test", selectables: generateSelection() });
+
+    provider = {
```
saskliutas, line 60: Why not `provider.getHiliteSet.reset()`?

saskliutas (changes requested):

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
jasdom: `HiliteSetProvider` was supposed to be exported, fixed.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -35,39 +35,45 @@ export interface HiliteSetProviderProps {
...
+  getHiliteSet(props: { selectables: Selectables }): AsyncIterableIterator<HiliteSet>;
+}
+
+/**
+ * Creates a hilite set provider.
+ * @beta
+ */
+export function createHiliteSetProvider(props: HiliteSetProviderProps): HiliteSetProvider {
+  return new HiliteSetProviderImpl(props);
+}
```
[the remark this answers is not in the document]
jasdom: Are they not already exported?

packages/unified-selection/src/test/CachingHiliteSetProvider.test.ts
```diff
@@ -78,27 +78,26 @@ describe("HiliteSetCache", () => {
       const executor2 = {} as IECSqlQueryExecutor;
       const metadataProvider = {} as IMetadataProvider;
 
-      hiliteSetCache.getHiliteSet({ iModelKey: "model1", queryExecutor: executor1, metadataProvider });
+      hiliteSetCache.getHiliteSet({ iModelKey: "model1" });
+      iModelProvider.returns({ queryExecutor: executor1, metadataProvider });
```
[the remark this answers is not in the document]
jasdom: Did not realize that `calledOnceWith` does a deep equality check. Modified the executor values to be different.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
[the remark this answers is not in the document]
grigasp: my bad, somehow missed them

grigasp (commented):
LGTM

saskliutas (approved):

PR #489, opened 2024-03-25
Tooltips for node filter buttons

jasdom:
Closes [470](https://github.com/iTwin/presentation/issues/470)

Added tooltips for node buttons "Clear filter" and "Open filtering dialog".


packages/hierarchies-react/src/presentation-hierarchies-react/Tree.tsx
```diff
@@ -50,6 +50,7 @@ export function TreeRenderer({
                 className="filtering-action-button"
                 styleType="borderless"
                 size="small"
+                title="Clear active filter"
```
grigasp, line 53: need to think about our localization strategy in this package

grigasp: filed a separate issue for tracking localization: https://github.com/iTwin/presentation/issues/490

.changeset/wise-kangaroos-thank.md
saskliutas, line 1: No need to create change files for this package as it is not release yet.

saskliutas (approved):

PR #493, opened 2024-03-25
Update tree unified selection handling

saskliutas:
Updated unified selection handling in tree to work when selection is not updated immediately after modification.

### DIFF
--- a/packages/components/src/test/tree/controlled/UseUnifiedSelection.test.ts
+++ b/packages/components/src/test/tree/controlled/UseUnifiedSelection.test.ts
@@ -22,13 +22,13 @@ import {
 import { using } from "@itwin/core-bentley";
 import { EmptyLocalization } from "@itwin/core-common";
 import { IModelConnection } from "@itwin/core-frontend";
-import { KeySet, NodeKey } from "@itwin/presentation-common";
+import { Key, KeySet, NodeKey } from "@itwin/presentation-common";
 import {
   ISelectionProvider,
   Presentation,
+  SelectionChangeEvent,
   SelectionChangeEventArgs,
   SelectionChangeType,
-  SelectionHandler,
   SelectionHelper,
   SelectionManager,
 } from "@itwin/presentation-frontend";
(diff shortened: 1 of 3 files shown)


grigasp (approved):

packages/components/src/test/tree/controlled/UseUnifiedSelection.test.ts
```diff
@@ -399,43 +400,123 @@ describe("UnifiedSelectionEventHandler", () => {
...
+      selectionManager.getSelection.returns(new KeySet());
+
+      using(createHandler(), (_) => {
+        // verify nodes selected based on initial unified selection
+        expect(modelSource.getModel().getNode(nodes[0].id)?.isSelected).to.be.false;
+        expect(modelSource.getModel().getNode(nodes[1].id)?.isSelected).to.be.false;
+
+        const selectionKeys = SelectionHelper.getKeysForSelection(nodes.map((n) => getItemKey(n.item)));
+        selectionManager.getSelection.reset();
+        selectionManager.getSelection.returns(new KeySet(selectionKeys));
```
grigasp, line 487: not sure `getSelection` needs to return a different result here - if the selection changes for a different iModel, then `getSelection` is going to return its old result

packages/components/src/test/tree/controlled/UseUnifiedSelection.test.ts
```diff
@@ -381,55 +400,125 @@ describe("UnifiedSelectionEventHandler", () => {
...
 
+        // verify selection change event was ignored
         expect(modelSource.getModel().getNode(nodes[0].id)?.isSelected).to.be.false;
         expect(modelSource.getModel().getNode(nodes[1].id)?.isSelected).to.be.false;
       });
     });
 
-    it("skips unified selection change if it is change event source", () => {
-      Object.assign(selectionHandler, { name: selectionEvent.source });
+    it("ignores selection changes on different ruleset", () => {
```
grigasp, line 496: I don't think we should care about the ruleset that's associated with the event

grigasp (approved):

Issue #469, opened 2024-03-12
"Contains" operator in instance filter builder should create a "like" rule with `%` on both sides of value


grigasp:
The "Contains" operator should match all entries that contain given string, so user should not be forced to enter `%` on both ends of the string. On the other hand, the `GenericInstanceFilter` stores the `like` operator, which may or may not have the `%` on either end of the string. So I think `%` should be added to both ends of the string when `GenericInstanceFilter` is created using `Contains` operator.

jasdom:
After some discussions decided that we cannot correctly handle all possible cases when converting `PresentationInstanceFilter` <-> `GenericInstanceFilter`. It's better to leave escaping/adding wildcard characters at output generation step. Added notes to `PresentationInstanceFilter` and `GenericInstanceFilter` that like operator should be handled as a contains operator: [6570](https://github.com/iTwin/itwinjs-core/pull/6570), [789](https://github.com/iTwin/appui/pull/789).

PR #496, opened 2024-03-26
Match all entries that contain given string by default in instance filter builder

jasdom:
Closes [459](https://github.com/iTwin/presentation/issues/469)

Modified instance filter builder to automatically append `%` to both sides of the value, to match all entries that contain given string by default.

### DIFF
--- a/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
+++ b/packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
@@ -228,7 +228,7 @@ function createGenericInstanceFilterRule(condition: PresentationInstanceFilterCo
 
   return {
     operator,
-    value: toGenericInstanceFilterRuleValue(value),
+    value: toGenericInstanceFilterRuleValue(operator, value),
     sourceAlias: propertyAlias,
     propertyName: property.name,
     propertyTypeName: field.type.typeName,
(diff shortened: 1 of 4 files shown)


packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
```diff
@@ -499,12 +499,27 @@ function toRelationshipStep(path: RelationshipPath): GenericInstanceFilterRelati
...
     return undefined;
   }
 
-  return { displayValue: primitiveValue.displayValue ?? "", rawValue: primitiveValue.value };
+  const displayValue = primitiveValue.displayValue ?? "";
+  let rawValue = primitiveValue.value;
+
+  if (operator === "like" && typeof rawValue === "string") {
+    if (!rawValue.startsWith("%")) {
+      rawValue = `%${rawValue}`;
```
saskliutas, line 515: How does the round trip works? `PresentationInstanceFilter` -> `GenericInstanceFilter` -> `PresentationInstanceFilter`. From current implementation it looks like if `someValue` was entered into UI then it was converted to `GenericInstanceFilter` and back, UI will be populated with `%someValue%`?

jasdom: You are right. Removed the appended `%` when parsing the filter.

saskliutas: I am not sure that this is enough. What about entering `%someValue` into UI and doing same roundtrip?

jasdom: In that case after reopening the filter `someValue` will be displayed, which I think is the expected behavior considering that manually inserting `%` at the beginning or the end has no effect on the filter. Though in that case adding a bunch of them at the start would cause them to be removed one by one instead of removing all of them.

saskliutas: I am not sure that this is an expected behavior. I think UI should show same value that was entered without modifying it.

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
```diff
@@ -354,11 +354,22 @@ function parseGenericFilterRuleGroup(
...
+  let value = rule.value ? rule.value.rawValue : undefined;
+
+  if (value && rule.operator === "like" && typeof value === "string") {
+    if (value.startsWith("%")) {
+      value = value.slice(1);
+    }
+
+    if (value.endsWith("%")) {
+      value = value.slice(0, -1);
+    }
```
saskliutas, line 366: I think you can use regex.

```suggestion
    value = value.replace(/^%|%$/g, "");
```

saskliutas (changes requested):

saskliutas:
After some discussions decided that we cannot correctly handle all possible cases when converting `PresentationInstanceFilter` <-> `GenericInstanceFilter`. It's better to leave escaping/adding wildcard characters at output generation step. However, we should add note to `GenericInstanceFilterRuleOperator` and `PropertyFilterRuleOperator` types that `like` operator should be handled as `contains`.

Issue #434, opened 2024-02-26
Investigate possibility to create only one level of grouping at a time


grigasp:
Currently all groupings are created when we receive instance nodes. This has a downside of possibly doing unnecessary work, because users might not even expand to all deeper grouping levels. We should look into creating only 1 grouping level at a time, only when parent node is expanded.

PR #498, opened 2024-03-26
Hierarchies: Grouping refactor

grigasp:
Closes https://github.com/iTwin/presentation/issues/434

### DIFF
--- a/packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
+++ b/packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
@@ -4,8 +4,16 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { concatAll, concatMap, from, Observable, of, toArray } from "rxjs";
+import { assert } from "@itwin/core-bentley";
 import { IMetadataProvider } from "../../ECMetadata";
-import { HierarchyNode, HierarchyNodeKey, ProcessedGroupingHierarchyNode, ProcessedHierarchyNode, ProcessedInstanceHierarchyNode } from "../../HierarchyNode";
+import {
+  HierarchyNode,
+  HierarchyNodeKey,
+  ParentHierarchyNode,
+  ProcessedGroupingHierarchyNode,
+  ProcessedHierarchyNode,
+  ProcessedInstanceHierarchyNode,
+} from "../../HierarchyNode";
 import { IPrimitiveValueFormatter } from "../../values/Formatting";
 import { BaseClassChecker, compareNodesByLabel, createNodeIdentifierForLogging, createOperatorLoggingNamespace, mergeSortedArrays } from "../Common";
 import { log } from "../LoggingUtils";
@@ -23,6 +31,7 @@ export const LOGGING_NAMESPACE = createOperatorLoggingNamespace(OPERATOR_NAME);
 /** @internal */
 export function createGroupingOperator(
   metadata: IMetadataProvider,
+  parentNode: ParentHierarchyNode | undefined,
   valueFormatter: IPrimitiveValueFormatter,
   localizedStrings: PropertiesGroupingLocalizedStrings,
   baseClassChecker: BaseClassChecker,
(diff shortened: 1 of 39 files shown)


packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
```diff
@@ -88,34 +100,33 @@ async function groupInstanceNodes(
   nodes: ProcessedInstanceHierarchyNode[],
   extraSiblings: number,
   groupingHandlers: GroupingHandler[],
-  onGroupingNodeCreated: undefined | ((groupingNode: ProcessedGroupingHierarchyNode) => void),
+  parentNode: ParentHierarchyNode | undefined,
```
Yato333, line 103:
```suggestion
  parentNode?: ParentHierarchyNode,
```

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
```diff
@@ -88,34 +100,33 @@ async function groupInstanceNodes(
...
-            }),
-          ),
-        ),
-        compareNodesByLabel,
-      ),
+      grouped: [...(curr?.grouped ?? []), ...groupings.grouped],
       ungrouped: groupings.ungrouped,
     };
   }
   if (curr) {
```
Yato333, line 115: We can reduce the amount of nesting here. How about this:
```ts
  if (!curr) {
    return nodes;
  }

  for (const groupingNode of curr.grouped) {
    onGroupingNodeCreated?.(groupingNode);
    if (!parentNode) {
      continue;
    }

    if (HierarchyNode.isGroupingNode(parentNode)) {
      groupingNode.nonGroupingAncestor = parentNode.nonGroupingAncestor;
      continue;
    }

    // not sure why type checker doesn't pick this up
    assert(HierarchyNode.isCustom(parentNode) || HierarchyNode.isInstancesNode(parentNode));
    groupingNode.nonGroupingAncestor = parentNode;
  }

  return [...curr.grouped, ...curr.ungrouped];
```

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
grigasp: In this context parent node is not optional - `undefined` means we're requesting root nodes. So I'd prefer to leave it as is.

packages/hierarchies/src/hierarchies/internal/operators/grouping/PropertiesGrouping.ts
```diff
@@ -44,18 +46,45 @@ export interface PropertyGroupInfo {
...
+            properties: [],
+          },
+          parentKeys: [],
+          groupedInstanceKeys: [],
+          label: localizedStrings.other,
+          children: [],
+        },
+        new: true,
+      };
+    }
```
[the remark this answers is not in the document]
grigasp: applied

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
grigasp: agree, fixed

packages/hierarchies/src/hierarchies/internal/operators/grouping/LabelGrouping.ts
```diff
@@ -57,8 +56,8 @@ export async function createLabelGroups(nodes: ProcessedInstanceHierarchyNode[])
   });
 
   return {
-    grouped: sortNodesByLabel(groupedNodes),
-    ungrouped: mergeSortedArrays(ungrouped, sortNodesByLabel(mergedNodes), compareNodesByLabel),
+    grouped: groupedNodes,
+    ungrouped: [...ungrouped, ...mergedNodes],
     groupingType: "label",
   };
```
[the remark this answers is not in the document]
grigasp: applied suggestion here and in some other places too

Yato333 (approved):

Issue #477, opened 2024-03-14
Provide a way to create initial value for `UniquePropertyValuesSelector`


saskliutas:
At the moment there is no easy way to create initial `UniquePropertyValuesSelector` value from `PropertyValue`s created property grid data provider. This makes it difficult to create initial `PresentationInstanceFilter` using `Equal/Not Equal` operators and `UniquePropertyValuesSelector`

PR #499, opened 2024-03-26
Add API for creating filter conditions compatible with unique values selector

saskliutas:
Closes https://github.com/iTwin/presentation/issues/477

### DIFF
--- a/.changeset/metal-tables-wonder.md
+++ b/.changeset/metal-tables-wonder.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Added `PresentationInstanceFilter.createEqualityCondition` function to provide a convienent way for creating equality filter conditions compatible with `UniquePropertyValuesSelector`.
--- a/.changeset/metal-tables-wonder.md
+++ b/.changeset/metal-tables-wonder.md
@@ -1,5 +1,5 @@
 ---
-"@itwin/presentation-components": patch
+"@itwin/presentation-components": minor
 ---
 
 Added `PresentationInstanceFilter.createEqualityCondition` function to provide a convienent way for creating equality filter conditions compatible with `UniquePropertyValuesSelector`.
--- a/.changeset/metal-tables-wonder.md
+++ b/.changeset/metal-tables-wonder.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-components": minor
 ---
 
-Added `PresentationInstanceFilter.createEqualityCondition` function to provide a convenient way for creating equality filter conditions compatible with `UniquePropertyValuesSelector`.
+Added `PresentationInstanceFilter.createPrimitiveValueEqualityCondition` function to provide a convenient way for creating equality filter conditions compatible with `UniquePropertyValuesSelector`.
--- a/.changeset/metal-tables-wonder.md
+++ b/.changeset/metal-tables-wonder.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-components": minor
 ---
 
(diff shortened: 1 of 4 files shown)


.changeset/metal-tables-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
```
grigasp, line 2: shouldn't this be a minor?

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
```diff
@@ -529,3 +546,20 @@ function isPoint3dLike(value: object): value is { x: number; y: number; z: numbe
 function isInstanceKeyLike(value: object): value is { id: string; className: string } {
   return (value as any).id !== undefined && (value as any).className !== undefined;
 }
+
+function createUniqueValue(value: PropertyValue): PrimitiveValue | undefined {
+  if (value.valueFormat !== PropertyValueFormat.Primitive || value.displayValue === undefined || value.value === undefined) {
+    return undefined;
```
grigasp, line 552: doesn't this mean that we were given an invalid `PrimitiveValue`? maybe we should throw instead?

saskliutas: I guess we could throw if `valueFormat` is not `Primitive`. But as this is an utility function for creating conditions using unique values selector, I think makes it more convenient for consumer not throwing error.

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
```diff
@@ -547,10 +555,11 @@ function isInstanceKeyLike(value: object): value is { id: string; className: str
   return (value as any).id !== undefined && (value as any).className !== undefined;
 }
 
-function createUniqueValue(value: PropertyValue): PrimitiveValue | undefined {
-  if (value.valueFormat !== PropertyValueFormat.Primitive || value.displayValue === undefined || value.value === undefined) {
-    return undefined;
-  }
+function isValidPrimitiveValue(val: PrimitiveValue): val is Required<PrimitiveValue> {
+  return val.valueFormat === PropertyValueFormat.Primitive && val.value !== undefined && val.displayValue !== undefined;
```
grigasp, line 559: is `val.valueFormat === PropertyValueFormat.Primitive` check necessary anymore?

packages/components/src/test/instance-filter-builder/PresentationInstanceFilter.test.ts
```diff
@@ -985,27 +985,34 @@ describe("PresentationInstanceFilter", () => {
     });
   });
 
-  describe("createEqualityCondition", () => {
+  describe("createPrimiiveValueEqualityCondition", () => {
```
grigasp, line 988:
```suggestion
  describe("createPrimitiveValueEqualityCondition", () => {
```

.changeset/metal-tables-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": minor
+---
+
+Added `PresentationInstanceFilter.createEqualityCondition` function to provide a convenient way for creating equality filter conditions compatible with `UniquePropertyValuesSelector`.
```
grigasp, line 5:
```suggestion
Added `PresentationInstanceFilter.createPrimitiveValueEqualityCondition` function to provide a convenient way for creating equality filter conditions compatible with `UniquePropertyValuesSelector`.
```

grigasp (approved):

Issue #472, opened 2024-03-12
Errors when creating hierarchy levels are not displayed by the tree component


grigasp:
When filtering child hierarchy level:
![Image](https://github.com/iTwin/presentation/assets/35135765/af6ada97-30e8-4514-b1c4-603f9db250d9)

When expanding a node:
![Image](https://github.com/iTwin/presentation/assets/35135765/7d710c98-1e81-4019-8c3f-e5e232023ac0)

I would expect to see `NonIdealState` in case of an error creating a hierarchy level.

PR #502, opened 2024-03-28
Hierarchy level creation error handling

jasdom:
Closes [472](https://github.com/iTwin/presentation/issues/472)

Hierarchy level creation errors are now displayed in tree component using `TreeModelInfoNode`.


packages/core-interop/src/test/Metadata.test.ts
```diff
@@ -63,6 +63,39 @@ describe("createMetadataProvider", () => {
...
+          .stub<[SchemaKey], CoreSchema>()
+          .withArgs(matchSchemaName)
+          .resolves({
+            name: "y",
+          } as unknown as CoreSchema),
+      } as unknown as SchemaContext;
+
+      const provider = createMetadataProvider(schemaContext);
+      await provider.getSchema("x");
+      await provider.getSchema("x");
```
grigasp, line 79: I'd say this should be `await Promise.all([provider.getSchema("x"), provider.getSchema("x")])`

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -46,25 +48,29 @@ export class HierarchyLoader implements IHierarchyLoader {
...
-            parentId: parent.id,
-            type: "ResultSetTooLarge",
-            message: err.message,
-          };
+          node.type = "ResultSetTooLarge";
+          node.message = err.message;
           return of([node]);
         }
-        throw err;
+        return of([node]);
```
grigasp, line 73: Creating an object and overwriting its attributes right away doesn't make much sense, IMO. If you don't want to repeat the same attributes, I suggest doing the following:
```ts
const nodeProps = {
  id: ...,,
  parentId: ...,
};
if (x) {
  return of([{ ...nodeProps, type: "", message: "" }]);
}
return of([{ ...nodeProsp, type: "", message: "" }]);
```

packages/hierarchies/src/hierarchies/HierarchyProvider.ts
```diff
@@ -314,7 +316,7 @@ export class HierarchyProvider {
     const loggingCategory = `${LOGGING_NAMESPACE}.GetCachedObservableEntry`;
     const { parentNode, ...restProps } = props;
     const cached = this._nodesCache.get(props);
-    if (cached) {
+    if (!props.ignoreCache && cached) {
```
grigasp, line 319: Why even request the cache entry if we're going to ignore it?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -59,18 +59,14 @@ export class HierarchyLoader implements IHierarchyLoader {
...
           id: `${parent.id ?? ""}-${err.message}`,
           parentId: parent.id,
-          type: "Unknown",
-          message: "Failed to create hierarchy level",
         };
         if (err instanceof RowsLimitExceededError) {
-          node.type = "ResultSetTooLarge";
-          node.message = err.message;
-          return of([node]);
+          return of([{ ...nodeProps, type: "ResultSetTooLarge", message: err.message } as TreeModelInfoNode]);
```
grigasp, line 67: why do we need `as TreeModelInfoNode`? would `type: "ResultSetTooLarge" as const,` drop this requirement?

jasdom: Yes, that worked, thanks.

grigasp (commented):
`hierarchies` and `core-interop` look good to me, leaving `hierarchies-react` for @saskliutas

saskliutas (approved):
`hierarchies-react` looks good

PR #504, opened 2024-03-29
Fix block handler showing incorrect results

Yato333:
Currently performance tests do not account for long main thread blocks that last until the end of the test. 
Now, before stopping the block handler, we will wait until block handler's callback is run the last time before stopping it.

Block handler will now use a much more convenient node's `timers/promises/setInterval` function which returns an async iterator.
Since `blocked` library is very simple and doesn't provide anything of value,  the dependency will be removed in favor of custom blocking time measuring logic.

Additionally, new optional logging will be added which might help the debugging experience:
- Logging of start and end times of the blocks
- Ability to log "pings" of the main thread

### DIFF
--- a/apps/performance-tests/src/util/BlockHandler.ts
+++ b/apps/performance-tests/src/util/BlockHandler.ts
@@ -33,7 +33,8 @@ export interface Summary {
  */
 export class BlockHandler {
   private readonly _samples = new SortedArray<number>((a, b) => a - b);
-  private _running = true;
+  private _running = false;
+  private _promise?: Promise<void>;
 
   public getSummary(): Summary {
     const arr = this._samples.extractArray();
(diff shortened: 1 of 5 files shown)


apps/performance-tests/src/util/BlockHandler.ts
```diff
@@ -21,7 +33,8 @@ export interface Summary {
  */
 export class BlockHandler {
   private readonly _samples = new SortedArray<number>((a, b) => a - b);
-  private _timer?: NodeJS.Timeout;
+  private _running = false;
```
grigasp, line 36: could we use `_promise` as the indication for a running timer?

grigasp (approved):

apps/performance-tests/src/util/BlockHandler.ts
Yato333: That's right. Made the suggested change

grigasp (approved):

PR #508, opened 2024-04-02
Reduce blocking for flat hierarchies

Yato333:
Add setTimeout(..., 0) in the middle of querying the rows and right after grouping.


packages/hierarchies/src/hierarchies/internal/operators/ReleaseMainThreadOn.ts
```diff
@@ -0,0 +1,36 @@
...
+export const releaseMainThread = async () => {
+  doLog({ category: LOGGING_NAMESPACE, message: /* istanbul ignore next */ () => "Releasing main thread" });
+  return new Promise<void>((resolve) => setTimeout(resolve, 0));
+};
+
+/**
+ * Emits a certain amount of values, then releases the main thread for other timers to use.
+ * @internal
+ */
+export function releaseMainThreadOn<T>(elementCount: number) {
```
grigasp, line 25: Did you follow some convention coming up with this function name? I'd personally rename it to `releaseMainThreadOnItemsCount` or something like that.

packages/hierarchies/src/hierarchies/internal/TreeNodesReader.ts
```diff
@@ -28,8 +31,12 @@ export class TreeQueryResultsReader {
   }
 
   public async *read(queryExecutor: ILimitingECSqlQueryExecutor, query: ECSqlQueryDef, limit?: number | "unbounded"): AsyncGenerator<ParsedHierarchyNode> {
+    let idx = 0;
     for await (const row of queryExecutor.createQueryReader(query, { rowFormat: "ECSqlPropertyNames", ...(limit !== undefined ? { limit } : undefined) })) {
       yield this._props.parser(row);
+      if (idx++ % MAX_ROWS_TO_EMIT_AT_A_TIME === 0) {
```
grigasp, line 37: Any reason not to do this in `createLimitingECSqlQueryExecutor`? That way the behavior would be used for other queries other than reading nodes.

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
```diff
@@ -62,6 +63,7 @@ export function createGroupingOperator(
         );
       }),
       concatAll(),
+      releaseMainThreadOn(100),
```
grigasp, line 66: why are we using a smaller number here compared to the one in `TreeQueryResultsReader`?

grigasp (commented):
good job finding the culprit for these blocks

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
Yato333: This is an optimal value that I determined through experimentation. When I use 1000 here, p95 of blocks increases from 30 ms to 135 ms on my machine. It seems like we are doing much more work for the postprocessed nodes than for those at the beginning.

packages/hierarchies/src/hierarchies/internal/operators/ReleaseMainThreadOn.ts
Yato333: I was going with rxjs-like names but that sounds better, thanks.

packages/hierarchies/src/hierarchies/internal/TreeNodesReader.ts
Yato333: It was simply much easier to implement it here. But since you think it's worth it, I've moved it to the limiting executor.
Also, since there are multiple for loops and I am not sure which max element count to pick, I've implemented a solution with time measuring (which will probably be more accurate than before)

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
grigasp: okay, let's keep it this way for now and look into making more micro-optimizations

grigasp (approved):

Issue #422, opened 2024-02-19
README


grigasp:
Create READMEs for:
- [x] `@itwin/presentation-shared`
- [x] `@itwin/presentation-core-interop`
- [x] `@itwin/presentation-hierarchies`
- [ ] `@itwin/presentation-hierarchies-react`.

PR #509, opened 2024-04-03
core-interop: API cleanup, README, LICENSE

grigasp:
Part of https://github.com/iTwin/presentation/issues/422.

Going to remove `createHierarchyLevelDescriptor` altogether in a separate PR.

### DIFF
--- a/packages/core-interop/src/core-interop/Logging.ts
+++ b/packages/core-interop/src/core-interop/Logging.ts
@@ -3,20 +3,43 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { Logger as CoreLogger, LogLevel as CoreLogLevel } from "@itwin/core-bentley";
+import { LogLevel as CoreLogLevel } from "@itwin/core-bentley";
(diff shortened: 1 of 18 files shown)


packages/core-interop/src/core-interop/Logging.ts
```diff
@@ -3,20 +3,43 @@
...
+interface ICoreLogger {
+  isEnabled(category: string, level: CoreLogLevel): boolean;
+  logError(category: string, message: string): void;
+  logWarning(category: string, message: string): void;
+  logInfo(category: string, message: string): void;
+  logTrace(category: string, message: string): void;
+}
+
+/**
+ * Create an `ILogger` that uses [Logger](https://www.itwinjs.org/reference/core-bentley/logging/logger/) API to log messages.
```
Yato333, line 22: In most cases you wrote "defines" or "registers", etc. Here (and in other cases below) you wrote "create". Is this intentional?

grigasp: No, I just unified the wording. Thanks for noticing.

Yato333 (approved):

PR #514, opened 2024-04-05
core-interop: Remove `createHierarchyLevelDescriptor`

grigasp:
Instead, added `HierarchyProvider.getNodeInstanceKeys` which allows building the descriptor. The new function, compared to `createHierarchyLevelDescriptor` implementation, also fully supports grouping & custom nodes.

Also, made a small optimization that affects the `HideNodesInHierarchy` operator. Internally it used a `reduce` operator which always created a new `Map`, even if there were no nodes. Since I needed a similar behavior outside of `HideNodesInHierarchy`, I added `reduceToMergeMapItem` and `reduceToMergeMapList` operators, which avoid creating a new `Map` instance until there's at least one item to add to it.


packages/hierarchies/src/hierarchies/HierarchyProvider.ts
```diff
@@ -396,6 +426,104 @@ export class HierarchyProvider {
...
+              label: "",
+            })),
+        ),
+      ),
+    );
+
+    // merge visible instance keys from this level & the ones we get recursively requesting from deeper levels
+    const visibleInstanceKeys = merge(
+      visibleNodeInstanceKeys.pipe(map(({ key }) => key)),
+      hiddenParentNodes.pipe(mergeMap((hiddenNode) => this.getNodeInstanceKeys({ parentNode: hiddenNode }))),
```
saskliutas, line 520: Shouldn't this be `from(this.getNodeInstanceKeys({ parentNode: hiddenNode }))`? It looks like there could be a private `getNodeInstanceKeysObs` functions that returns `Observable` instead of iterator to avoid converting iterator -> observable -> iterator?

grigasp: `mergeMap` works with `ObservableInput` and `AsyncIterableIterator` is one, so that was fine. However, I used your suggestion to extract creating observable into a separate function, so now it gets an `Observable` rather than `AsyncIterableIterator`.

saskliutas (approved):

PR #516, opened 2024-04-05
hierarchies: Change `HierarchyProvider.getNodes` from `Promise<HierarchyNode[]>` to `AsyncIterableIterator<HierarchyNode>`

grigasp:
Stay consistent with `getNodeInstanceKeys` and provides ability to stream nodes as they get loaded rather than having to wait for the whole list.


Yato333:
The performance of large flat hierarchy and base class grouping has worsened dramatically. Why would that be?

packages/hierarchies/src/hierarchies/internal/operators/EachValueFrom.ts
```diff
@@ -0,0 +1,128 @@
...
+import { Observable } from "rxjs";
+
+/**
+ * This is pretty much a copy of https://github.com/benlesh/rxjs-for-await/blob/94f9cf9cb015ac3700dfd1850eb81d36962eb70f/src/index.ts#L33,
+ * except that we're using a linked list-based queue rather than an array-based one.
+ *
+ * @internal
+ */
+export async function* eachValueFrom<T>(source: Observable<T>): AsyncIterableIterator<T> {
+  const deferreds = new Queue<ResolvablePromise<{ value?: T; done: boolean }>>();
```
Yato333, line 15: Is there a point for copying the exact same code? 
`deferred` variable can be renamed to something like `deferredPromises`, `null` can be changed to `undefined`, and all `else`s in `while(true)` can be removed.

grigasp:
> The performance of large flat hierarchy and base class grouping has worsened dramatically. Why would that be?

Turns out that was due to excessive use of `Array.shift` in `eachValueFrom` implementation that we were using. Changed the implementation to use linked list instead - the performance is back on track now.

packages/hierarchies/src/hierarchies/internal/operators/EachValueFrom.ts
grigasp: don't see a point in renaming `deferred` to `deferredPromise`, but agree about the rest - cleaned up

packages/hierarchies/src/hierarchies/internal/operators/EachValueFrom.ts
Yato333, line 1: This isn't an operator. Shouldn't it go to the internal folder?

packages/hierarchies/src/hierarchies/internal/operators/EachValueFrom.ts
Yato333: Sorry, I meant `deferreds` variable which is not a word and was added to the spell checker exceptions

packages/hierarchies/src/hierarchies/internal/operators/EachValueFrom.ts
grigasp: Monday... :D

packages/hierarchies/src/hierarchies/internal/operators/EachValueFrom.ts
grigasp: I think that's valid word: https://www.dictionary.com/browse/deferred

Yato333: It says that deferred is an adjective and here it's used as a noun in plural. I'm not sure if it's valid

grigasp: let's try this, then :) https://en.wiktionary.org/wiki/deferreds#English

Yato333 (approved):

PR #518, opened 2024-04-08
Unified selection: Hilite set streaming

### LINKED ISSUE #501: unified-selection: Streaming hilite sets on interval
https://github.com/iTwin/presentation/pull/488 added support for hilite sets - `HiliteSetProvider.getHiliteSet()` returns an `AsyncIterableIterator<HiliteSet>`, which allows us to stream the results, but the PR didn't implement the streaming part and just emits a single, final hilite set.

Ideally, we'd like to:
1. Have an observable that streams instance keys from given selectables.
2. Change `getHilitedModels`, `getHilitedSubCategories` and `getHilitedElements` to return `Observable<Id64String>` and stream the ids as they arrive from the queries.
3. Accumulate the results of the 3 observables into a `HiliteSet` and emit it on some interval.

Additionally, we may want to consider splitting some queries and executing multiple queries rather than a single union - that may give us better overall performance.

Yato333:
Closes #501


packages/unified-selection/src/test/HiliteSetProvider.test.ts
```diff
@@ -332,6 +339,54 @@ describe("HiliteSetProvider", () => {
...
+
+        mockClass("TestElement", "BisCore", "Element");
+        mockQuery([], [], elementPromises);
+
+        const selectables = Selectables.create(elementKeys);
+        const iter = provider.getHiliteSet({ selectables });
+        elementPromises[0].resolve(elementKeys[0].id);
+
+        const nextValuePromise = iter.next();
+        await Promise.all([expect(waitForPromise(nextValuePromise, 5)).to.eventually.be.undefined, timers.tickAsync(5)]);
```
grigasp, line 368: I'm not a fan of the `waitForPromise` function... Maybe just do this?
```
expect(await Promise.race([
  nextValuePromise,
  timers.tickAsync(5).then(() => "timeout"),
])).to.eq("timeout");
```

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -105,43 +125,72 @@ class HiliteSetProviderImpl implements HiliteSetProvider {
...
+      return res;
+    });
+
+    // Add the promise to cache to prevent `getTypeImpl` being called multiple times.
+    this._classRelationCache.set(key.className, promise);
+    return promise;
+  }
 
+  private async getTypeImpl(key: SelectableInstanceKey): Promise<InstanceIdType> {
+    const keyClass = await this.getClass(key.className);
```
grigasp, line 140: Do you think using `BaseClassChecker` that we have in `hierarchies` package would be helpful here?

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -54,49 +56,67 @@ class HiliteSetProviderImpl implements HiliteSetProvider {
...
-      map.set(keyType, array);
-    }
+    let hiliteSet: HiliteSet = { models: [], subCategories: [], elements: [] };
+    let lastEmitTime = performance.now();
+    const subject = new Subject<HiliteSet>();
+    const subscriptions = (["models", "subCategories", "elements"] as const).map((key) =>
+      observables[key]!.subscribe({
+        next(val) {
+          hiliteSet[key].push(val);
+          if (performance.now() - lastEmitTime < 20) {
```
grigasp, line 93: let's define a `const` for this value

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
Yato333: It probably would be useful, but should `unified-selection` now have `hierarchies` as a dependency? Or do you suggest we create another package?

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
Yato333: For `observables[key]!` or the result of `subscribe`?

grigasp: `20`

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
grigasp: `hierarchies` should definitely not be a dependency for `unified-selection`. We'll need to come up with a way to share such code between packages, but in the meantime I suggest we simply have a hard copy of the thing.

Yato333: `BaseClassChecker` has a little bit different purpose. It caches, whether one class is base of another. Here we don't care about this information, we only want to know class names by instance ID. I think, we should leave it as it is now, or I can extract this logic to a separate class if you think it might be useful

packages/unified-selection/src/test/HiliteSetProvider.test.ts
Yato333: Agree, applied suggested changes

packages/unified-selection/src/test/_helpers/PromiseUtils.ts
```diff
@@ -0,0 +1,58 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * A promise that can be manually resolved.
+ * @internal Used for testing only.
+ */
+export class ResolvablePromise<T> implements Promise<T> {
```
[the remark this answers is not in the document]
Yato333: Moved all instances of classes similar to ResolvablePromise to test-utilities

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
Yato333: Oh, that's right. I guess, it's Monday for me as well today :D
Extracted to constant

packages/unified-selection/src/test/_helpers/PromiseUtils.ts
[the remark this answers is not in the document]
grigasp: Thanks

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
Yato333: Oh yes, my bad, you are right. But still, current implementation caches `instanceIdType` which helps us prevent doing this for each selectable:
```ts
    const checkType = async (idType) => await baseClassChecker.is(selectable, idType) ? idType : undefined);
    return (
      (await checkType("subject")) ??
      (await checkType("model")) ??
      (await checkType("category")) ??
      (await checkType("subCategory")) ??
      (await checkType("functionalElement", "Functional")) ??
      (await checkType("groupInformationElement")) ??
      (await checkType("geometricElement")) ??
      (await checkType("element")) ??
      "unknown"
    );
 ```
 Base class checker doesn't help in this regard

grigasp: okay, I now see that we already have `_classRelationCache` for storing `class name -> end result` map, so I agree there's no need to another, intermediate, cache

grigasp (approved):

PR #520, opened 2024-04-10
Tree node load duration reporting

jasdom:
Changes needed for [801](https://github.com/iTwin/viewer-components-react/issues/801).

Added `ReportingTreeNodeLoader` which wraps an existing node loader and reports how long it took to load the node.


packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
```diff
@@ -198,7 +214,8 @@ function useTreeState(props: TreeStateProps) {
     const { treeModel, ...providerProps } = props;
     const modelSource = new TreeModelSource(new MutableTreeModel(treeModel));
     const dataProvider = new PresentationTreeDataProvider(providerProps);
-    const nodeLoader = new PagedTreeNodeLoader(dataProvider, modelSource, providerProps.pagingSize);
+    const pagedLoader = new PagedTreeNodeLoader(dataProvider, modelSource, providerProps.pagingSize);
+    const nodeLoader = props.onNodeLoaded ? new ReportingTreeNodeLoader(pagedLoader, props.onNodeLoaded) : pagedLoader;
```
saskliutas, line 218: New `nodeLoader` is also created in `onReload` callback.

packages/components/src/test/tree/ReportingTreeNodeLoader.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      });
+
+      await waitFor(() => {
+        expect(loadedNodes).to.not.be.undefined;
+        expect(loadedNodes).to.have.lengthOf(3);
+        expect(reportStub).to.be.calledWithMatch({ node: "id", duration: 300 });
+      });
+    });
+
+    it("only reports the first page", async () => {
```
saskliutas, line 58: Can you add test to verify that calling `loadNode` for first page multiple times logs only one message?

packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
```diff
@@ -108,6 +108,15 @@ describe("usePresentationTreeState", () => {
     await waitFor(() => expect(result.current?.nodeLoader).to.not.eq(oldNodeLoader));
   });
 
+  it("creates new reporting nodeLoader when `onNodeLoaded` callback passed", async () => {
```
saskliutas, line 111: Does not look like this test actually verifies that reporting nodeLoader is created. Looking at previous test it I would  assume that in this test new provider was created because of `pageSize` and `onNodeLoader` probably didn't have any effect.

saskliutas (changes requested):

packages/components/src/test/tree/ReportingTreeNodeLoader.test.ts
jasdom: Adjusted the implementation to not track the same request multiple times and to not report if unsubscribed before completion. Added tests for these cases too.

packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
```diff
@@ -46,6 +46,7 @@ describe("usePresentationTreeState", () => {
     imodel,
     ruleset: rulesetId,
     pagingSize: 5,
+    onNodeLoaded: undefined,
```
saskliutas, line 49: We is this necessary?

packages/components/src/test/tree/ReportingTreeNodeLoader.test.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  describe("loadNode", () => {
+    it("reports node loading duration", async () => {
+      const performanceStub = sinon.stub(performance, "now");
+      performanceStub.onCall(0).returns(0);
+      performanceStub.onCall(1).returns(300);
+
+      const observable = reportingNodeLoader.loadNode({ id: "id" } as TreeModelNode, 0);
+      let loadedNodes: TreeNodeItem[] = [];
```
saskliutas, line 49: I think there is no need to track loaded nodes. Also maybe it would be simpler to use `Subject` instead of `of` to mock `loadNode`. Eg:

```ts
const lodeNodeSubject = new Subject<TreeNodeLoadResult>();
nodeLoaderStub.loadNode.returns(loadNodeSubject);

reportingNodeLoader.loadNode({ id: "id" } as TreeModelNode, 0).subscribe();
nodeLoadSubject.complete();

await waitFor(() => {
  expect(reportStub).to.be.calledWithMatch({ node: "id", duration: 300 });
});
```

Same in other tests. `Subject` will allow you to manually control when `loadNode` observable completes/errors.

packages/components/src/test/tree/ReportingTreeNodeLoader.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      });
+
+      await waitFor(() => {
+        expect(loadedNodes).to.not.be.undefined;
+        expect(loadedNodes).to.have.lengthOf(6);
+        expect(reportStub).to.be.calledOnce;
+      });
+    });
+
+    it("does not report if no longer subscribed", async () => {
```
saskliutas, line 97: I guess we also need to test that time is not logged if `loadNode` observable errors?

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
```diff
@@ -212,13 +227,17 @@ function useTreeState(props: TreeStateProps) {
...
-    const { modelSource, dataProvider } = reloadedTree;
-    const nodeLoader = new PagedTreeNodeLoader(dataProvider, modelSource, dataProvider.pagingSize!);
-    setState({ dataProvider, nodeLoader });
-  }, []);
+      const { modelSource, dataProvider } = reloadedTree;
+      const pagedLoader = new PagedTreeNodeLoader(dataProvider, modelSource, dataProvider.pagingSize!);
+      const nodeLoader = props.onNodeLoaded ? new ReportingTreeNodeLoader(pagedLoader, props.onNodeLoaded) : pagedLoader;
+      setState({ dataProvider, nodeLoader });
+    },
+    [props.onNodeLoaded],
```
saskliutas, line 239: Can you add `onNodeLoaded` function into `useRef` so that `usePresentationTreeState` confumers don't need to memoize this function?

packages/components/src/presentation-components/tree/ReportingTreeNodeLoader.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (childIndex !== 0 || this._trackedRequests.has(parentId)) {
+      return observable;
+    }
+
+    let time: number;
+    this._trackedRequests.add(parentId);
+    const tracked = toRxjsObservable(observable).pipe(
+      tap({
+        subscribe: () => (time = performance.now()),
+        unsubscribe: () => this._trackedRequests.delete(parentId),
```
saskliutas, line 43: I think `this._trackedRequests.delete(parentId)` should be called on `error` too.

saskliutas (changes requested):

packages/components/src/test/tree/ReportingTreeNodeLoader.test.ts
jasdom: I am aware that tracking the loaded nodes is not necessary for testing whether the load duration is reported, however we should probably also test to see if the `ReportingTreeNodeLoader` returns the same result as the wrapped loader?

saskliutas: Ok we can leave nodes tracking but I still think using `Subject` simplifies tests.

packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
jasdom: I am not sure why but it was necessary, however with the latest changes it was no longer needed.

packages/components/src/test/tree/ReportingTreeNodeLoader.test.ts
jasdom: Switched to `Subject`.

packages/components/src/test/tree/ReportingTreeNodeLoader.test.ts
jasdom: Added extra tests.

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
jasdom: Added into `useRef`.

packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
```diff
@@ -61,12 +62,16 @@ describe("usePresentationTreeState", () => {
...
 
+    sinon.stub(IModelApp, "initialized").get(() => true);
     sinon.stub(Presentation, "presentation").get(() => presentationManager);
+    sinon.stub(Presentation, "localization").get(() => localization);
     sinon.stub(IModelApp, "quantityFormatter").get(() => ({
       onActiveFormattingUnitSystemChanged,
     }));
-    await UiComponents.initialize(new EmptyLocalization());
+    await Presentation.initialize();
+    await UiComponents.initialize(localization);
```
saskliutas, line 74: Can you replace these with `sinon.stub`?

saskliutas (changes requested):

packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
jasdom: Replaced `Presentation.initialize()` with a stub.

jasdom: Okay, I have no idea what this was about. I added it yesterday because it was complaining `Presentation` was not initialized when calling `loadNode`, but now it seems to work fine for whatever reason 😃

.changeset/quiet-rocks-enjoy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
```
saskliutas, line 2:
```suggestion
"@itwin/presentation-components": minor
```

saskliutas (approved):

PR #521, opened 2024-04-10
Introduce `@itwin/presentation-shared` package for shared code

grigasp:


packages/hierarchies/src/hierarchies/internal/LoggingUtils.ts
Yato333, line 1: Can we also extract the logic for checking the log category from here?
```ts
export function log(logger: ILogger, props: LogMessageProps) {
  const severity = props.severity ?? "trace";
  if (!logger.isEnabled(props.category, severity)) {
    return;
  }
  const logFunc = getLogFunc(logger, severity);
  logFunc(props.category, props.message());
}
```

apps/test-app/frontend/src/components/app/App.tsx
```diff
@@ -386,3 +385,8 @@ function createInstanceKeysSelectQuery(keys: InstanceKey[]): PresentationQuery {
   });
   return { query, bindings };
 }
+
+function parseFullClassName(fullClassName: string) {
```
Yato333, line 389: Can't you use `parseFullClassName` from the shared package here?

Yato333 (commented):
You can also move ResolvablePromise from test-utils to shared and reuse it in hierarchies/internal/EachValueFrom

packages/shared/src/test/ecsql-snippets/ECSqlJoinSnippets.test.ts
```diff
@@ -5,21 +5,23 @@
...
-import { createRelationshipPathJoinClause } from "../../../hierarchies/queries/ecsql-snippets/ECSqlJoinSnippets";
-import { trimWhitespace } from "../../../hierarchies/Utils";
-import { ClassStubs, createClassStubs } from "../../Utils";
+import * as sinon from "sinon";
+import { createRelationshipPathJoinClause } from "../../shared/ecsql-snippets/ECSqlJoinSnippets";
+import * as m from "../../shared/Metadata";
+import { trimWhitespace } from "../../shared/Utils";
+import { createMetadataProviderStub } from "../MetadataProviderStub";
 
+import EC = m.EC;
```
saskliutas, line 14: Why is this needed? Why not `import { EC } from "../../shared/Metadata";`?

packages/hierarchies/src/hierarchies/internal/LoggingUtils.ts
grigasp: I don't see any virtue in doing that. Can you explain why you think it would help?

apps/test-app/frontend/src/components/app/App.tsx
grigasp: I could, but I'm about to remove all this selection-related code from test app, including the usage of this function, so didn't want to introduce the `@itwin/presentation-shared` dependency unnecessarily.

grigasp:
> You can also move ResolvablePromise from test-utils to shared and reuse it in hierarchies/internal/EachValueFrom

https://github.com/iTwin/presentation/pull/522

packages/shared/src/test/ecsql-snippets/ECSqlJoinSnippets.test.ts
grigasp: Was used as a workaround for something that is now irrelevant. Cleaned up.

saskliutas (approved):

PR #522, opened 2024-04-10
hierarchies: Remove the need of `ResolvablePromise` in our `eachValueFrom` implementation

grigasp:


packages/hierarchies/src/hierarchies/internal/EachValueFrom.ts
```diff
@@ -6,13 +6,16 @@
 import { Observable } from "rxjs";
 
 /**
- * This is pretty much a copy of https://github.com/benlesh/rxjs-for-await/blob/94f9cf9cb015ac3700dfd1850eb81d36962eb70f/src/index.ts#L33,
- * except that we're using a linked list-based queue rather than an array-based one.
+ * This is pretty much a combination of:
+ * - https://github.com/benlesh/rxjs-for-await/blob/94f9cf9cb015ac3700dfd1850eb81d36962eb70f/src/index.ts#L33,
+ * - https://github.com/ReactiveX/rxjs/blob/2587ee852eb43eeb0883a7787bac2944d823392b/packages/observable/src/observable.ts#L922.
+ *
+ * Out implementation uses a linked list rather than an array to store values, which is more efficient.
```
saskliutas, line 13:
```suggestion
 * Our implementation uses a linked list rather than an array to store values, which is more efficient.
```

saskliutas (approved):

packages/hierarchies/src/hierarchies/internal/EachValueFrom.ts
Yato333, line 13:
```suggestion
 * Our implementation uses a linked list rather than an array to store values, which is more efficient.
```

Yato333 (approved):

PR #524, opened 2024-04-12
Bump `@itwin/eslint-plugin` and `typescript`

saskliutas:


packages/shared/src/shared/Formatting.ts
```diff
@@ -37,6 +37,7 @@ export function createDefaultValueFormatter(): IPrimitiveValueFormatter {
         return result;
       }
     }
+    // eslint-disable-next-line @typescript-eslint/no-base-to-string
```
grigasp, line 40: would this help avoid ignoring the rule?

```suggestion
    assert(typeof x === "undefined" || typeof x === "boolean" || typeof x === "number" || typeof x === "string");
```

grigasp (approved):

packages/shared/src/shared/Formatting.ts
saskliutas: Yes, added assert

PR #527, opened 2024-04-12
Update `itwinjs-core` package dependencies to `4.5.1`

grigasp:
Lots of stuff...
- Fix tests that were expecting selection manager to return changed selection immediately after the change.
- Stop using deprecated nodes', content, distinct values apis in favor of the new "iterator" ones (while keeping in mind that we didn't change the peer dependency version, so the new API may not exist).
- Stop using `FavoritePropertiesManager` `has` and `sortFields` functions in favor of the new `async` versions (while keeping in mind that we didn't change the peer dependency version, so the new API may not exist).

Also, not directly related to the version bump:
- ~Use `source-map-support` in `presentation-components`.~
- ~Tell `source-map-support` that it's running in `node` environment to avoid XHR warnings in debug console. This also seems to improve tests startup times.~
- More test utility sharing through `presentation-test-utilities`.


grigasp:
The `PresentationPropertyDataProvider` changes to add `Promise` return types to `isFieldFavorite` and `sortFields` are breaking... Going to reconsider how to handle that without breaking changes.

packages/components/.mocharc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "require": ["ignore-styles"],
+  "require": ["ignore-styles", "source-map-support/register"],
```
saskliutas, line 2: Is `source-map-support/register` is not needed when we call `sourceMapSuppor.install()` in setup file.

grigasp: I'm going to remove this altogether when https://github.com/iTwin/presentation/pull/528 merges.

Issue #103, opened 2023-03-17
Provide a React hook for setting up unified selection viewport


grigasp:
We currently support 4 unified selection components and all of them use hooks except the `ViewportComponent`, which still uses an outdated HOC API. That also forces us to have `imodel-components-react` dependency which is otherwise unnecessary. We should deprecate the HOC and add a proper hook for that.

We should consider taking `IModelConnection` as a prop rather than a `Viewport` or viewport component.

**Note:** we should think about how to make hooking viewports (iModels) into unified selection easy, without forcing consumers that use AppUi to define a unified-selection-friendly `IModelViewportControl` like [here](https://github.com/iTwin/viewer/blob/f12937cef8bc0b104d6f07c18aaa5545a5a44fc4/packages/modules/viewer-react/src/components/app-ui/providers/UnifiedSelectionViewportControl.tsx).

PR #530, opened 2024-04-16
Viewport and unified selection synchronization

jasdom:
Closes [103](https://github.com/iTwin/presentation/issues/103).


packages/unified-selection/src/test/_helpers/Utils.ts
```diff
@@ -0,0 +1,24 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { BeDuration, StopWatch } from "@itwin/core-bentley";
+
+export async function waitFor<T>(check: () => Promise<T> | T, timeout?: number): Promise<T> {
```
grigasp, line 8: use https://github.com/iTwin/presentation/blob/1a0cd35be662e7ccead170a2492e10d35829b371/packages/test-utilities/src/test-utilities/WaitFor.ts#L8 instead

packages/unified-selection/README.md
```diff
@@ -166,3 +166,21 @@ const queryExecutor = createECSqlQueryExecutor(imodel);
 // Returns the parent element, or the element itself if it does not have a parent, for each element specified in `elementIds` argument.
 const selection = computeSelection({ queryExecutor, elementIds, scope: { id: "element", ancestorLevel: 1 } });
 ```
+
+## Viewport synchronization with unified selection
```
grigasp, line 170: I'm not sure it's correct to say that we're syncing **viewport** with unified selection, when we're actually syncing **iModel's selection set** with unified selection. 

Also, I believe saying that this hook should be used in a viewport component is not correct. What's going to happen if there are two or more viewports for the same iModel?

jasdom: We are also syncing iModel's hilited set, so I was not sure what a good title for it would be. Should the `syncViewportWithUnifiedSelection` function be renamed as well?

Regarding the multiple viewports, in the previous implementation `SelectionManager` had the methods `setSyncWithIModelToolSelection` and `suspendIModelToolSelectionSync` which were responsible for registering the listeners and ensuring that only one was active at a time. I did not add an equivalent for it as I am not sure what is the use case for attempting to synchronize from multiple places at the same time. In the current implementation it would be problematic if the function is called more than once, regardless whether it is in a viewport component or not. Would it be better to then mention that it should only be called once?

grigasp: > We are also syncing iModel's hilited set, so I was not sure what a good title for it would be. Should the `syncViewportWithUnifiedSelection` function be renamed as well?

Yes, the function should not lie about what it does. Also, what I don't like about `syncViewportWithUnifiedSelection` is that it sounds like a one time action, when in fact it's enabling a continuous sync. I think the name could be, for example `enableUnifiedSelectionSyncWithIModel`. It's not great, so I welcome suggestions for better names.

> In the current implementation it would be problematic if the function is called more than once, regardless whether it is in a viewport component or not. Would it be better to then mention that it should only be called once?

I think the function being clear that it's syncing with an IModel rather than a Viewport is a good first step. The next step is API docs and README making it clear what it does and providing a good example.

packages/unified-selection/src/test/_helpers/Utils.ts
jasdom: Switched to using `presentation-test-utilities`.

packages/unified-selection/README.md
jasdom: I think `enableUnifiedSelectionSyncWithIModel` is good enough. Renamed the functions to better reflect what they are doing. Updated docs and README.

.changeset/gold-pears-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/unified-selection": minor
+---
+
+Added API for enabling synchronization between iModel selection and unified selection.
```
grigasp, line 5:
```suggestion
Added `enableUnifiedSelectionSyncWithIModel` for enabling synchronization between iModel's tool selection and unified selection storage.
```

packages/unified-selection/README.md
```diff
@@ -166,3 +166,21 @@ const queryExecutor = createECSqlQueryExecutor(imodel);
 // Returns the parent element, or the element itself if it does not have a parent, for each element specified in `elementIds` argument.
 const selection = computeSelection({ queryExecutor, elementIds, scope: { id: "element", ancestorLevel: 1 } });
 ```
+
+## iModel selection synchronization with unified selection
+
+The `@itwin/unified-selection` package delivers a `enableUnifiedSelectionSyncWithIModel` function to enable selection synchronization between an iModel and a `SelectionStorage`. When called it returns a cleanup function that should be used to disable the synchronization. There should only be one active synchronization between a single iModel and a `SelectionStorage` at a given time. For example, this function could be used inside a `useEffect` hook in a viewport component but should not be used if there is more than one viewport for the same iModel (in such cases the function could be called in a higher scope so that all viewport components would share the synchronization):
```
grigasp, line 172:
```suggestion
The `@itwin/unified-selection` package delivers a `enableUnifiedSelectionSyncWithIModel` function to enable selection synchronization between an iModel and a `SelectionStorage`. When called, it returns a cleanup function that should be used to disable the synchronization. There should only be one active synchronization between a single iModel and a `SelectionStorage` at a given time. For example, this function could be used inside a `useEffect` hook in a viewport component but should not be used if there is more than one viewport for the same iModel (in such cases the function could be called in a higher scope so that all viewport components would share the synchronization):
```

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
```diff
@@ -0,0 +1,46 @@
...
+  /** SelectionStorage to synchronize selection with. */
+  selectionStorage: SelectionStorage;
+  /** Caching hilite set provider. */
+  cachingHiliteSetProvider: CachingHiliteSetProvider;
+  /** iModel ECSql query executor. */
+  queryExecutor: IECSqlQueryExecutor;
+  /** iModel metadata provider. */
+  metadataProvider: IMetadataProvider;
+  /** Active scope provider. */
+  activeScopeProvider: () => ElementSelectionScopeProps | { id: SelectionScope } | SelectionScope;
```
grigasp, line 34:
```suggestion
  activeScopeProvider: () => ComputeSelectionProps["scope"];
```

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+import { IdArg, IModelConnection, SelectionSetEvent, SelectionSetEventType } from "./IModel";
+
+/** @internal */
+export interface IModelSelectionHandlerProps {
+  iModel: IModelConnection;
+  selectionStorage: SelectionStorage;
+  cachingHiliteSetProvider: CachingHiliteSetProvider;
+  queryExecutor: IECSqlQueryExecutor;
+  metadataProvider: IMetadataProvider;
+  activeScopeProvider: () => ElementSelectionScopeProps | { id: SelectionScope } | SelectionScope;
```
grigasp, line 29:
```suggestion
  activeScopeProvider: () => ComputeSelectionProps["scope"];
```

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ElementSelectionScopeProps, SelectionScope } from "../SelectionScope";
+import { SelectionStorage } from "../SelectionStorage";
+import { IModelConnection } from "./IModel";
+import { IModelSelectionHandler } from "./IModelSelectionHandler";
+
+/**
+ * Props for `enableUnifiedSelectionSyncWithIModel`.
+ * @internal Not exported through barrel, but used in public API as an argument. May be supplemented with optional attributes any time.
+ */
+export interface EnableUnifiedSelectionSyncWithIModelProps {
```
grigasp, line 22: does this need to be exported at all?

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
```diff
@@ -0,0 +1,46 @@
...
+  /** iModel to synchronize selection for. */
+  iModel: IModelConnection;
+  /** SelectionStorage to synchronize selection with. */
+  selectionStorage: SelectionStorage;
```
grigasp, line 26:
```suggestion
  /** Selection storage to synchronize IModel's tool selection with. */
  selectionStorage: SelectionStorage;
```

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
```diff
@@ -0,0 +1,46 @@
...
+  /** Caching hilite set provider. */
+  cachingHiliteSetProvider: CachingHiliteSetProvider;
```
grigasp, line 28:
```suggestion
  /** A caching hilite set provider used to retrieve hilite sets for an iModel. */
  cachingHiliteSetProvider: CachingHiliteSetProvider;
```

grigasp, line 28: I think often `enableUnifiedSelectionSyncWithIModel` is going to be the only user of `CachingHiliteSetProvider`. Do you think we could make providing it optional? I understand creating a `CachingHiliteSetProvider` requires a function to get `queryExecutor` and `IMetadataProvider` based on imodel key, which we can't provide. But we have the required `queryExecutor` and `IMetadataProvider`, so maybe we can come up with something?

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+import { IMetadataProvider } from "../queries/ECMetadata";
+import { IECSqlQueryExecutor } from "../queries/ECSqlCore";
+import { SelectableInstanceKey, Selectables } from "../Selectable";
+import { StorageSelectionChangeEventArgs, StorageSelectionChangeType } from "../SelectionChangeEvent";
+import { computeSelection, ElementSelectionScopeProps, SelectionScope } from "../SelectionScope";
+import { SelectionStorage } from "../SelectionStorage";
+import { IdArg, IModelConnection, SelectionSetEvent, SelectionSetEventType } from "./IModel";
+
+/** @internal */
+export interface IModelSelectionHandlerProps {
```
grigasp, line 23: Why not reuse `EnableUnifiedSelectionSyncWithIModelProps`?

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
```diff
@@ -0,0 +1,46 @@
...
+import { IModelConnection } from "./IModel";
+import { IModelSelectionHandler } from "./IModelSelectionHandler";
+
+/**
+ * Props for `enableUnifiedSelectionSyncWithIModel`.
+ * @internal Not exported through barrel, but used in public API as an argument. May be supplemented with optional attributes any time.
+ */
+export interface EnableUnifiedSelectionSyncWithIModelProps {
+  /** iModel to synchronize selection for. */
+  iModel: IModelConnection;
```
grigasp, line 24: It seems really weird that we require `IModelConnection` and, separately - IModel's query executor and metadata provider. I think `IModelConnection` needs to be renamed to reflect it's purpose more closely and not confuse consumers into thinking that this is a real `IModelConnection`.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (args.iModelKey !== this._iModel.key || args.level !== 0) {
+      return;
+    }
+
+    this._cancelOngoingChanges.next();
+    this.handleUnifiedSelectionChange(args.changeType, args.selectables);
+  };
+
+  private applyCurrentHiliteSet(clearBefore = true) {
+    if (clearBefore) {
```
grigasp, line 124: It's unclear what the argument means in the call sites - the following should help:

```suggestion
  private applyCurrentHiliteSet({ activeSelectionAction }: { activeSelectionAction: "clear" | "keep" }) {
    if (activeSelectionAction === "clear") {
```

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+    from(this._cachingHiliteSetProvider.getHiliteSet({ iModelKey: this._iModel.key }))
+      .pipe(takeUntil(this._cancelOngoingChanges))
+      .subscribe({
+        next: (ids) => {
+          this.applyHiliteSet(ids);
+        },
+      });
+  }
+
+  private applyHiliteSet(set: HiliteSet) {
```
grigasp, line 140: Maybe rename this to `addHiliteSet`? "Apply" sounds like it's replacing whatever's set to the provided hilite set, while in fact it's just adding new items to already existing sets.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    const elementIds = this.getSelectionSetChangeIds(event);
+    const scopedSelection = computeSelection({ queryExecutor: this._queryExecutor, elementIds, scope: this._activeScopeProvider() });
+    await this.handleIModelSelectionChange(event.type, scopedSelection);
+  };
+
+  private async handleIModelSelectionChange(type: SelectionSetEventType, iterator: AsyncIterableIterator<SelectableInstanceKey>) {
+    if (type === SelectionSetEventType.Replace) {
+      this._selectionStorage.clearSelection({ iModelKey: this._iModel.key, source: this._selectionSourceName });
+    }
```
grigasp, line 188: Can we move this under handling of `Remove` to get it closed to the other part of `Replace` handling?

packages/unified-selection/src/unified-selection/iModel/IModel.ts
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Payload sent to [[SelectionSet.onChanged]] event listeners to describe how the contents of the set have changed.
+ */
+export type SelectionSetEvent = SelectAddEvent | SelectRemoveEvent | SelectReplaceEvent;
+
+/**
+ * Manages a set of *listeners* for a particular event and notifies them when the event is raised.
+ */
+export interface Event<T> {
```
grigasp, line 75: I think we should move existing `Event` interface in `core-interop` to `presentation-shared` and reuse it from there.

packages/unified-selection/src/unified-selection/iModel/IModel.ts
```diff
@@ -0,0 +1,136 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+/** @packageDocumentation
+ * @module UnifiedSelection
+ */
+
+/**
+ * Used as an argument to a function that can accept one or more [[Id64String]]s.
+ */
+export type IdArg = string | Set<string> | string[];
```
grigasp, line 13: I don't think we need it here, we can import `Id64Arg` from `core-bentley`

packages/unified-selection/src/unified-selection/iModel/IModel.ts
```diff
@@ -0,0 +1,136 @@
...
+  /** The set of hilited models. */
+  models: Uint32Set;
+  /** Remove all elements from the hilited set. */
+  clear(): void;
+}
+
+/**
+ * Interface representing a connection to iModel.
+ */
+export interface IModelConnection {
```
grigasp, line 129: I think this is the only type in this file that needs to be exported.

grigasp (commented):
Please integrate this into our test app.

.changeset/gold-pears-add.md
jasdom: Changed.

packages/unified-selection/README.md
jasdom: Fixed.

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
jasdom: Was not aware you can do that, changed.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
jasdom: Changed.

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
jasdom: Changed.

jasdom: Made it optional by using a function that returns the passed `queryExecutor` and `IMetadataProvider` when creating `CachingHiliteSetProvider`.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
jasdom: Reused.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
jasdom: Changed.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
jasdom: I agree, especially considering that we have a function called `removeHiliteSet`.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
jasdom: Moved it closer to the add part.

packages/unified-selection/src/unified-selection/iModel/IModel.ts
jasdom: Moved the interface into `presentation-shared` and reused it.

packages/unified-selection/src/unified-selection/iModel/IModel.ts
jasdom: Switched to using `Id64Arg`.

packages/unified-selection/src/unified-selection/iModel/IModel.ts
jasdom: `SelectionSetEvent` was also needed. Changed to no longer export the other types.

apps/test-app/frontend/src/components/viewport/ViewportContentControl.tsx
```diff
@@ -27,11 +28,21 @@ export default function ViewportContentComponent(props: ViewportContentComponent
     setPrevIModel(props.imodel);
   }
   useEffect(() => {
+    const schemas = new SchemaContext();
+    schemas.addLocater(new ECSchemaRpcLocater(props.imodel.getRpcProps()));
```
grigasp, line 32: I think this should be moved to `MyAppFrontend` and shared across components (e.g. with stateless tree)

packages/unified-selection/README.md
```diff
@@ -169,18 +169,18 @@ const selection = computeSelection({ queryExecutor, elementIds, scope: { id: "el
...
 useEffect(() => {
   return enableUnifiedSelectionSyncWithIModel({
-    iModel,
+    iModelSelection: iModel,
     selectionStorage,
-    cachingHiliteSetProvider,
     queryExecutor: createECSqlQueryExecutor(iModel),
     metadataProvider: createMetadataProvider(iModel),
     activeScopeProvider: () => "element",
+    cachingHiliteSetProvider, // optional.
```
grigasp, line 183: I'd say there's no need to specify it in the example

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
grigasp: Can't say I love the name either. After discussing the topic with @saskliutas, we came up with a plan to refactor the way we access imodels in our new packages. So let's leave this as it is now and refactor that afterwards.

grigasp (commented):
Need to make sure we don't publish until the `shared` package is also ready to be published.

packages/unified-selection/README.md
jasdom: Removed.

apps/test-app/frontend/src/components/viewport/ViewportContentControl.tsx
jasdom: Turns out `MyAppFrontend` already has a function for it, so switched to using it here and in stateless tree.

packages/unified-selection/src/unified-selection/iModel/EnableUnifiedSelectionSyncWithIModel.ts
```diff
@@ -10,28 +10,28 @@
...
-  /** Caching hilite set provider. */
-  cachingHiliteSetProvider: CachingHiliteSetProvider;
   /** iModel ECSql query executor. */
   queryExecutor: IECSqlQueryExecutor;
   /** iModel metadata provider. */
   metadataProvider: IMetadataProvider;
   /** Active scope provider. */
-  activeScopeProvider: () => ElementSelectionScopeProps | { id: SelectionScope } | SelectionScope;
+  activeScopeProvider: () => ComputeSelectionProps["scope"];
+  /** A caching hilite set provider used to retrieve hilite sets for an iModel. */
```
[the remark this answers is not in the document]
jasdom: Added an explanation.

packages/unified-selection/src/unified-selection/iModel/IModel.ts
```diff
@@ -26,91 +24,93 @@ export enum SelectionSetEventType {
...
-}
-
 /**
  * A set of *currently selected* elements for an IModelConnection.
  */
-export interface SelectionSet {
+interface SelectionSet {
   /** iModel of the set. */
-  iModel: IModelConnection;
+  iModel: IModelSelection;
```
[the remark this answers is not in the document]
jasdom: It was used for checking whether the selection change event was of the correct iModel, but I removed it since the event iModel should always be the same.

packages/unified-selection/src/unified-selection/iModel/IModel.ts
```diff
@@ -26,91 +24,93 @@ export enum SelectionSetEventType {
...
   /** Event called whenever elements are added or removed from this SelectionSet. */
-  onChanged: Event<(ev: SelectionSetEvent) => void>;
+  onChanged: Event<(ev?: SelectionSetEvent) => void>;
```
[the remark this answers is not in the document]
jasdom: Without it the listener does not satisfy the constraint `() => void`, so as a workaround I made it optional.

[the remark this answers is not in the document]
grigasp: We shouldn't workaround issues in our own code, especially pieces of it that we can refactor, if necessary. Let's instead improve the `Event` interface to allow creating events with required arguments.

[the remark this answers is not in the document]
jasdom: Updated the `Event` interface to support required arguments.

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
```diff
@@ -0,0 +1,767 @@
...
+      deleteIds: sinon.stub<[Id64Arg], void>(),
+    },
+    subcategories: {
+      addIds: sinon.stub<[Id64Arg], void>(),
+      deleteIds: sinon.stub<[Id64Arg], void>(),
+    },
+  };
+
+  const selectionSet = {
+    iModel: {} as IModelSelection,
```
grigasp, line 56: I believe `iModel` was removed from `SelectionSet`

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
```diff
@@ -0,0 +1,767 @@
...
+        cachingHiliteSetProviderFactory.getCall(0).args[0].iModelProvider;
+
+      const result = iModelProvider("iModel");
+      expect(result.queryExecutor).to.be.eq(queryExecutor);
+      expect(result.metadataProvider).to.be.eq(metadataProvider);
+    });
+
+    it("clears selection", async () => {
+      const addedKeys = [createSelectableInstanceKey(1), createSelectableInstanceKey(2)];
+      queryExecutor.createQueryReader.returns(createFakeQueryReader(toQueryResponse(addedKeys)));
```
grigasp, line 208: Is this needed for the "clear" test?

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
```diff
@@ -0,0 +1,767 @@
...
+      }
+      cachingProvider.getHiliteSet.callsFake(() => generator());
+
+      triggerSelectionChange();
+
+      await waitFor(() => {
+        expect(hilited.clear).to.be.calledOnce;
+        expect(hilited.models.addIds).to.be.calledOnceWith([id]);
+        expect(selectionSet.emptyAll).to.be.calledOnce;
+        expect(selectionSet.add).to.not.be.calledOnce;
```
grigasp, line 429:
```suggestion
        expect(selectionSet.add).to.not.be.called;
```

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
```diff
@@ -0,0 +1,767 @@
...
+      }
+      cachingProvider.getHiliteSet.callsFake(() => generator());
+
+      triggerSelectionChange();
+
+      await waitFor(() => {
+        expect(hilited.clear).to.be.calledOnce;
+        expect(hilited.subcategories.addIds).to.be.calledOnceWith([id]);
+        expect(selectionSet.emptyAll).to.be.calledOnce;
+        expect(selectionSet.add).to.not.be.calledOnce;
```
grigasp, line 448:
```suggestion
        expect(selectionSet.add).to.not.be.called;
```

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
```diff
@@ -0,0 +1,767 @@
...
+
+      await waitFor(() => {
+        expect(hilited.clear).to.not.be.called;
+        expect(hilited.elements.deleteIds).to.be.calledOnceWith([instanceKey.id]);
+        expect(selectionSet.emptyAll).to.not.be.called;
+        expect(selectionSet.remove).to.be.calledOnceWith([instanceKey.id]);
+      });
+    });
+
+    it("removes and reads elements to hilite after remove event", async () => {
```
grigasp, line 563:
```suggestion
    it("removes and re-adds elements to hilite after remove event", async () => {
```

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
```diff
@@ -0,0 +1,767 @@
...
+        expect(hilited.clear).to.not.be.called;
+        expect(hilited.elements.deleteIds).to.be.calledOnceWith(removedIds);
+        expect(hilited.elements.addIds).to.be.calledAfter(hilited.elements.deleteIds).and.calledOnceWith([hilitedId]);
+        expect(selectionSet.emptyAll).to.not.be.called;
+        expect(selectionSet.remove).to.be.calledOnceWith(removedIds);
+        expect(selectionSet.add).to.be.calledAfter(selectionSet.remove).and.calledOnceWith([hilitedId]);
+      });
+    });
+
+    it("removes and reads models to hilite after remove event", async () => {
```
grigasp, line 596:
```suggestion
    it("removes and re-adds models to hilite after remove event", async () => {
```

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
```diff
@@ -0,0 +1,767 @@
...
+        expect(hilited.clear).to.not.be.called;
+        expect(hilited.models.deleteIds).to.be.calledOnceWith(removedIds);
+        expect(hilited.models.addIds).to.be.calledAfter(hilited.models.deleteIds).and.calledOnceWith([hilitedId]);
+        expect(selectionSet.emptyAll).to.not.be.called;
+        expect(selectionSet.remove).to.not.be.called;
+        expect(selectionSet.add).to.not.be.called;
+      });
+    });
+
+    it("removes and reads subcategories to hilite after remove event", async () => {
```
grigasp, line 631:
```suggestion
    it("removes and re-adds subcategories to hilite after remove event", async () => {
```

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
```diff
@@ -0,0 +1,222 @@
...
+      });
+  }
+
+  private onUnifiedSelectionChanged = (args: StorageSelectionChangeEventArgs) => {
+    // iModels are only interested in top-level selection changes
+    if (args.iModelKey !== this._iModelSelection.key || args.level !== 0) {
+      return;
+    }
+
+    this._cancelOngoingChanges.next();
```
grigasp, line 120: Should we always cancel on unified selection change? Imagine the following situation:
1. User selects `X`, we start processing until "cancel".
2. User ctrl+clicks on `Y`, expecting it to be **added** to selection. However, this cancels handling of step 1.
3. Finally, only `Y` is hilited.

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
jasdom: Right, removed.

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
jasdom: Changed.

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
jasdom: Changed.

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
jasdom: Changed.

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
jasdom: Changed.

packages/unified-selection/src/test/IModel/IModelSelectionHandler.test.ts
jasdom: Changed.

packages/unified-selection/src/unified-selection/iModel/IModelSelectionHandler.ts
jasdom: Changed to not cancel when adding or removing.

grigasp (approved):

saskliutas:
I think you will need to include these changes in the new implementation too: https://github.com/iTwin/presentation/pull/538

Issue #441, opened 2024-02-28
Investigate possibility to make `UniqueValuesSelector` searchable.


saskliutas:

grigasp:
Looked into filtering distinct values on the backend and I think there's no big difference on whether we do that there, or on the frontend. 

We get distinct values on the backend by requesting raw values, formatting them, and grouping by formatted value. That means we can't use user's input as a query parameter, because formatted value may be completely different from the raw value that's stored in the iModel. So we would have to match against user's input **after** the formatting step, at which point there's not that much difference compared to filtering on the frontend. The downside is that response has to include values that we're going to drop (downloading them unnecessarily), but the upside is that it doesn't change the RPC / public API, so we don't have to bump peer dependencies or apply any hacks in our components.

Based on the above, I'd say we should filter the values in the component.

PR #531, opened 2024-04-17
Searchable UniqueValuesSelector

jasdom:
Closes [441](https://github.com/iTwin/presentation/issues/441).


packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -251,27 +247,40 @@ function useUniquePropertyValuesLoader({ imodel, ruleset, field, descriptorInput
...
+        options.push(uniqueValue);
+
+        if (matchesSearchInput(uniqueValue)) {
+          filteredOptions.push(uniqueValue);
         }
       }
 
+      setLoadedCount((prev) => prev + items.length);
+      setLoadedOptions([...loadedOptions, ...options]);
+      setHasMore(options.length === UNIQUE_PROPERTY_VALUES_BATCH_SIZE);
```
saskliutas, line 276: I think this should be merged into single `useState` as they are always updated together

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -221,27 +198,46 @@ function getFieldClassInfos(field?: Field) {
...
+  property: PropertyDescription;
+  descriptor: Descriptor;
   ruleset?: Ruleset;
   field?: Field;
   descriptorInputKeys?: Keys;
 }
 
-function useUniquePropertyValuesLoader({ imodel, ruleset, field, descriptorInputKeys }: UseUniquePropertyValuesLoaderProps) {
+function useUniquePropertyValuesLoader({ imodel, property, descriptor, ruleset, field, descriptorInputKeys }: UseUniquePropertyValuesLoaderProps) {
+  const [loadedCount, setLoadedCount] = useState(0);
```
saskliutas, line 209: Do we need `loadedCount`? Isn't `loadedOptions` enough?

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -221,27 +198,46 @@ function getFieldClassInfos(field?: Field) {
...
+  const [loadedOptions, setLoadedOptions] = useState<UniqueValue[]>([]);
+  const [hasMore, setHasMore] = useState(false);
+
+  useEffect(() => {
+    setLoadedOptions([]);
```
saskliutas, line 214: Shouldn't we reset `loadedCount` too?

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
jasdom: `loadedCount` and `loadedOptions.length` can differ. The `loadedCount`  is used for determining the starting index for the next page and it keeps track of the total amount of values regardless of their content while `loadedOptions` does not include values with missing properties (for example, with no `displayValue`).

saskliutas: Ok. I still think all three pieces of state should be merged into one.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
jasdom: We should, fixed and updated tests to verify that the correct page index is used.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
jasdom: I agree, merged into one.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -196,27 +198,44 @@ function getFieldClassInfos(field?: Field) {
...
 
-function useUniquePropertyValuesLoader({ imodel, ruleset, field, descriptorInputKeys }: UseUniquePropertyValuesLoaderProps) {
+function useUniquePropertyValuesLoader({ imodel, property, descriptor, ruleset, field, descriptorInputKeys }: UseUniquePropertyValuesLoaderProps) {
+  const [loadedOptions, setLoadedOptions] = useState<{ count: number; options: UniqueValue[]; hasMore: boolean }>({ count: 0, options: [], hasMore: false });
+
+  useEffect(() => {
+    setLoadedOptions({ count: 0, options: [], hasMore: false });
+  }, [property, descriptor]);
+
   const loadTargets = useCallback(
```
saskliutas, line 215: I think `useCallback` can be removed.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -87,24 +89,24 @@ export function UniquePropertyValuesSelector(props: UniquePropertyValuesSelector
...
-
   const ruleset = useUniquePropertyValuesRuleset(descriptor.ruleset, field);
-  const loadTargets = useUniquePropertyValuesLoader({ imodel, ruleset, field, descriptorInputKeys });
+  const loadValues = useUniquePropertyValuesLoader({ imodel, property, descriptor, ruleset, field, descriptorInputKeys });
 
   return (
     <AsyncSelect
       value={selectedValues}
-      loadOptions={async (_, options) => loadTargets(options.length)}
+      loadOptions={async (_, options) => loadValues(searchInput, options.length)}
```
saskliutas, line 98: Shouldn't we use first argument of `loadOptions`? Also maybe it's worth debounce loading options while typing in selector? `AsyncSelect` supports passing debounce time if I remember correctly.

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
```diff
@@ -478,6 +481,250 @@ describe("UniquePropertyValuesSelector", () => {
...
+        <UniquePropertyValuesSelector property={propertyDescription} onChange={() => {}} imodel={testImodel} descriptor={descriptor} />,
+      </TestComponentWithPortalTarget>,
+    );
+
+    // open menu
+    const menuSelector = await waitFor(() => getByText("unique-values-property-editor.select-values"));
+    await user.click(menuSelector);
+
+    // add some delay to see if a second page load will be requested
+    await new Promise((f) => setTimeout(f, 1000));
```
saskliutas, line 498: This looks hacky. Can we remove this?

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
jasdom: Should we? I am using the `searchInput` instead of the passed value just because its already stored in lowercase. From my testing it seems like the state is always updated before the `loadOptions` call.

saskliutas: I think using search value passed to `loadOptions` are correct approach especially if we enable debounce. From our discussion I understood that `searchInput` state is mainly for refreshing cache inside `AsyncSelect`.

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
jasdom: I tried to avoid it when writing the test, however I was not able to since we need to wait for `AsyncSelect` to try to load the second page (which might not happen) so there is no element that we could wait for.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -226,27 +245,42 @@ function useUniquePropertyValuesLoader({ imodel, ruleset, field, descriptorInput
...
-          filteredOptions.push({ displayValue: option.displayValue, groupedRawValues: groupedValues });
+
+        if (groupedValues.length === 0) {
+          continue;
+        }
+
+        const uniqueValue = { displayValue: option.displayValue, groupedRawValues: groupedValues };
+        options.push(uniqueValue);
+
+        if (matchesSearchInput(uniqueValue)) {
```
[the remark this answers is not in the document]
jasdom: Moved the filtering.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
jasdom: Removed.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
jasdom: Yes the purpose of `searchInput` is for refreshing the cache. Switched to using the passed value and added debounce timeout.

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
saskliutas: Then maybe we don't need this test and it's enough to validate it manually?

jasdom: Removed the timeout part and kept the renamed test for coverage.

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
```diff
@@ -77,10 +77,16 @@ function NoOptionsMessage<TOption, IsMulti extends boolean = boolean>({ children
 }
 
 function ValueContainer<TOption, IsMulti extends boolean = boolean>({ children, ...props }: ValueContainerProps<TOption, IsMulti>) {
+  const childrenArray = Children.toArray(children);
+  const values = childrenArray.slice(0, -1);
```
saskliutas, line 81: I think we should check how many values are selected.

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
```diff
@@ -77,10 +77,16 @@ function NoOptionsMessage<TOption, IsMulti extends boolean = boolean>({ children
...
+  const input = childrenArray[childrenArray.length - 1];
   return (
-    <TagContainer {...props.innerProps} className="presentation-async-select-values-container" as={"div"}>
-      {children}
-    </TagContainer>
+    <div className="presentation-async-select-values-container">
+      {input}
+      <TagContainer {...props.innerProps} className="presentation-async-select-tag-container" overflow="truncate" as={"div"}>
+        {props.selectProps.menuIsOpen && props.selectProps.inputValue.length !== 0 ? undefined : values}
```
saskliutas, line 87: I think we should avoid rendering whole `TagContainer` if we are not rendering values.

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
jasdom: Changed to no longer render the `TagContainer` when not required.

packages/components/src/presentation-components/properties/inputs/AsyncSelect.tsx
jasdom: Sliced the children based on selected option count.

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
```diff
@@ -478,6 +481,281 @@ describe("UniquePropertyValuesSelector", () => {
...
+      // open menu
+      const menuSelector = await waitFor(() => getByText("unique-values-property-editor.select-values"));
+      await user.click(menuSelector);
+
+      // type in
+      const searchSelector = await waitFor(() => container.querySelector(".presentation-async-select-values-container"));
+      await user.type(searchSelector!, "test");
+
+      // ensure only the searched for value is shown
+      await waitFor(() => getByText("TestValue2"));
```
saskliutas, line 531: I think `expect(queryByText("TestValue2")).to.not.be.null` would be more descriptive. Also we probably should check that `Value1` is not visible. Same in other tests.

saskliutas (changes requested):

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
jasdom: Updated the tests.

saskliutas (approved):

PR #536, opened 2024-04-22
`onFilterApplied` and `onHierarchyLimitExceeded` callbacks

jasdom:
Added callbacks necessary for [792](https://github.com/iTwin/viewer-components-react/issues/792).

### DIFF
--- a/.changeset/great-eggs-leave.md
+++ b/.changeset/great-eggs-leave.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Added `onFilterApplied` and `onHierarchyLimitExceeded` callbacks.
--- a/.changeset/great-eggs-leave.md
+++ b/.changeset/great-eggs-leave.md
@@ -1,5 +1,5 @@
 ---
-"@itwin/presentation-components": patch
+"@itwin/presentation-components": minor
 ---
 
 Added `onFilterApplied` and `onHierarchyLimitExceeded` callbacks.
--- a/packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
+++ b/packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
@@ -35,6 +35,7 @@ import { useHierarchyLevelFiltering } from "./UseHierarchyLevelFiltering";
  */
 export interface PresentationTreeRendererProps extends Omit<TreeRendererProps, "nodeRenderer"> {
   nodeLoader: AbstractTreeNodeLoaderWithProvider<IPresentationTreeDataProvider>;
+  onFilterApplied?: () => void;
 }
 
 /**
@@ -53,13 +54,14 @@ export interface FilterableTreeProps {
  */
 export interface useFilterablePresentationTreeProps {
   nodeLoader: AbstractTreeNodeLoaderWithProvider<IPresentationTreeDataProvider>;
+  onFilterApplied?: () => void;
 }
 
 /**
  * Hook that enables hierarchy level filtering with action handlers for setting and clearing filters.
  * @beta
  */
-export function useFilterablePresentationTree({ nodeLoader }: useFilterablePresentationTreeProps): FilterableTreeProps {
+export function useFilterablePresentationTree({ nodeLoader, onFilterApplied }: useFilterablePresentationTreeProps): FilterableTreeProps {
   const { applyFilter, clearFilter } = useHierarchyLevelFiltering({ nodeLoader, modelSource: nodeLoader.modelSource });
   const [filterNode, setFilterNode] = useState<PresentationTreeNodeItem>();
 
@@ -70,6 +72,7 @@ export function useFilterablePresentationTree({ nodeLoader }: useFilterablePrese
         onApply={(info) => {
           info === undefined ? clearFilter(filterNode.id) : applyFilter(filterNode.id, info);
           setFilterNode(undefined);
+          onFilterApplied?.();
         }}
         onClose={() => {
           setFilterNode(undefined);
@@ -97,7 +100,10 @@ export function useFilterablePresentationTree({ nodeLoader }: useFilterablePrese
  * @beta
  */
 export function PresentationTreeRenderer(props: PresentationTreeRendererProps) {
-  const { onClearFilterClick, onFilterClick, filterDialog } = useFilterablePresentationTree({ nodeLoader: props.nodeLoader });
+  const { onClearFilterClick, onFilterClick, filterDialog } = useFilterablePresentationTree({
+    nodeLoader: props.nodeLoader,
+    onFilterApplied: props.onFilterApplied,
+  });
   const filterableNodeRenderer = (nodeProps: TreeNodeRendererProps) => {
     return <PresentationTreeNodeRenderer {...nodeProps} onFilterClick={onFilterClick} onClearFilterClick={onClearFilterClick} />;
   };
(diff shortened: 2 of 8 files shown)


packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
```diff
@@ -221,7 +233,7 @@ function useTreeState(props: UseTreeStateProps) {
   useEffect(() => {
     const { treeModel, ...providerProps } = props.treeStateProps;
     const modelSource = new TreeModelSource(new MutableTreeModel(treeModel));
-    const dataProvider = new PresentationTreeDataProvider(providerProps);
+    const dataProvider = new PresentationTreeDataProvider({ ...providerProps, onHierarchyLimitExceeded: onHierarchyLimitExceededRef.current });
```
saskliutas, line 236:
```suggestion
    const dataProvider = new PresentationTreeDataProvider({ ...providerProps, onHierarchyLimitExceeded: () => onHierarchyLimitExceededRef.current?.() });
```

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
```diff
@@ -70,6 +72,7 @@ export function useFilterablePresentationTree({ nodeLoader }: useFilterablePrese
         onApply={(info) => {
           info === undefined ? clearFilter(filterNode.id) : applyFilter(filterNode.id, info);
           setFilterNode(undefined);
+          onFilterApplied?.();
```
saskliutas, line 75: I don't think callback name matches when it's called. Because it will be also called when filter is removed. If we are only interested in filter being applied I think we should not invoke `onFilterApplied` when `info` is `undefined`.

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
```diff
@@ -213,6 +220,11 @@ function useTreeState(props: UseTreeStateProps) {
     onNodeLoadedRef.current = props.onNodeLoaded;
   }, [props.onNodeLoaded]);
 
+  const onHierarchyLimitExceededRef = useRef(props.onHierarchyLimitExceeded);
+  useEffect(() => {
+    onHierarchyLimitExceededRef.current = props.onHierarchyLimitExceeded;
+  }, [props.onHierarchyLimitExceeded]);
```
saskliutas, line 226: This pattern is repeated 3 times here. Can it be generalized and extracted into custom hook?

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
```diff
@@ -221,7 +233,7 @@ function useTreeState(props: UseTreeStateProps) {
   useEffect(() => {
     const { treeModel, ...providerProps } = props.treeStateProps;
     const modelSource = new TreeModelSource(new MutableTreeModel(treeModel));
-    const dataProvider = new PresentationTreeDataProvider(providerProps);
+    const dataProvider = new PresentationTreeDataProvider({ ...providerProps, onHierarchyLimitExceeded: onHierarchyLimitExceededRef.current });
     const pagedLoader = new PagedTreeNodeLoader(dataProvider, modelSource, providerProps.pagingSize);
     const nodeLoader = new ReportingTreeNodeLoader(pagedLoader, () => onNodeLoadedRef.current);
```
saskliutas, line 238: I missed this during PR review but looking at this now I think it would be more readable if this was not a getter for `onNodeLoaded` callback. I would suggest changing it into `(nodeLoadedProps) => onNodeLoadedRef.current?.(nodeLoadedProps)`

saskliutas (changes requested):

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
jasdom: Changed and added a test.

packages/components/src/presentation-components/tree/controlled/PresentationTreeRenderer.tsx
jasdom: My intention was to have it be called on filter clear too but it seems that is not required so changed to only call when a filter is added.

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
jasdom: Extracted into a `useLatest` hook.

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
```diff
@@ -208,22 +216,17 @@ interface PresentationTreeFilteringProps {
...
 
   useEffect(() => {
     const { treeModel, ...providerProps } = props.treeStateProps;
     const modelSource = new TreeModelSource(new MutableTreeModel(treeModel));
-    const dataProvider = new PresentationTreeDataProvider(providerProps);
+    const dataProvider = new PresentationTreeDataProvider({ ...providerProps, onHierarchyLimitExceeded: () => onHierarchyLimitExceededRef.current?.() });
     const pagedLoader = new PagedTreeNodeLoader(dataProvider, modelSource, providerProps.pagingSize);
-    const nodeLoader = new ReportingTreeNodeLoader(pagedLoader, () => onNodeLoadedRef.current);
+    const nodeLoader = new ReportingTreeNodeLoader(pagedLoader, (nodeLoadedProps) => onNodeLoadedRef.current?.(nodeLoadedProps));
+    const prevState = prevStateRef.current;
```
saskliutas, line 229: Why is this needed?

packages/components/src/presentation-components/hooks/UseLatest.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/** @packageDocumentation
+ * @module Core
+ */
+
+/**
+ * Hook for creating a reference that uses the latest passed value.
+ * @internal
+ */
+export function useLatest<T>(value: T) {
```
saskliutas, line 16: I don't think we need to move this into separate file. It is used only in one place.

saskliutas (changes requested):

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
jasdom: ESLint complains about the cleanup function: `The ref value 'prevStateRef.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'prevStateRef.current' to a variable inside the effect, and use that variable in the cleanup function.`.

packages/components/src/presentation-components/hooks/UseLatest.ts
jasdom: Moved it back.

packages/components/src/presentation-components/tree/controlled/UsePresentationTreeState.ts
saskliutas: Can you add eslint ignore for this and revert to how it was before? It should use changed (latest) value when cleanup runs.

jasdom: Done.

.changeset/great-eggs-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": minor
+---
+
+Added `onFilterApplied` and `onHierarchyLimitExceeded` callbacks.
```
saskliutas, line 5:
```suggestion
Added `onFilterApplied` and `onHierarchyLimitExceeded` callbacks for tracking when hierarchy level is filtered or exceeds the limit.
```

saskliutas (approved):

PR #538, opened 2024-04-23
Avoid changing `selectionSet` unnecessary when unified selection change is caused by `Viewport`

saskliutas:
Avoid remove/adding back elements from/to `selectionSet` when unified selection change is caused by selecting something in a viewport. Remove/adding elements causes tool buttons flicker if they are shown only when something is in `selectionSet`

### DIFF
--- a/apps/test-app/frontend/src/components/app/App.tsx
+++ b/apps/test-app/frontend/src/components/app/App.tsx
@@ -109,6 +109,26 @@ export function App() {
     };
   }, []);
 
+  useEffect(() => {
+    if (!state.imodel) {
+      return;
+    }
+
+    const removeSelectionListener = Presentation.selection.selectionChange.addListener((args: SelectionChangeEventArgs) => {
+      // eslint-disable-next-line no-console
+      console.log("Unified selection changed", args);
+    });
+    const selectionSetListener = state.imodel.selectionSet.onChanged.addListener((args) => {
+      // eslint-disable-next-line no-console
+      console.log("Selection set changed", args);
+    });
+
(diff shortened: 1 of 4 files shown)


apps/test-app/frontend/src/components/app/App.tsx
```diff
@@ -109,6 +109,26 @@ export function App() {
...
+    const selectionSetListener = state.imodel.selectionSet.onChanged.addListener((args) => {
+      // eslint-disable-next-line no-console
+      console.log("Selection set changed", args);
+    });
+
+    return () => {
+      removeSelectionListener();
+      selectionSetListener();
+    };
+  }, [state.imodel]);
```
jasdom, line 130: Were these changes added intentionally or are they debugging leftovers?

packages/components/src/test/viewport/WithUnifiedSelection.test.tsx
```diff
@@ -394,6 +399,29 @@ describe("ViewportSelectionHandler", () => {
       });
     });
 
+    it("does not clear selection set if unified selection change was caused by viewport", async () => {
+      const elementId = "0x";
```
jasdom, line 403:
```suggestion
      const elementId = "0x1";
```

jasdom (approved):

packages/components/src/test/viewport/WithUnifiedSelection.test.tsx
saskliutas: Changed

grigasp (approved):

PR #539, opened 2024-04-23
Hierarchies react split renderer

saskliutas:
Split Tree into `TreeRenderer` and `TreeNodeRenderer` to make is more convenient to customize.


packages/hierarchies-react/src/presentation-hierarchies-react/TreeRenderer.tsx
```diff
@@ -0,0 +1,92 @@
...
+        overflow: "auto",
+      }}
+    >
+      <Tree<PresentationTreeNode> data={rootNodes} nodeRenderer={nodeRenderer} getNode={getNode} enableVirtualization={true} />
+    </div>
+  );
+}
+
+/** @beta */
+export function getTreeNode(node: PresentationTreeNode, isNodeSelected: (nodeId: string) => boolean): ReturnType<TreeProps<PresentationTreeNode>["getNode"]> {
```
grigasp, line 60: maybe this should be called "create" instead?

saskliutas: Renamed

grigasp (approved):

Issue #515, opened 2024-04-05
hierarchies: Optimize code to reduce amount of main thread blocking


grigasp:
https://github.com/iTwin/presentation/issues/425 PoC proved that using a background worker for cases, when input to / output from the worker is large, doesn't really help us reduce the amount of main thread blocking, as (de)serializing the data has the same downsides as processing it.

Instead, we should look into doing a number of micro-optimizations to avoid blocking. Some ideas:
- don't re-create nodes when processing them in our rxjs operators,
- avoid converting nodes' stream to array for processing,
- ...

PR #542, opened 2024-04-25
hierarchies: Improvements for performance problems debugging

grigasp:
This is an initial set of changes related to https://github.com/iTwin/presentation/issues/515. The PR doesn't contain any perf improvements, but rather improves the tooling for performance profiling and debugging.

### DIFF
--- a/apps/performance-tests/tsconfig.json
+++ b/apps/performance-tests/tsconfig.json
@@ -4,7 +4,9 @@
     "outDir": "./lib",
     "sourceMap": true,
     "esModuleInterop": true,
-    "allowJs": true
+    "allowJs": true,
+    "declaration": true,
+    "declarationMap": true
   },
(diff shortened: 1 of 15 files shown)


apps/performance-tests/tsconfig.json
```diff
@@ -4,7 +4,9 @@
     "outDir": "./lib",
     "sourceMap": true,
     "esModuleInterop": true,
-    "allowJs": true
+    "allowJs": true,
+    "declaration": true,
```
Yato333, line 8: Is this needed?

grigasp: hm, I'm pretty sure I needed this for debugging *.ts rather than *.js, but I just checked this and it seems the change is not necessary - reverted.

Yato333 (approved):

PR #547, opened 2024-04-25
Expose hierarchy level size limit

saskliutas:
Hierarchy level size limit might be overridden for specific hierarchy level. This might be necessary information when building instance filter and checking if filter reduces instance count to set limit.

### DIFF
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
@@ -11,7 +11,7 @@ import { useUnifiedTreeSelection, UseUnifiedTreeSelectionProps } from "./interna
 import { PresentationHierarchyNode, PresentationTreeNode } from "./Types";
 
 /** @beta */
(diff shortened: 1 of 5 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -13,6 +13,7 @@ import { PresentationHierarchyNode, PresentationTreeNode } from "./Types";
 /** @beta */
 export interface HierarchyLevelFilteringOptions {
   hierarchyNode: HierarchyNode;
+  hierarchyLevelSizeLimit?: number | "unbounded";
```
grigasp, line 16: I don't feel like the limit belongs to an interface we call "filtering options". Maybe the interface (and the methods returning it) should be renamed to be more generic? E.g. `HierarchyLevelConfiguration` or something like that?

saskliutas: Rename

PR #548, opened 2024-04-26
hierarchies: Small performance optimizations

grigasp:
This is the first set of small optimizations for https://github.com/iTwin/presentation/issues/515:
- Use our own version of `partition` that iterates source observable only once.
- Use `merge`, `mergeAll`, `mergeMap` instead of `concat` versions where possible.
- Drop unnecessary `defer` usages.
- When passing nodes through our operators, mutate them with `Object.assign` instead of handling them in an immutable way, constantly creating new objects.
- Refactor grouping operator to use observables instead of plain arrays. This allows us start handling nodes as soon as they come rather than having to wait for them all to be reduced into an array first.


packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
```diff
@@ -89,56 +113,57 @@ export type GroupingHandler = (
...
+        map((result) => ({ handlerIndex: handlerIndex + 1, result: { ...result, grouped: mergeInPlace(curr?.grouped, result.grouped) } })),
+        log({
+          category: PERF_LOGGING_NAMESPACE,
+          message: /* istanbul ignore next */ () => `Total time for grouping handler ${handlerIndex}: ${timer.elapsedSeconds.toFixed(3)} s.`,
+        }),
+        delay(0),
+      );
+    }),
+    skip(groupingHandlers.length),
+    take(1),
```
Yato333, line 149: Maybe `last()` will work here?

Yato333 (commented):
Performance of `hide if no children required to finalize root, w/ children` test seems to have worsened significantly. Is this a red flag or just a random fluctuation?

packages/hierarchies/src/hierarchies/internal/operators/Partition.ts
```diff
@@ -0,0 +1,52 @@
...
+/** @internal */
+export function partition<T, U extends T>(source: ObservableInput<T>, predicate: (value: T) => value is U): [Observable<U>, Observable<Exclude<T, U>>];
+/** @internal */
+export function partition<T>(source: ObservableInput<T>, predicate: (value: T) => boolean): [Observable<T>, Observable<T>];
+
+/**
+ * This is similar to `rxjs` `partition` operator, but it subscribes to source observable only once.
+ * @internal
+ */
+export function partition<T>(source: ObservableInput<T>, predicate: (value: T) => boolean): [Observable<T>, Observable<T>] {
```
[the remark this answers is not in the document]
grigasp: The whole point of introducing our own implementation was to avoid having to traverse the source observable twice. BTW, my implementation only subscribes to source when consumer subscribes to one of the returned observables.

packages/hierarchies/src/hierarchies/internal/operators/Grouping.ts
grigasp: yes, fixed

grigasp:
> Performance of `hide if no children required to finalize root, w/ children` test seems to have worsened significantly. Is this a red flag or just a random fluctuation?

Re-ran the pipeline a couple of times, it looks like it's not random. Will have a look.

packages/hierarchies/src/hierarchies/internal/operators/Partition.ts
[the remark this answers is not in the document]
Yato333: Oh ok, I thought the main problem was that RXJS subscribes 2 times to the same observable without sharing the replay

packages/hierarchies/src/hierarchies/internal/operators/Partition.ts
```diff
@@ -0,0 +1,52 @@
...
+  const matches = new ReplaySubject<T>();
+  const nonMatches = new ReplaySubject<T>();
+  let didSubscribe = false;
+  function subscribe() {
+    if (didSubscribe) {
+      return;
+    }
+    didSubscribe = true;
+    from(source).subscribe({
```
saskliutas, line 26: Shouldn't we unsubscribe if `matches` and `nonMatches` observables are unsubscribed?

Yato333 (approved):

PR #552, opened 2024-05-02
Fix tree auto update

saskliutas:
Need to pass ruleset instead of id to the data provider when reloading tree because that ruleset might not be registered on presentation-frontend.

Fixes https://github.com/iTwin/viewer-components-react/issues/827

### DIFF
--- a/packages/components/src/presentation-components/tree/controlled/UseTreeReload.ts
+++ b/packages/components/src/presentation-components/tree/controlled/UseTreeReload.ts
@@ -6,10 +6,12 @@
  * @module Tree
  */
 
-import { useEffect } from "react";
+import { MutableRefObject, useEffect } from "react";
 import { RenderedItemsRange, Subscription, TreeModelSource } from "@itwin/components-react";
 import { IModelApp } from "@itwin/core-frontend";
+import { Ruleset } from "@itwin/presentation-common";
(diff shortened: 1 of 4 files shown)


packages/components/src/presentation-components/tree/controlled/UseTreeReload.ts
```diff
@@ -38,96 +40,97 @@ export function useTreeReload(params: TreeReloadParams & { enable: boolean }) {
...
   useEffect(() => {
-    if (!enable || !rulesetId || !modelSource) {
+    if (!enable || !modelSource) {
       return;
     }
 
     let subscription: Subscription | undefined;
     const removeListener = Presentation.presentation.onIModelHierarchyChanged.addListener((args: IModelHierarchyChangeEventArgs) => {
-      if (args.rulesetId !== rulesetId || args.imodelKey !== dataProviderProps.imodel.key) {
+      if (args.rulesetId !== ruleset || args.imodelKey !== dataProviderProps.imodel.key) {
```
grigasp, line 52: Should `args.rulesetId !== ruleset` check instead use `getRulesetId(ruleset)`?

saskliutas: Missed this one.

grigasp (approved):

Issue #537, opened 2024-04-22
unified-selection: Add performance tests


saskliutas:
The `unified-selection` package contains APIs that run multiple complex queries and we have no idea how well the queries perform. We should add performance tests for that and set up a benchmark action for unified-selection related PRs.

PR #557, opened 2024-05-03
Unified-selection performance tests

jasdom:
Closes [537](https://github.com/iTwin/presentation/issues/537).


apps/performance-tests/src/unified-selection/HiliteSet.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  const additionalElementCount = 5;
+  const additionalFunctionalElement2dCount = 1;
+
+  function createSelection({ fullClassName, start, count }: { fullClassName: string; start: number; count: number }) {
+    const selectables: Selectable[] = [];
+    for (let i = start; i < start + count + additionalElementCount; i++) {
+      selectables.push({ className: fullClassName, id: `0x${i.toString(16)}` });
+    }
+    return Selectables.create(selectables);
+  }
```
grigasp, line 24: Not a fan of all those magic numbers... What do they mean?

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  runHiliteTest({
+    testName: "50k elements",
+    iModelName: "50k nested elements",
+    selection: createSelection({ fullClassName: "Generic:PhysicalObject", start: 14, count: 50_000 }),
+    expectedCounts: { elements: 50000 },
+  });
```
grigasp, line 31: I don't understand this. As far as I can see, the `50k nested elements` iModel has 25k 2d and 25k 3d elements. 
- The input (selection) is 50k 3d elements - what elements are they, if the iModel only contains 25k?
- How do we get 50k elements in the results if the iModel only has 25k 3d elements?

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -127,4 +164,203 @@ export class Datasets {
 
     console.log(`${numElements} elements: Done.`);
   }
+
+  private static async createNestedElementModel(name: string, localPath: string, numElements: number) {
```
grigasp, line 168: Please add a comment next to each of these `create***IModel` functions explaining the IModel content. For example:

```
Create an iModel with `numElements` elements, half of them 2d and the other half - 3d, all belonging to the same 2d
and 3d model and category. The elements are set up in hierarchical manner, with 1000 top level elements each having 1
child element, which has 1 child element, and so on until the depth of 25 elements is reached.
```

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    testName: "50k group elements",
+    iModelName: "50k group member elements",
+    selection: createSelection({ fullClassName: "Generic:Group", start: 15, count: 50_000 }),
+    expectedCounts: { elements: 50000 },
+  });
+
+  runHiliteTest({
+    testName: "1k subjects",
+    iModelName: "1k nested subjects",
+    selection: createSelection({ fullClassName: "BisCore:Subject", start: 11, count: 2_000 }),
```
grigasp, line 43: What does `count: 2_000` mean if we only have 1k subjects?

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    expectedCounts: { models: 20 },
+  });
+
+  runHiliteTest({
+    testName: "50k subcategories",
+    iModelName: "50k subcategories",
+    selection: createSelection({ fullClassName: "BisCore:SpatialCategory", start: 11, count: 2_000 }),
+    expectedCounts: { subCategories: 50001 },
```
grigasp, line 51: why not 50k?

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    testName: "1k subjects",
+    iModelName: "1k nested subjects",
+    selection: createSelection({ fullClassName: "BisCore:Subject", start: 11, count: 2_000 }),
+    expectedCounts: { models: 20 },
+  });
+
+  runHiliteTest({
+    testName: "50k subcategories",
+    iModelName: "50k subcategories",
+    selection: createSelection({ fullClassName: "BisCore:SpatialCategory", start: 11, count: 2_000 }),
```
grigasp, line 50: same as above - what does `count: 2_000` mean here?

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -127,4 +164,203 @@ export class Datasets {
...
+        for (let i = 0; i < membersPerGroup; ++i) {
+          let parentId: string | undefined;
+
+          for (let j = 0; j < elementsPerMember; ++j) {
+            parentId = insertPhysicalElement({ builder, parentId, modelId, categoryId }).id;
+            if (j === 0) {
+              builder.insertRelationship({ sourceId: groupId, targetId: parentId, classFullName: "BisCore.ElementGroupsMembers" });
+            }
+          }
+        }
```
grigasp, line 270: Do we need the grouped elements have children? Maybe it would be better to have more groups with more direct members, without any child elements - that would isolate the grouping case from elements' hierarchy case.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
```diff
@@ -0,0 +1,117 @@
...
+    expectedCounts: { subCategories: 50001 },
+  });
+
+  runHiliteTest({
+    testName: "50k functional 3D elements",
+    iModelName: "50k nested functional 3D elements",
+    selection: createSelection({ fullClassName: "Functional:FunctionalElement", start: 16, count: 100_000 }),
```
grigasp, line 57: what does `count: 100_000` mean here?

apps/performance-tests/src/unified-selection/SelectionScope.test.ts
```diff
@@ -0,0 +1,166 @@
...
+import { createECSqlQueryExecutor } from "@itwin/presentation-core-interop";
+import { computeSelection } from "@itwin/unified-selection";
+import { ComputeSelectionProps } from "@itwin/unified-selection/lib/cjs/unified-selection/SelectionScope";
+import { Datasets, IModelName } from "../util/Datasets";
+import { run, RunOptions } from "../util/TestUtilities";
+
+describe("compute", () => {
+  const additionalElementCount = 5;
+
+  function createElementKeys({ start, count }: { start: number; count: number }) {
```
grigasp, line 17: Looks like you're creating ids here rather than keys

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -75,6 +91,27 @@ export class Datasets {
...
+      case "50k group member elements":
+        return async (name: string, localPath: string) => this.createNestedGroupModel(name, localPath, elementCount);
+      case "1k nested subjects":
+        return async (name: string, localPath: string) => this.createNestedSubjectModel(name, localPath, elementCount);
+      case "50k subcategories":
+        return async (name: string, localPath: string) => this.createNestedCategoriesModel(name, localPath, elementCount);
+      case "50k nested functional 3D elements":
+        return async (name: string, localPath: string) => this.createFunctionalElements3dModel(name, localPath, elementCount);
+      case "10k nested functional 2D elements":
+        return async (name: string, localPath: string) => this.createNestedFunctionalElements2dModel(name, localPath, elementCount);
```
grigasp, line 109: There's inconsistency with iModel names and their factory function names. I'd rather drop the "nested" term from both - key and function name.

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -127,4 +164,203 @@ export class Datasets {
...
+      const numberOfGroups = 1000;
+      const elementsPerGroup = numElements / numberOfGroups;
+
+      for (let i = 0; i < numberOfGroups; ++i) {
+        let functionalElementParentId: string | undefined;
+        let graphicsElementParentId = insertDrawingGraphic({
+          builder,
+          categoryId,
+          modelId: drawingModelId,
+        }).id;
```
grigasp, line 336: why is this top-level element created outside the loop?

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -127,4 +164,203 @@ export class Datasets {
...
+
+        for (let k = 0; k < elementsPerGroup; ++k) {
+          functionalElementParentId = insertFunctionalElement({
+            parentId: functionalElementParentId,
+            builder,
+            modelId: functionalModelId,
+            representedElementId: k === elementsPerGroup - 1 ? graphicsElementParentId : graphicsElementId,
```
grigasp, line 343: Please extract either condition or the resulting element id into a variable whose name explains the meaning. At the moment it's not clear what you're doing here.

apps/performance-tests/src/unified-selection/SelectionScope.test.ts
```diff
@@ -0,0 +1,166 @@
...
+      };
+    },
+    test: async (props) => {
+      const iterator = computeSelection({ elementIds: testProps.elementIds, scope: testProps.scope, queryExecutor: props.queryExecutor });
+      const counts = new Map<string, number>();
+
+      for await (const instanceKey of iterator) {
+        const count = counts.get(instanceKey.className) ?? 0;
+        counts.set(instanceKey.className, count + 1);
+      }
```
grigasp, line 156: I'm a worried the Map lookup is causing the tests to be slower than necessary. Can you try the following to see if there's any positive effect?

```suggestion
      const counts = new Map<string, { count: number }>();

      for await (const instanceKey of iterator) {
        const res = counts.get(instanceKey.className);
        if (res) {
          ++res.count;
        } else {
          res.set(instanceKey.className, { count: 1 });
        }
      }
```

grigasp (commented):
It's not clear to me why we only got 1 PR comment from the action. You made changes to hierarchies-related workflow and performance test files, so I'd expect 2 comments - one for hierarchies and one for unified selection. Can you check?

apps/performance-tests/src/unified-selection/SelectionScope.test.ts
jasdom: Switched to the suggestion, does not seem to have any real effect on the results.

apps/performance-tests/src/util/Datasets.ts
jasdom: Added a comment instead.

apps/performance-tests/src/util/Datasets.ts
jasdom: Refactored the entire functional 2D iModel setup, to make it more clear.

apps/performance-tests/src/unified-selection/SelectionScope.test.ts
jasdom: Renamed.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: Same as with subjects, changed.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: Same as with subjects, changed.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: Seems like when a spatial category is inserted a subcategory is also added, changed the iModel to generate one subcategory less.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: Since additional elements were inserted between subjects, their ids were not sequential. Instead of doing this, updated `createSelection` function to create keys based on passed `step` value.

apps/performance-tests/src/util/Datasets.ts
jasdom: Added comments.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: That's because the class name is only relevant to determine what the type of the element is, in this case 50k elements were passed as `Generic:PhysicalObject` but the query was treating it as `BisCore:Element` so the 25k 2D elements were also included. This was intended except for the fact that ``BisCore:Element` was supposed to be used instead of `Generic:PhysicalObject`. Fixed.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: I messed up the starting indices so it was using selecting too many elements. Fixed.

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -75,6 +91,27 @@ export class Datasets {
     return path.resolve(localPath);
   }
 
+  private static getIModelFactory(key: IModelName, elementCount: number) {
+    switch (key) {
+      case "50k elements":
+        return async (name: string, localPath: string) => this.createFlatIModel(name, localPath, elementCount);
```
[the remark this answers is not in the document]
jasdom: Renamed.

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -5,10 +5,32 @@
...
-export const IMODEL_NAMES = ["baytown", "50k elements"] as const;
+export const IMODEL_NAMES = [
+  "baytown",
+  "50k elements",
+  "50k nested elements",
+  "50k group member elements",
+  "1k nested subjects",
+  "50k subcategories",
+  "50k nested functional 3D elements",
+  "10k nested functional 2D elements",
```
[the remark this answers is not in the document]
jasdom: The query was timing out, though it seems that was happening because it was performing a larger query than I intended. Fixed and switched to using 50k 2D elements.

apps/performance-tests/src/util/Datasets.ts
```diff
@@ -75,6 +91,27 @@ export class Datasets {
...
+      case "1k nested subjects":
+        return async (name: string, localPath: string) => this.createNestedSubjectModel(name, localPath, elementCount);
+      case "50k subcategories":
+        return async (name: string, localPath: string) => this.createNestedCategoriesModel(name, localPath, elementCount);
+      case "50k nested functional 3D elements":
+        return async (name: string, localPath: string) => this.createFunctionalElements3dModel(name, localPath, elementCount);
+      case "10k nested functional 2D elements":
+        return async (name: string, localPath: string) => this.createNestedFunctionalElements2dModel(name, localPath, elementCount);
+      default:
+        return async () => {};
```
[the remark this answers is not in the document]
jasdom: Excluded `baytown` from `key`.

.github/workflows/unified-selection-benchmark-pr.yml
```diff
@@ -0,0 +1,68 @@
...
+    paths:
+      - apps/performance-tests/**
+      - packages/unified-selection/**
+      - pnpm-lock.yaml
+      - .github/workflows/unified-selection-benchmark-pr.yml
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  benchmark:
+    name: Benchmark performance (unified-selection)
```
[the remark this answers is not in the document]
jasdom: Updated.

.github/workflows/unified-selection-benchmark-pr.yml
```diff
@@ -0,0 +1,68 @@
...
+on:
+  pull_request:
+    branches:
+      - master
+      - stable
+    paths:
+      - apps/performance-tests/**
+      - packages/unified-selection/**
+      - pnpm-lock.yaml
+      - .github/workflows/unified-selection-benchmark-pr.yml
```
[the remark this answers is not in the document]
jasdom: Updated.

.github/workflows/unified-selection-benchmark-pr.yml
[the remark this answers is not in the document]
jasdom: Renamed.

.github/workflows/benchmark-pr.yml
```diff
@@ -1,4 +1,4 @@
-name: PR performance benchmark
+name: PR hierarchies performance benchmark
 
 on:
   pull_request:
```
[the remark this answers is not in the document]
jasdom: Updated.

.github/workflows/benchmark-pr.yml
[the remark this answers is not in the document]
jasdom: Renamed.

.github/workflows/benchmark-pr.yml
```diff
@@ -16,7 +16,7 @@ on:
 
 jobs:
   benchmark:
-    name: Benchmark performance
+    name: Benchmark performance (hierarchies)
```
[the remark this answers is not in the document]
jasdom: Updated.

package.json
```diff
@@ -25,7 +25,9 @@
     "change:empty": "changeset add --empty",
     "check": "changeset status",
     "release": "npm run build:all && changeset publish",
-    "benchmark": "pnpm run -C ./apps/performance-tests benchmark"
+    "benchmark": "pnpm run -C ./apps/performance-tests benchmark",
```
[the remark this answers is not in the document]
jasdom: `benchmark.yml` uses the script.

apps/performance-tests/src/util/Datasets.ts
jasdom: I added the children as the hilite elements query contains a step for traversing children of group members which would otherwise not be tested. Should I still remove the children in this case?

package.json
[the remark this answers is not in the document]
grigasp: Okay. Please add `shared` and `unified-selection` packages to the list of paths in that `yml`.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
grigasp: The `start` and `step` are still magic and are very fragile. E.g. it's easily possible that an extra element is inserted into an iModel in a future itwinjs-core version and that will result in a hard-to-debug test failure here. Instead, I suggest providing an ECSQL query to select ECInstanceIds that you want to create the `Selectables` from. Of course, that query has to run outside of the test to make sure it's execution is not included in our performance measure.

package.json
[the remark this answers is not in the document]
jasdom: Right, added.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
grigasp: Please see [my other comment](https://github.com/iTwin/presentation/pull/557#discussion_r1590905205) which, I think, applies here.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: Would we not run into a similar problem with ECSQL queries? For example, if we use a query to select all elements and then try to hilite them in the test. If an additional element is inserted the query would return it too. This is already the case for the functional 2D element iModel that contains an extra element which is not included in the test but would be returned by the query.

jasdom: Nvm, fetched the ids based on `UserLabel`.

grigasp: I actually think that supplying a whole ECSQL query, rather than full class name + user label, would be both - easier to understand and more flexible. So instead of:
```ts
{
  fullClassName: "Generic:PhysicalObject",
  userLabel: "test_element",
}
```

do:
```ts
{
  inputQuery: `SELECT ECInstanceId FROM Generic.PhysicalObject WHERE UserLabel = 'test_element'`,
}
```

I also don't think you need to pick those elements by label and should be able to pick them by class.

apps/performance-tests/src/util/Datasets.ts
grigasp: Okay, I guess that makes sense. But maybe we should focus more on grouped elements here, rather than child ones, and increase `membersPerGroup` to `1000`?

grigasp: And I think the function description needs to be clarified. I understand a "group element" as a `GroupInformationElement`, but this kind of element doesn't have a category and doesn't belong to physical model. Instead, we got 10 group (`GroupInformationElement`) elements and each of them have related physical elements through the `ElementGroupsMembers` relationship.

jasdom: Increased the number of group members and fixed the description.

apps/performance-tests/src/unified-selection/HiliteSet.test.ts
jasdom: Supplied the query directly and changed to use user labels only where necessary.

jasdom:
> It's not clear to me why we only got 1 PR comment from the action. You made changes to hierarchies-related workflow and performance test files, so I'd expect 2 comments - one for hierarchies and one for unified selection. Can you check?

To leave separate comments, each benchmark needs to have a unique `name` property, otherwise, a single comment would be used and the results would override each other, however the comment is only left when there is a previous run to compare to, and since the benchmark names changed, no such benchmark is found, so no comment is left until the pull request is merged.

The comment you are seeing is the one left when benchmark names are not changed, so the results of the new tests would be visible while the changes are in-review.

grigasp (approved):

grigasp (approved):

PR #560, opened 2024-05-06
Make sure nodes load duration is reported always

saskliutas:
Moved nodes load duration reporting into `next` callback to make sure it is reported immediately after nodes are loaded. Reporting from `complete` callback leaves open possibility to unsubscribe between `next` and `complete` without reporting duration.

### DIFF
--- a/.changeset/odd-plums-marry.md
+++ b/.changeset/odd-plums-marry.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Report nodes load duration immediately after nodes are loaded.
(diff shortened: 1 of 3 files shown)


.changeset/odd-plums-marry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Report nodes load duration immediately after nodes are loaded.
```
grigasp, line 5: I think this should be re-phrased to clarify what changed for consumers:

```suggestion
Fixed `usePresentationTreeState` callback prop `onNodeLoaded` not being called under certain race conditions.
```

PR #561, opened 2024-05-06
Load unique property values only of instances specified by the input keys

saskliutas:
When content descriptor does not contain ruleset we use `ContentInstancesOfSpecificClasses` specification to load unique property values. This specification loads content for all instances and does not take into account supplied input keys. In case input keys are provided use `SelectedNodeInstances` specification to load property only for the instances specified by the input keys.

### DIFF
--- a/.changeset/empty-birds-fail.md
+++ b/.changeset/empty-birds-fail.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Load only unique property values of instances specified by input keys.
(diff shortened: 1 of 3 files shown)


.changeset/empty-birds-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": patch
+---
+
+Load only unique property values of instances specified by input keys.
```
grigasp, line 5: I think this needs clarification:

```suggestion
Fix instance filter builder component's unique values selector not filtering values by provided instance keys.
```

grigasp (approved):

PR #566, opened 2024-05-07
Split benchmark.yml

jasdom:
`benchmark.yml` needs to be split into separate workflows to store test results separately.

### DIFF
--- a/.github/workflows/benchmark-hierarchies.yml
+++ b/.github/workflows/benchmark-hierarchies.yml
@@ -7,6 +7,7 @@ on:
       - master
     paths:
       - apps/performance-tests/**
+      - apps/performance-tests/util/**
       - packages/core-interop/**
       - packages/hierarchies/**
       - packages/models-tree/**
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmark-hierarchies.yml
```diff
@@ -7,15 +7,15 @@ on:
       - master
     paths:
       - apps/performance-tests/**
```
grigasp, line 9: should this point to `hierarchies` sub-directory?

.github/workflows/benchmark-unified-selection.yml
```diff
@@ -0,0 +1,74 @@
+name: Unified-selection performance benchmark
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    paths:
+      - apps/performance-tests/**
```
grigasp, line 9: shouldn't this be removed?

.github/workflows/benchmark-hierarchies.yml
jasdom: Yes, it should. Fixed.

.github/workflows/benchmark-unified-selection.yml
jasdom: Removed.

grigasp (approved):

Issue #481, opened 2024-03-19
Investigate different selection modes support in `@itwin/hierarchies-react`


saskliutas:
Current `ControlledTree`  component support multiple selection modes: `Ctrl` + Select, `Shift` + Select, Single select. Need to investigate ability of supporting these modes in new implementation.

jasdom:
Need to wait for changes to be made in iTwinUI as support for a similar feature is planned https://github.com/iTwin/iTwinUI/discussions/2027.

grigasp:
> Need to wait for changes to be made in iTwinUI as support for a similar feature is planned [iTwin/iTwinUI#2027](https://github.com/iTwin/iTwinUI/discussions/2027).

It doesn't look like iTwinUI team is going to look into this anytime soon. Could you evaluate what it would take to implement this on our end?

jasdom:
> > Need to wait for changes to be made in iTwinUI as support for a similar feature is planned [iTwin/iTwinUI#2027](https://github.com/iTwin/iTwinUI/discussions/2027).
> 
> It doesn't look like iTwinUI team is going to look into this anytime soon. Could you evaluate what it would take to implement this on our end?

My first guess is it would not take very long to implement on our side, however the main issue is that iTwinUI does not expose the events needed to detect whether shift or ctrl was used. We would have to either use global event listeners or wrap each node into a wrapper element and rely on event propagation to detect click events. None of these seem like good options.

jasdom:
> > > > Need to wait for changes to be made in iTwinUI as support for a similar feature is planned [iTwin/iTwinUI#2027](https://github.com/iTwin/iTwinUI/discussions/2027).
> > > 
> > > 
> > > It doesn't look like iTwinUI team is going to look into this anytime soon. Could you evaluate what it would take to implement this on our end?
> > 
> > 
> > My first guess is it would not take very long to implement on our side, however the main issue is that iTwinUI does not expose the events needed to detect whether shift or ctrl was used. We would have to either use global event listeners or wrap each node into a wrapper element and rely on event propagation to detect click events. None of these seem like good options.
> 
> Maybe the options aren't good, but as long as they don't require us to spend a ton of time on them and can be implemented in a way that lets us remove the implementation afterwards without introducing breaking changes, I think we need to work on this. This is a show-stopper feature for us.

Okay, in that case I'll start working on a temporary implementation.

PR #567, opened 2024-05-09
Hierarchies-react: selection mode support

jasdom:
Closes [481](https://github.com/iTwin/presentation/issues/481).


packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -113,9 +114,9 @@ export class TreeActions {
     return TreeModel.getNode(this._currentModel, nodeId);
   }
 
-  public selectNode(nodeId: string, isSelected: boolean) {
+  public selectNode(nodeIds: Array<string>, changeType: SelectionChangeType) {
```
saskliutas, line 117: I guess now `selectNode` should be renamed into `selectNodes` as it takes array instead of single node.

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+/** @internal */
+export type SelectionChangeType = "add" | "replace" | "remove";
+
+interface UseSelectionHandlerProps {
+  rootNodes: Array<PresentationTreeNode> | undefined;
+  selectNode: (nodeIds: Array<string>, changeType: SelectionChangeType) => void;
+  selectionMode: SelectionMode;
+}
+
+interface useSelectionHandlerResult {
```
saskliutas, line 40:
```suggestion
interface UseSelectionHandlerResult {
```

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+    selection.select !== "range" && (previousSelectionRef.current = nodeId);
+    selectNodeRef.current(nodes, selection.type);
+  };
+
+  const onNodeClick = useRef((nodeId: string, isSelected: boolean, event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+    return onNodeSelect(nodeId, isSelected, event.shiftKey, event.ctrlKey);
+  }).current;
+
+  const onNodeKeyDown = useRef((nodeId: string, isSelected: boolean, event: React.KeyboardEvent<HTMLElement>) => {
+    if (event.key === " " || event.key === "Spacebar" || event.key === "Enter") {
```
saskliutas, line 99: Do we need `" "` and `"Spacebar"` or one of these should be enough?

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+    if (selection.type === "disabled") {
+      return;
+    }
+
+    const nodes = selection.select === "range" ? getNodeRange(previousSelectionRef.current, nodeId) : [nodeId];
+    selection.select !== "range" && (previousSelectionRef.current = nodeId);
+    selectNodeRef.current(nodes, selection.type);
+  };
+
+  const onNodeClick = useRef((nodeId: string, isSelected: boolean, event: React.MouseEvent<HTMLElement, MouseEvent>) => {
```
saskliutas, line 94: Do we need to have stable reference for `onNodeClick` and `onNodekeyDown`?

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+  Single = SelectionModeFlags.Single,
+  /** Only one item selected at a time; allows deselecting. */
+  SingleAllowDeselect = SelectionModeFlags.Single | SelectionModeFlags.DeselectEnabled,
+  /** Toggles items. */
+  Multiple = SelectionModeFlags.DeselectEnabled,
+  /** Toggles items; allows the use of Ctrl & Shift Keys. */
+  Extended = SelectionModeFlags.KeysEnabled | SelectionModeFlags.DeselectEnabled,
+  /** Disables selection */
+  None = SelectionModeFlags.None,
+}
```
saskliutas, line 29: @grigasp Do you think we need all of this? Or can we simplify this into `export type SelectionMode = "none" | "single" | "extended" | "multiple".

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+
+interface FlatTreeState {
+  flatNodeList: Array<string>;
+  nodeIdToIndexMap: Map<string, number>;
+}
+
+/** @beta */
+export function useSelectionHandler(props: UseSelectionHandlerProps): useSelectionHandlerResult {
+  const { rootNodes, selectionMode, selectNode } = props;
+  const selectionModeRef = useLatest(selectionMode);
```
saskliutas, line 53: Enum is primitive value so there is no need to wrap it into `useRef`

packages/hierarchies-react/src/test/internal/UseUnifiedSelection.test.tsx
```diff
@@ -21,11 +22,14 @@ describe("useUnifiesSelection", () => {
...
   const source = "test-source";
   const getNode = sinon.stub<[string], TreeModelNode | undefined>();
+  const getNodeRange = sinon.stub<[string?, string?], TreeModelHierarchyNode[]>();
 
   const initialProps = {
     imodelKey,
     sourceName: source,
     getNode,
+    getNodeRange,
+    selectionMode: SelectionMode.Multiple,
```
saskliutas, line 32: I guess these are leftovers from initial implementation?

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
jasdom: Fixed.

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
jasdom: Based on documentation, older browsers can return "Spacebar" instead of " ".

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
jasdom: It is not necessary, removed.

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
jasdom: Switched to string union.

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
jasdom: Removed.

packages/hierarchies-react/src/test/internal/UseUnifiedSelection.test.tsx
jasdom: Removed.

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
```diff
@@ -0,0 +1,135 @@
...
+      return { select: "node", type: "add" };
+    case "extended":
+      if (shiftDown) {
+        return { select: "range", type: "replace" };
+      }
+      if (ctrlDown) {
+        return { select: "node", type: "add" };
+      }
+      return { select: "node", type: "replace" };
+  }
```
saskliutas, line 113: Can you separate different selection modes handling to make it easier to read? I would suggest something like this:
switch(selectionMode) {
    case "node":
      return { select: "node", type: "disabled" };
    case "single":
      return { select: "node", type: "replace" };
    case "extended":
      return getExtendedSelectionAction();
    case "multiple":
      return getMultipleSelectionAction();
}

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
jasdom: That is definitely better, changed.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,261 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { expect } from "chai";
+import sinon from "sinon";
+import { renderHook } from "@testing-library/react";
```
saskliutas, line 8: Use our wrapper from `TestUtils`.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,261 @@
...
+import { PresentationHierarchyNode, PresentationInfoNode, PresentationTreeNode } from "../presentation-hierarchies-react";
```
saskliutas, line 9: Avoid importing from barrel file.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,261 @@
...
+describe("useSelectionHandler", () => {
+  const selectNodesStub = sinon.stub<[Array<string>, SelectionChangeType], void>();
+
+  const createMouseEvent = (key?: "Shift" | "Ctrl") => {
+    return { shiftKey: key === "Shift", ctrlKey: key === "Ctrl" } as unknown as React.MouseEvent<HTMLDivElement, MouseEvent>;
+  };
+
+  const createKeyboardEvent = (mainKey?: " " | "Spacebar" | "Enter", extraKey?: "Shift" | "Ctrl") => {
+    return { key: mainKey, shiftKey: extraKey === "Shift", ctrlKey: extraKey === "Ctrl" } as unknown as React.KeyboardEvent<HTMLDivElement>;
+  };
```
saskliutas, line 21: Maybe instead of faking events we could render hook inside test component and perform actions using `user-event` library? It has more logic to replicated how interactions work in browser. Something like this: 

```ts
function TestComponent({ rootNodes, onSelect, selectionMode }) {
  const { onNodeKeyDown, onNodeClick } = useSelectionHandler({ rootNodes, onSelect, selectionMode });
  return <>
    {rootNodes.map((node) => {
      return <div 
        onClick={(e) => onNodeClick(node.id, !node.isSelected, e)} 
        onKeyDown={(e) => onNodeKeyDown(node.id, !node.isSelected, e)}
      >
        {node.label}
      </div>;
    })}
  </>
}

```

saskliutas (changes requested):

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
jasdom: Switched to wrapper.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
jasdom: Switched to `user-event` to simulate clicks.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+
+    describe("`none` selection mode", () => {
+      it("does nothing when node is clicked", async () => {
+        const rootNode = createHierarchyNode("node");
+        const { user, getByText } = render(<TestComponent {...createProps([rootNode], "none", true)} />);
+
+        const node = getByText("node");
+        node.focus();
+
+        await user.keyboard(`{Control>}`);
```
saskliutas, line 81: Why `Ctrl` used here?

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+      const selectionMode = "single";
+
+      it("replaces selection when node is clicked", async () => {
+        const rootNode = createHierarchyNode("node");
+        const { user, getByText } = render(<TestComponent {...createProps([rootNode], selectionMode, true)} />);
+
+        const node = getByText("node");
+        node.focus();
+
+        await user.keyboard(`{Shift>}`);
```
saskliutas, line 99: Is `Shift` needed here?

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+      });
+
+      it("removes from selection when a selected node is clicked", async () => {
+        const rootNode = createHierarchyNode("node");
+        const { user, getByText } = render(<TestComponent {...createProps([rootNode], selectionMode, false)} />);
+
+        const node = getByText("node");
+        node.focus();
+
+        await user.keyboard(`{Control>}`);
```
saskliutas, line 113: Do we need `Ctrl` here?

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+      const selectionMode = "multiple";
+
+      it("adds to selection when node is clicked", async () => {
+        const rootNode = createHierarchyNode("node");
+        const { user, getByText } = render(<TestComponent {...createProps([rootNode], selectionMode, true)} />);
+
+        const node = getByText("node");
+        node.focus();
+
+        await user.keyboard(`{Control>}`);
```
saskliutas, line 131: Do we need `Ctrl` here?

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+      });
+
+      it("removes from selection when a selected node is clicked", async () => {
+        const rootNode = createHierarchyNode("node");
+        const { user, getByText } = render(<TestComponent {...createProps([rootNode], selectionMode, false)} />);
+
+        const node = getByText("node");
+        node.focus();
+
+        await user.keyboard(`{Shift>}`);
```
saskliutas, line 145: Do we need `Ctrl` here?

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+        node2.focus();
+
+        await user.keyboard(`{Shift>}`);
+        await clickNode(user, node2);
+        await user.keyboard(`{/Shift}`);
+
+        expect(selectNodesStub).to.be.calledOnceWith(["node-1", "node-2"], "replace");
+      });
+
+      it("does not update previous selection when using `shift`", async () => {
```
saskliutas, line 319: Test does many selection changes and it is not clear from test name what it checks. Can you update test name and clarify what is the goal of it?

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+        node.focus();
+
+        await user.keyboard(`{Shift>}`);
+        await clickNode(user, node);
+        await user.keyboard(`{/Shift}`);
+
+        expect(selectNodesStub).to.be.calledOnceWith(["node-3", "node-4"], "replace");
+      });
+
+      it("clears selection when invalid node passed", async () => {
```
saskliutas, line 370: We probably should ignore if there is something `onNodeClicked` is called with invalid node.

saskliutas (changes requested):

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
jasdom: Changed to do nothing if invalid node is selected.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
jasdom: Renamed the test.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
jasdom: Added these extra keys to check if `ctrl` and `shift` would be ignored in non extended selection mode. Moved this into separate tests to be more clear.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+  const createProps = (rootNodes: Array<PresentationTreeNode> | undefined, selectionMode: SelectionMode, isSelected: boolean) => {
+    return { rootNodes, selectNodes: selectNodesStub, selectionMode, isSelected };
+  };
+
+  afterEach(() => {
+    selectNodesStub.reset();
+  });
+
+  const selectionTests = (clickNode: (user: UserEvent, node: HTMLElement) => Promise<void>) => {
+    it("does nothing when no root nodes passed", async () => {
```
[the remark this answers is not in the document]
jasdom: This was added for coverage, however was able to cover it in another test, so removed this one.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -0,0 +1,417 @@
...
+      })}
+    </>
+  );
+}
+
+describe("useSelectionHandler", () => {
+  const selectNodesStub = sinon.stub<[Array<string>, SelectionChangeType], void>();
+
+  const createHierarchyNode = (id: string, children: Array<PresentationTreeNode> = [], isExpanded: boolean = true) => {
+    return { id, isExpanded, children } as PresentationHierarchyNode;
```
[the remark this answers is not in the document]
jasdom: Replaced.

packages/hierarchies-react/src/presentation-hierarchies-react/UseSelectionHandler.ts
```diff
@@ -0,0 +1,134 @@
...
+interface FlatTreeState {
+  flatNodeList: Array<string>;
+  nodeIdToIndexMap: Map<string, number>;
+}
+
+/** @beta */
+export function useSelectionHandler(props: UseSelectionHandlerProps): UseSelectionHandlerResult {
+  const { rootNodes, selectionMode, selectNodes } = props;
+  const previousSelectionRef = useRef<string | undefined>(undefined);
+  const [state, setState] = useState<FlatTreeState>({
```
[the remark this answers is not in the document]
jasdom: No a rerender is not necessary, switched to `useRef`.

saskliutas (approved):

PR #571, opened 2024-05-15
hierarchies: Release main thread

grigasp:
Closes https://github.com/iTwin/presentation/issues/515.


grigasp:
I see flat and label grouping cases got worse - looking into those.

saskliutas (approved):

PR #582, opened 2024-05-17
Hierarchies-react cleanup API

saskliutas:
Made some improvements to the hierarchies-react package API:
* Create `HierarchyProvider` inside `useTree`. This should reduce code on consumer side and allow us to better handle filtering.
* Added ability to open filtering dialog from info node when hierarchy level limit is exceeded.
* Added ability to increase hierarchy level limit from info node. (Probably need more work on UI)

Other changes:
* Added ability to pass `restartToken` to query executor. This can be used to cancel ongoing queries before running different variant of same query,


packages/shared/api/presentation-shared.api.md
```diff
@@ -281,6 +281,8 @@ export interface ECSqlQueryExecutor {
 
 // @beta
 export interface ECSqlQueryReaderOptions {
+    // (undocumented)
+    restartToken?: string;
```
grigasp, line 285: Need a changelog for this and the change in `core-interop`

packages/hierarchies-react/src/presentation-hierarchies-react/TreeNodeRenderer.tsx
```diff
@@ -107,21 +123,69 @@ function PlaceholderNode(props: Omit<TreeNodeProps, "onExpanded">) {
...
+        <Anchor
+          underline
+          onClick={(e) => {
+            e.stopPropagation();
+            onFilterClick();
+          }}
+        >
+          additional filtering
+        </Anchor>
+        <Text>- there are more items than allowed limit of {`${limit}`}.</Text>
```
grigasp, line 167:
```suggestion
        <Text>- there are more items than allowed limit of {limit}.</Text>
```

packages/hierarchies-react/src/presentation-hierarchies-react/TreeNodeRenderer.tsx
```diff
@@ -107,21 +123,69 @@ function PlaceholderNode(props: Omit<TreeNodeProps, "onExpanded">) {
...
+      {limit < MAX_LIMIT_OVERRIDE ? (
+        <Flex flexDirection="row" gap="3xs">
+          <Anchor
+            underline
+            onClick={(e) => {
+              e.stopPropagation();
+              onOverrideLimit(getLimitOverride(limit));
+            }}
+          >
+            Increase hierarchy level limit
```
grigasp, line 178:
```suggestion
            Increase hierarchy level size limit
```

packages/hierarchies-react/src/presentation-hierarchies-react/TreeNodeRenderer.tsx
```diff
@@ -107,21 +123,69 @@ function PlaceholderNode(props: Omit<TreeNodeProps, "onExpanded">) {
...
+    </Flex>
+  );
+}
+
+function getLimitOverride(current: number) {
+  if (current < MAX_LIMIT_OVERRIDE / 2) {
+    return MAX_LIMIT_OVERRIDE / 2;
+  }
+  return MAX_LIMIT_OVERRIDE;
+}
```
grigasp, line 191: I think for simplicity we should only allow overriding the limit once. We should also mention what the override is in the `Anchor`.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -44,25 +61,34 @@ interface UseTreeResult {
    * or tree is reloading.
    */
   isLoading: boolean;
+  /**
+   * Provider used to load tree nodes.
+   */
+  hierarchyProvider?: HierarchyProvider;
```
grigasp, line 67: should this be marked `@internal`?

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -44,25 +61,34 @@ interface UseTreeResult {
...
+function useTreeInternal({
+  imodelAccess,
+  getHierarchyDefinition,
+  getFilteredPaths,
+}: UseTreeProps): UseTreeResult & { getNode: (nodeId: string) => TreeModelRootNode | TreeModelNode | undefined } {
   const [state, setState] = useState<TreeState>({
     model: { idToNode: new Map(), parentChildMap: new Map(), rootNode: { id: undefined, nodeData: undefined } },
     rootNodes: undefined,
   });
+  const [hierarchySource, setHierarchySource] = useState<{ HierarchyProvider?: HierarchyProvider; isFiltering: boolean }>({ isFiltering: false });
```
grigasp, line 91: typo?


```suggestion
  const [hierarchySource, setHierarchySource] = useState<{ hierarchyProvider?: HierarchyProvider; isFiltering: boolean }>({ isFiltering: false });
```

packages/hierarchies-react/src/test/UseTree.test.tsx
```diff
@@ -340,7 +477,20 @@ describe("useTree", () => {
...
+    const { result } = renderHook(useTree, { initialProps });
+
+    await waitFor(() => {
+      expect(result.current.rootNodes).to.have.lengthOf(1);
+      const node = result.current.rootNodes![0] as PresentationGenericInfoNode;
+      expect(node.type).to.be.eq("Unknown");
+    });
+  });
+
+  it("reloads tree when `getHierarchyDefinitionsProvider` changes", async () => {
```
grigasp, line 493:
```suggestion
  it("reloads tree when `getHierarchyDefinition` changes", async () => {
```

packages/shared/api/presentation-shared.api.md
saskliutas: Maybe it is enough to add changelog only to `shared`? `core-interop` should get `Updated dependencies:
@itwin/presentation-shared@0.2.0`

grigasp: Hm, I'm not totally sure... IMO, a dependency update means you made a `package.json` change, but doesn't mean you actually changed anything else. So I think that having an entry, saying that you're now making use of the new feature that came with the dependency update, makes sense.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -23,6 +23,10 @@ import { SelectionChangeType } from "./UseSelectionHandler";
 /** @beta */
 export interface HierarchyLevelConfiguration {
   hierarchyNode: HierarchyNode | undefined;
+  getInstanceKeysIterator: (props?: {
+    instanceFilter?: GenericInstanceFilter;
+    hierarchyLevelSizeLimit?: number | "unbounded";
+  }) => AsyncIterableIterator<InstanceKey>;
```
grigasp, line 29: Somehow this function doesn't fit into what I understand as "configuration". Could it be moved to `UseTreeResult` maybe? Or `HierarchyLevelConfiguration` needs some other name.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
saskliutas: Removed `hierarchyProvider` it was returned in order to get instance keys for filtering dialog. Added getter on `HierarchyLevelConfiguration` where all other stuff related to filtering dialog was.

apps/test-app/frontend/src/components/tree-widget/StatelessTree.tsx
```diff
@@ -88,14 +91,12 @@ function Tree({ imodel, imodelAccess, height, width }: { imodel: IModelConnectio
...
-    if (!hierarchyProvider || !filteringOptions) {
+    if (!filteringOptions) {
       return undefined;
     }
 
     return async () => {
-      const inputKeysIterator = hierarchyProvider.getNodeInstanceKeys({
-        parentNode: filteringOptions.options.hierarchyNode,
-      });
+      const inputKeysIterator = filteringOptions.options.getInstanceKeysIterator();
```
grigasp, line 99: it seems `filteringOptions` needs a new name, since it stores not only filtering options anymore

grigasp (commented):
I like the idea to hide `HierarchyProvider` completely and instead expose it's functionality through `useTree` 👍

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
saskliutas: Renamed to `HierarchyLevelDetails`. Couldn't come up with better name. 😆 I would like to keep it on `HierarchyLevelConfiguration/Details`. Putting it directly on `UseTreeResult` doesn't allow us to hide parent hierarchy node and requires consumers to pass `HierarchyNode` for which instance keys should be loaded.

apps/test-app/frontend/src/components/tree-widget/StatelessTree.tsx
saskliutas: Left `filteringOptions` name as it is but changed `filteringOptions.options` -> `filteringOptions.hierarchyDetails`. I think `filteringOptions` is good enough name in this situation because in the context of the component it stores options needed for filtering: node id and hierarchy details.

PR #583, opened 2024-05-17
Hierarchies-react: add localization support

jasdom:
Closes https://github.com/iTwin/presentation/issues/490.

### DIFF
--- a/packages/hierarchies-react/README.md
+++ b/packages/hierarchies-react/README.md
@@ -250,7 +250,12 @@ Localization can be enabled for `TreeRenderer` component and `useTree` and `useU
 Example:
 
 ```tsx
-const treeRendererLocalizedStrings = {
+const localizedStrings = {
+  // strings for the `useUnifiedSelectionTree` hook
+  unspecified: "Unspecified",
+  other: "Other",
+
+  // strings for `TreeRenderer` and `TreeNodeRenderer`
   loading: "Loading...",
   filterHierarchyLevel: "Apply filter",
   clearHierarchyLevelFilter: "Clear active filter",
@@ -261,17 +266,12 @@ const treeRendererLocalizedStrings = {
   increaseHierarchyLimitWithFiltering: "Or, <link>increase the hierarchy level size limit to {{limit}}.</link>",
 };
 
-const hierarchyProviderLocalizedStrings = {
-  unspecified: "Unspecified",
-  other: "Other",
-};
-
 function MyTreeComponent({ imodelAccess, imodelKey }: { imodelAccess: IModelAccess; imodelKey: string }) {
   const { rootNodes, ...state } = useUnifiedSelectionTree({
     sourceName: "MyTreeComponent",
     imodelKey,
     imodelAccess,
-    localizedStrings: hierarchyProviderLocalizedStrings,
+    localizedStrings,
     getHierarchyDefinition: () => ({
       defineHierarchyLevel: async () => [],
     }),
@@ -279,11 +279,11 @@ function MyTreeComponent({ imodelAccess, imodelKey }: { imodelAccess: IModelAcce
   if (!rootNodes) {
     return localizedStrings.loading;
   }
-  return <TreeRenderer rootNodes={rootNodes} localizedStrings={treeRendererLocalizedStrings} {...state} />;
+  return <TreeRenderer rootNodes={rootNodes} localizedStrings={localizedStrings} {...state} />;
 }
 ```
 
-To enable localization for a custom `TreeNodeRenderer` component, the localized strings should be supplied through `LocalizationContextProvider`:
+In case the `TreeNodeRenderer` component is used within a custom tree renderer, the tree component should supply localized strings through `LocalizationContextProvider`:
 
 ```tsx
 export function MyTreeRenderer(props: MyTreeRendererProps) {


packages/hierarchies-react/src/presentation-hierarchies-react/LocalizationContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface LocalizationContextProviderProps {
+  localization?: Localization;
+}
+
+/**
+ * Provides localized strings.
+ * @beta
+ */
+export function LocalizationContextProvider({ localization, children }: PropsWithChildren<LocalizationContextProviderProps>) {
```
grigasp, line 46: After some discussions with @saskliutas, we came up with the following plan:

1. This file contains localized strings that are only relevant to iTwinUI-based renderers, so it should be moved into the `itwinui` directory.
2. We believe it's going to be more convenient for consumers to pass localized strings as a prop to the `TreeRenderer` (which is already true), so all this context-related stuff should be `@internal`.
3. The `useTree` and `useUnifiedSelectionTree` hooks are responsible for created a hierarchy provider, which also takes localized strings - we need to ensure consumers can pass them through hook props to the provider.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -50,18 +53,20 @@ export function TreeRenderer({
...
-        <TreeNodeRenderer
-          {...nodeProps}
-          expandNode={expandNode}
-          getHierarchyLevelDetails={getHierarchyLevelDetails}
-          onFilterClick={onFilterClick}
-          onNodeClick={onNodeClick}
-          onNodeKeyDown={onNodeKeyDown}
-          getIcon={getIcon}
-        />
+        <LocalizationContextProvider localization={localization}>
```
saskliutas, line 56: No need to wrap each node in context. You can wrap `<Tree<RendererdTreeNode> /> instead.

packages/hierarchies-react/src/presentation-hierarchies-react/LocalizationContext.tsx
```diff
@@ -0,0 +1,59 @@
...
+  increaseHierarchyLimit: string;
+  increaseHierarchyLimitWithFiltering: string;
+}
+
+/** @internal */
+export interface LocalizationContext {
+  localization: Localization;
+}
+
+const defaultLocalization: Localization = {
```
saskliutas, line 25: For someone localizing these string it could be inconvenient to find default values. Maybe also add them as a comment for each attribute of `Localization` type?

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
jasdom: Right, it was not supposed to wrap each node, fixed.

packages/hierarchies-react/src/presentation-hierarchies-react/LocalizationContext.tsx
jasdom: 1. Moved the context into `itwinui` directory.
2. Left the context as `@beta` as it should be used to provide localization when `TreeNodeRenderer` is used individually. Added `README` section for it.
3. Added the ability to pass localized string to hierarchy provider.

packages/hierarchies-react/README.md
```diff
@@ -242,3 +242,70 @@ function MyTreeComponentInternal({ imodelAccess, imodelKey }: { imodelAccess: IM
...
+  loading: "Loading...",
+  filterHierarchyLevel: "Apply filter",
+  clearHierarchyLevelFilter: "Clear active filter",
+  noFilteredChildren: "No child nodes match current filter",
+  resultLimitExceeded: "There are more items than allowed limit of {{limit}}.",
+  resultLimitExceededWithFiltering: "Please provide <link>additional filtering</link> - there are more items than allowed limit of {{limit}}.",
+  increaseHierarchyLimit: "<link>Increase the hierarchy level size limit to {{limit}}.</link>",
+  increaseHierarchyLimitWithFiltering: "Or, <link>increase the hierarchy level size limit to {{limit}}.</link>",
+  unspecified: "Unspecified",
+  other: "Other",
```
saskliutas, line 263: These two used by `useTree` should be separated from the other ones. `useTree` can be used without `TreeRenderer` so in that case it does not make sense to require all these additional strings that are used only by renderer.

packages/hierarchies-react/README.md
```diff
@@ -242,3 +242,70 @@ function MyTreeComponentInternal({ imodelAccess, imodelKey }: { imodelAccess: IM
...
+    }),
+  });
+  if (!rootNodes) {
+    return "Loading...";
+  }
+  return <TreeRenderer rootNodes={rootNodes} localizedStrings={localizedStrings} {...state} />;
+}
+```
+
+To enable localization for the `TreeNodeRenderer` component, the localized strings should be supplied through `LocalizationContextProvider`:
```
saskliutas, line 283:
```suggestion
To enable localization for the custom `TreeNodeRenderer` component, the localized strings should be supplied through `LocalizationContextProvider`:
```

saskliutas (changes requested):

packages/hierarchies-react/README.md
jasdom: They are separate, I just merged the object into one in the `README`. Split into two to be more clear.

packages/hierarchies-react/README.md
```diff
@@ -242,3 +242,70 @@ function MyTreeComponentInternal({ imodelAccess, imodelKey }: { imodelAccess: IM
   return <TreeRenderer rootNodes={rootNodes} {...state} />;
 }
 ```
+
+## Localization
+
+Localization can be enabled for `TreeRenderer` component and `useTree` and `useUnifiedSelectionTree` hooks by providing an object with localized strings that will be used instead of the default english ones.
```
grigasp, line 248:
```suggestion
Localization can be enabled for `TreeRenderer` component and `useTree` and `useUnifiedSelectionTree` hooks by providing an object with localized strings that will be used instead of the default English ones.
```

packages/hierarchies-react/README.md
```diff
@@ -242,3 +242,70 @@ function MyTreeComponentInternal({ imodelAccess, imodelKey }: { imodelAccess: IM
...
+    sourceName: "MyTreeComponent",
+    imodelKey,
+    imodelAccess,
+    localizedStrings,
+    getHierarchyDefinition: () => ({
+      defineHierarchyLevel: async () => [],
+    }),
+  });
+  if (!rootNodes) {
+    return "Loading...";
```
grigasp, line 280:
```suggestion
    return localizedStrings.loading;
```

packages/hierarchies-react/README.md
grigasp: Not sure this is correct. It's not the `TreeNodeRenderer` that has to be custom for the requirement to use `LocalizationContextProvider`, but the `TreeRenderer`. I suggest this:

```suggestion
In case the `TreeNodeRenderer` component is used within a custom tree renderer, the tree component should supply localized strings through `LocalizationContextProvider`:
```

packages/hierarchies-react/README.md
```diff
@@ -242,3 +242,70 @@ function MyTreeComponentInternal({ imodelAccess, imodelKey }: { imodelAccess: IM
...
+  loading: "Loading...",
+  filterHierarchyLevel: "Apply filter",
+  clearHierarchyLevelFilter: "Clear active filter",
+  noFilteredChildren: "No child nodes match current filter",
+  resultLimitExceeded: "There are more items than allowed limit of {{limit}}.",
+  resultLimitExceededWithFiltering: "Please provide <link>additional filtering</link> - there are more items than allowed limit of {{limit}}.",
+  increaseHierarchyLimit: "<link>Increase the hierarchy level size limit to {{limit}}.</link>",
+  increaseHierarchyLimitWithFiltering: "Or, <link>increase the hierarchy level size limit to {{limit}}.</link>",
+};
+
```
grigasp, line 299: The above example already shows what `localizedStrings` looks like, so I think we can omit it here for the sake of simplicity.

```suggestion
```

packages/hierarchies-react/README.md
```diff
@@ -242,3 +242,70 @@ function MyTreeComponentInternal({ imodelAccess, imodelKey }: { imodelAccess: IM
...
+export function TreeRenderer({ ...props }: TreeRendererProps) {
```
grigasp, line 300: Let's rename to make it clear it's something that consumers would be creating.

```suggestion
export function MyTreeRenderer(props: MyTreeRendererProps) {
```

packages/hierarchies-react/README.md
grigasp: I personally liked the non-split version better. Maybe adding a couple of comments would've been a better change?
```ts
const localizedStrings = {
  // strings for the `useTree` hook
  ...

  // strings for `TreeRenderer` and `TreeNodeRenderer`
  ...
};
```

jasdom: Reverted the change and left comments.

packages/hierarchies-react/README.md
jasdom: Updated based on suggestion.

PR #586, opened 2024-05-21
Avoid clearing tree model when load is cancelled

saskliutas:
`reduce` operator invokes `next` when inner observable completes. This means that if `takeUntil` causes unsubscription reduce will emit seed or accumulated result. Moved `takeUntil` after reduce to make sure that only `complete` is invoked when loading is cancelled.


grigasp (commented):
Could we have a test for this fix?

grigasp (approved):

PR #598, opened 2024-05-27
`models-tree`: Pre-load subject-models hierarchy to improve performance of loading subjects hierarchy

grigasp:
Part of https://github.com/iTwin/presentation/issues/597.

image

There are still a few cases that need to be looked at, but this provides a noticeable improvement for large subjects' hierarchy cases.

Issue #597, opened 2024-05-27
Stateless models tree: Investigate cases where creating the hierarchy is slower compared to stateful implementation


grigasp:

grigasp:
Pushed a couple of changes to improve performance for creating the subject's / models hierarchy. Looks like another bottleneck our class hierarchy inspector, which requires the whole schema to be loaded - filed a separate backlog item to look into that: https://github.com/iTwin/presentation/issues/601.

PR #599, opened 2024-05-27
`models-tree`: Pre-determine subjects' children flag

grigasp:
Part of https://github.com/iTwin/presentation/issues/597.

Often, to determine whether a Subject node needs to be displayed or not, we have to run a bunch of queries to see if its branch contains at least one Model with elements. However, with https://github.com/iTwin/presentation/pull/598, we already have that information, so here we just make use of it.

image

image

There're still a few cases that need to be look at.

### DIFF
--- a/packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
+++ b/packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
@@ -565,29 +565,27 @@ class SubjectModelIdsCache {
   }
 
   public async getParentSubjectIds(): Promise<Id64String[]> {
-    await this.initCache();
-    if (!this._parentSubjectIds) {
-      this._parentSubjectIds = (async () => {
-        const hasChildModels = (subjectId: Id64String) => {
-          if ((this._subjectModels!.get(subjectId)?.size ?? 0) > 0) {
-            return true;
-          }
-          const childSubjectIds = this._subjectsHierarchy!.get(subjectId);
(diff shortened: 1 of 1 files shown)


packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
```diff
@@ -559,6 +564,33 @@ class SubjectModelIdsCache {
       });
   }
 
+  public async getParentSubjectIds(): Promise<Id64String[]> {
+    await this.initCache();
+    if (!this._parentSubjectIds) {
+      this._parentSubjectIds = (async () => {
```
Yato333, line 570: `this._parentSubjectIds` doesn't have to be a promise. Also I think you can apply a `??=` operator here.

grigasp: > `this._parentSubjectIds` doesn't have to be a promise.

why?

Yato333: There are no asynchronous operations needed to create it, so it can be stored as an array instead of a promise. Am I missing something?

grigasp: Gotcha. The `initCache` should've been inside the parent subjects function. Made the change together with the suggested use of `??=` operator.

Yato333 (approved):

PR #600, opened 2024-05-28
Hierarchies-react: Add performance metrics reporting

saskliutas:
Added API for reporting performance metrics

### DIFF
--- a/.changeset/weak-spies-hide.md
+++ b/.changeset/weak-spies-hide.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": minor
+---
+
+Added API for performance metrics reporting.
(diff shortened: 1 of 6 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -283,3 +287,27 @@ class NoopTreeLoader implements ITreeLoader {
     return EMPTY;
   }
 }
+
+class TimeTracker {
+  private _start: number;
+  private _stopped: boolean = false;
+
+  constructor(private _onFinish: (time: number) => void) {
+    this._start = performance.now();
```
grigasp, line 296: AFAIK, `Date.now()` is more performant. And we don't need the precision advantage that `performance.now()` provides.

.changeset/weak-spies-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": minor
+---
+
+Added API for performance metrics reporting.
```
grigasp, line 5: Please be more specific about what the API is and include an example.

grigasp (approved):

PR #605, opened 2024-05-29
Hierarchies-react: `onHierarchyLimitExceeded` callback

jasdom:
Added callback needed to track when hierarchy limit was exceeded to add feature reporting.

### DIFF
--- a/.changeset/friendly-bananas-smoke.md
+++ b/.changeset/friendly-bananas-smoke.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": minor
+---
+
+Added `onHierarchyLimitExceeded` callback for tracking when hierarchy level exceeds the limit.
--- a/packages/hierarchies-react/src/test/internal/TreeLoader.test.ts
+++ b/packages/hierarchies-react/src/test/internal/TreeLoader.test.ts
@@ -17,8 +17,8 @@ describe("TreeLoader", () => {
     getNodes: sinon.stub<Parameters<HierarchyProvider["getNodes"]>, ReturnType<HierarchyProvider["getNodes"]>>(),
   };
 
-  function createLoader() {
-    return new TreeLoader(hierarchyProvider as unknown as HierarchyProvider);
+  function createLoader(onHierarchyLimitExceeded = () => {}) {
+    return new TreeLoader(hierarchyProvider as unknown as HierarchyProvider, onHierarchyLimitExceeded);
   }
 
   beforeEach(() => {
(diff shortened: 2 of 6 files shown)


packages/hierarchies-react/src/test/internal/TreeLoader.test.ts
```diff
@@ -17,8 +17,8 @@ describe("TreeLoader", () => {
     getNodes: sinon.stub<Parameters<HierarchyProvider["getNodes"]>, ReturnType<HierarchyProvider["getNodes"]>>(),
   };
 
-  function createLoader() {
-    return new TreeLoader(hierarchyProvider as unknown as HierarchyProvider);
+  function createLoader(onHierarchyLimitExceeded = () => {}) {
+    return new TreeLoader(hierarchyProvider as unknown as HierarchyProvider, onHierarchyLimitExceeded);
```
saskliutas, line 21:
```suggestion
  const onHierarchyLimitExceededStub = sinon.stub();

  function createLoader() {
    return new TreeLoader(hierarchyProvider as unknown as HierarchyProvider, onHierarchyLimitExceededStub);
```

.changeset/friendly-bananas-smoke.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": minor
+---
+
+Added `onHierarchyLimitExceeded` callback for tracking when hierarchy level exceeds the limit.
```
grigasp, line 5: Please mention where it was added. Ideally, an example should also be included.

packages/hierarchies-react/src/test/internal/TreeLoader.test.ts
jasdom: Changed.

.changeset/friendly-bananas-smoke.md
jasdom: Updated.

.changeset/friendly-bananas-smoke.md
```diff
@@ -0,0 +1,19 @@
...
+
+```typescript
+import { TreeRenderer, useTree } from "@itwin/presentation-hierarchies-react";
+
+function MyTree(props: MyTreeProps) {
+  const state = useTree({
+    ...props,
+    onHierarchyLimitExceeded: () => {
+      console.log(`Hierarchy limit exceeded`);
+    }
```
grigasp, line 15: Looking at this example, it seems it would make sense to get the parent node whose hierarchy level exceeded the limit. Maybe even additional config like instance filter and hierarchy level size limit too?

jasdom: I agree it is worth adding the extra information. Changed.

grigasp: Did you consider passing a props object rather than 3 different arguments? That would allow us to add an additional non-nullable parameter in the future without breaking the API.

jasdom: No, did not consider that we might want to add other information but it definitely does not hurt to future-proof the callback. Updated.

grigasp (approved):

PR #607, opened 2024-05-29
Hierarchies-react: Reload tree on data change

saskliutas:
Updated `reloadTree` options to allow to notify `HierarchyProvider` about data changes. This makes sure that `HierarchyProvider` has cleared caches and pulls new data when tree is reloading.

### DIFF
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
@@ -104,7 +104,7 @@ interface UseTreeResult {
    * or tree is reloading.
    */
   isLoading: boolean;
-  reloadTree: (options?: { discardState?: boolean }) => void;
+  reloadTree: (options?: { discardState?: boolean; dataMightHaveChanged?: boolean }) => void;
   expandNode: (nodeId: string, isExpanded: boolean) => void;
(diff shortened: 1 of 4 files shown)


packages/hierarchies-react/src/test/UseTree.test.tsx
```diff
@@ -539,6 +543,39 @@ describe("useTree", () => {
     });
   });
 
+  it("notifies hierarchy provider about changed date when `reloadTree` is called with `dataMightHaveChanged`", async () => {
```
grigasp, line 546:
```suggestion
  it("notifies hierarchy provider about changed data source when `reloadTree` is called with `dataMightHaveChanged`", async () => {
```

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -104,7 +104,7 @@ interface UseTreeResult {
    * or tree is reloading.
    */
   isLoading: boolean;
-  reloadTree: (options?: { discardState?: boolean }) => void;
+  reloadTree: (options?: { discardState?: boolean; dataMightHaveChanged?: boolean }) => void;
```
grigasp, line 107: what do you think about changing the attribute name to `dataSourceChanged`?

grigasp (approved):

PR #613, opened 2024-06-03
`testing`: When creating a test iModel, limit its file name length to fit into file system limits

grigasp:


packages/testing/src/presentation-testing/InternalUtils.ts
```diff
@@ -27,19 +27,36 @@ export function setTestOutputDir(directoryPath: string | undefined) {
...
 export function createFileNameFromString(str: string) {
   return sanitize(str.replace(/[ ]+/g, "-").replaceAll("`", "").replaceAll("'", "")).toLocaleLowerCase();
 }
+
+/** @internal */
+export function limitFilePathLength(filePath: string) {
+  const { dir, name, ext } = path.parse(filePath);
+
+  const allowedFileNameLength = 260 - 12 - 1 - (dir.length + 1) - ext.length;
+  if (allowedFileNameLength <= 0) {
```
Yato333, line 52: It has to be at least 3 symbols since shortened name has `...` in it

grigasp: should be okay now

Yato333 (approved):

PR #615, opened 2024-06-03
Hierarchies-react: pass node to `onNodeClick` instead of node ID

jasdom:
Changes needed for https://github.com/iTwin/viewer-components-react/issues/867. Also memoized the `onNodeClick` and `onNodeKeyDown` as `TreeRenderer` contains a hook that depends on them.

### DIFF
--- a/.changeset/orange-clocks-sin.md
+++ b/.changeset/orange-clocks-sin.md
@@ -1,5 +1,24 @@
 ---
-"@itwin/presentation-hierarchies-react": patch
+"@itwin/presentation-hierarchies-react": minor
 ---
 
-Modified `onNodeClick` and `onNodeKeyDown` callbacks to pass entire node instead of id in `TreeNodeRenderer`.
+Modified `onNodeClick` and `onNodeKeyDown` callbacks to pass entire node instead of id in `TreeNodeRenderer`. This change breaks the existing API, so the consuming code needs to be adjusted to use `node.id` instead of `nodeId`:
+
+```typescript
+import { TreeNodeRenderer } from "@itwin/presentation-hierarchies-react";
+
+export function MyTreeNodeRenderer(props) {
+  return (
+    <TreeNodeRenderer
+      {...props}
+      // Previous implementation
+      onNodeClick={(nodeId, isSelected, event) => someFunc(nodeId)}
+      onNodeKeyDown={(nodeId, isSelected, event) => someFunc(nodeId)}
+
+      // After the change
+      onNodeClick={(node, isSelected, event) => someFunc(node.id)}
+      onNodeKeyDown={(node, isSelected, event) => someFunc(node.id)}
+    />
+  );
+}
+```
--- a/.changeset/orange-clocks-sin.md
+++ b/.changeset/orange-clocks-sin.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Modified `onNodeClick` and `onNodeKeyDown` callbacks to pass entire node instead of id in `TreeNodeRenderer`.
--- a/.changeset/orange-clocks-sin.md
+++ b/.changeset/orange-clocks-sin.md
@@ -2,7 +2,7 @@
 "@itwin/presentation-hierarchies-react": minor
 ---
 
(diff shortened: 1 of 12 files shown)


.changeset/orange-clocks-sin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
```
saskliutas, line 2: This should be a minor change since API was broken. Also could you mention that this breaks existing API and how code should be adjusted?

grigasp (commented):
We have other APIs (e.g. see `UseTreeResult`), which work with node ids rather than nodes. So this PR kind of moves away from that pattern - why? How likely is it that we'll need to modify those other APIs to work with nodes rather than ids?

jasdom:
> We have other APIs (e.g. see `UseTreeResult`), which work with node ids rather than nodes. So this PR kind of moves away from that pattern - why? How likely is it that we'll need to modify those other APIs to work with nodes rather than ids?

When only the id is passed the consumer has to either find a way to retrieve the data for the node (for example if they need access to `extendedData`) or wrap the callback manually to include the node which is inconvenient. I personally would be for passing nodes instead of ids in other APIs as well as it is more future-proof and easier to use. We also aready pass the entire node to `getIcon` and `getSublabel` so that pattern was already broken.

saskliutas:
> > We have other APIs (e.g. see `UseTreeResult`), which work with node ids rather than nodes. So this PR kind of moves away from that pattern - why? How likely is it that we'll need to modify those other APIs to work with nodes rather than ids?
> 
> When only the id is passed the consumer has to either find a way to retrieve the data for the node (for example if they need access to `extendedData`) or wrap the callback manually to include the node which is inconvenient. I personally would be for passing nodes instead of ids in other APIs as well as it is more future-proof and easier to use. We also aready pass the entire node to `getIcon` and `getSublabel` so that pattern was already broken.

I think all callbacks that are passed to `TreeRenderer`/`TreeNodeRenderer` (related to rendering nodes) should get node instead of `nodeId`. I don't think there is need to change functions retuned by `useTree` to take nodes instead of ids. They usually will be called from UI component event handlers so consumer should have access to the node before calling them.

grigasp:
> > > We have other APIs (e.g. see `UseTreeResult`), which work with node ids rather than nodes. So this PR kind of moves away from that pattern - why? How likely is it that we'll need to modify those other APIs to work with nodes rather than ids?
> > 
> > 
> > When only the id is passed the consumer has to either find a way to retrieve the data for the node (for example if they need access to `extendedData`) or wrap the callback manually to include the node which is inconvenient. I personally would be for passing nodes instead of ids in other APIs as well as it is more future-proof and easier to use. We also aready pass the entire node to `getIcon` and `getSublabel` so that pattern was already broken.
> 
> I think all callbacks that are passed to `TreeRenderer`/`TreeNodeRenderer` (related to rendering nodes) should get node instead of `nodeId`. I don't think there is need to change functions retuned by `useTree` to take nodes instead of ids. They usually will be called from UI component event handlers so consumer should have access to the node before calling them.

Sounds good. @jasdom, please also change `TreeNodeRendererOwnProps.onFilterClick` to take node instead of id.

.changeset/orange-clocks-sin.md
jasdom: Changed to minor and provided more information.

apps/test-app/frontend/src/components/tree-widget/StatelessTree.tsx
```diff
@@ -89,22 +89,17 @@ function Tree({ imodel, imodelAccess, height, width }: { imodel: IModelConnectio
...
-    (nodeId: string | undefined) => {
-      const hierarchyLevelDetails = getHierarchyLevelDetails(nodeId);
-      setFilteringOptions(hierarchyLevelDetails ? { nodeId, hierarchyLevelDetails } : undefined);
-    },
-    [getHierarchyLevelDetails],
-  );
+  const [filteringOptions, setFilteringOptions] = useState<HierarchyLevelDetails>();
+  const onFilterClick = useCallback((hierarchyLevelDetails: HierarchyLevelDetails) => {
+    setFilteringOptions(hierarchyLevelDetails);
+  }, []);
```
saskliutas, line 95: I think this is not needed any more. You can pass `setFilteringOptions` directly.

jasdom: Right, the `nodeId` is no longer stored. Fixed.

saskliutas (approved):

grigasp (approved):
LGTM

PR #620, opened 2024-06-04
hierarchies: Expose node filtering props

Yato333:
These properties are already exposed in the resulting nodes but the type doesn't reflect it.
Related to: https://github.com/iTwin/viewer-components-react/issues/868

### DIFF
--- a/packages/hierarchies/src/hierarchies/HierarchyNode.ts
+++ b/packages/hierarchies/src/hierarchies/HierarchyNode.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { ConcatenatedValue, InstanceKey, OmitOverUnion, PrimitiveValue } from "@itwin/presentation-shared";
+import { HierarchyNodeIdentifiersPath } from "./HierarchyNodeIdentifier";
 import {
   ClassGroupingNodeKey,
   GroupingNodeKey,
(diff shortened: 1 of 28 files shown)


packages/hierarchies/src/hierarchies/HierarchyNode.ts
```diff
@@ -340,3 +342,19 @@ export type ParsedHierarchyNode<TBase = ParsedCustomHierarchyNode | ParsedInstan
...
+  isFilterTarget?: boolean;
+  hasFilterTargetAncestor?: boolean;
+  filteredChildrenIdentifierPaths?: HierarchyNodeIdentifiersPath[];
+}
+
+/**
+ * A `HierarchyNode` that has been fully processed and only has information, needed for rendering.
+ * @beta
+ */
+export type FinalizedNode = HierarchyNode & FilteredNodeProps;
```
grigasp, line 360: I'm not a big fan of this... `HierarchyNode` is already the type that means a fully finalized node, so not sure introducing another one makes much sense. What do you think about adding those props to `BaseHierarchyNode`?

```ts
interface BaseHierarchyNode {
  // ...

  filtering?: {
    isFilterTarget?: boolean;
    filteredChildrenIdentifierPaths?: HierarchyNodeIdentifiersPath[];
  };
}

```

Yato333: Made suggested change

grigasp (approved):

PR #638, opened 2024-06-17
Components: Fix UniqueValuesSelector not always loading values when searching.

jasdom:
Fixes needed for https://github.com/iTwin/viewer-components-react/issues/907.

### DIFF
--- a/packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
+++ b/packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
@@ -725,7 +725,12 @@ describe("UniquePropertyValuesSelector", () => {
       const menuSelector = await waitFor(() => getByText("unique-values-property-editor.select-values"));
       await user.click(menuSelector);
 
-      /// reset scroll height to enable loading of pages
+      // wait for first page to be loaded
+      await waitFor(() => {
+        expect(getDistinctValuesIteratorStub).to.be.calledOnce;
+      });
+
+      // reset scroll height to enable loading of pages
       sinon.stub(Element.prototype, "scrollHeight").get(() => 0);
 
       // type in search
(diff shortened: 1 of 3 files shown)


packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
```diff
@@ -772,6 +773,93 @@ describe("UniquePropertyValuesSelector", () => {
 
       // expect "clear" button to not be shown
       await waitFor(() => expect(container.querySelector(".clear-indicator")).to.be.null);
+
+      const closeFilteringDialogButton = container.querySelector(".presentation-instance-filter-dialog-close-button");
+      await user.click(closeFilteringDialogButton!);
```
saskliutas, line 778: Is this needed?

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
```diff
@@ -772,6 +773,93 @@ describe("UniquePropertyValuesSelector", () => {
...
+        { displayValue: undefined, groupedRawValues: [] },
+      ];
+      for (let i = 2; i < UNIQUE_PROPERTY_VALUES_BATCH_SIZE; i++) {
+        const name = `SkippedValue${i}`;
+        pageItems.push({ displayValue: name, groupedRawValues: [name] });
+      }
+      getDistinctValuesIteratorStub.onCall(0).resolves({
+        total: 0,
+        items: createAsyncIterator([]),
+      });
```
saskliutas, line 834: Are the real cases where first request returns 0 items, and second returns something?

saskliutas (changes requested):

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
jasdom: No, I missed this when removing the others, however it turns out that without this other tests fail. I fixed the setup in this test because the new changes were causing it to load infinite number of pages and that resolved all the issues. Not sure how or why it was causing other tests to fail.

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
jasdom: No, the first call was only used to not cause more pages to be loaded. I changed it to return a single value instead. It is still not accurate to real cases as an entire page would be loaded, however that causes flakiness in tests because for some reason `AsyncPaginate` sometimes believes the menu is being continuously scrolled and starts loading all the pages prematurely.

saskliutas (approved):

PR #639, opened 2024-06-18
Hierarchies-react: Added ability to retry loading hierarchy level

saskliutas:
Required for https://github.com/iTwin/viewer-components-react/issues/888


packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -94,6 +94,17 @@ interface UseTreeProps {
...
+interface ReloadTreeOptions {
+  /** Specifies whether tree state should be discarded when reloading tree or kept. */
+  discardState?: boolean;
+  /** Specifies that data source changed and caches should be cleared before reloading tree. */
+  dataSourceChanged?: boolean;
+  /** Specifies whether existing nodes should be removed prior reloading or should be kept until reload is completed. */
+  force?: boolean;
+  /** Specifies parent node under which sub tree should be reloaded instead of reloading whole tree. */
+  parentNodeId?: string;
+}
```
grigasp, line 106: A couple of things I don't like about this:

1. It seems `dataSourceChanged` doesn't make much sense if you also pass `parentNodeId`. They should be mutually exclusive? `{ dataSourceChanged?: boolean } | { parentNodeId: string }`?

2. Not a fan of `force`. It sounds like it does pretty much what `discardState` does, but additionally clears subtree state *before* reload and ignores cache. So maybe instead of these two, we could have a prop like this? `{ state?: "keep" | "discard" | "reset" }`, where:
   - `keep` is the default value is means `{ discardState: false, force: false }`.
   - `discard` means we keep the subtree state until we get the fresh data, but throw it away when the new subtree is added to the model, matches `{ discardState: true, force: false }`.
   - `clear` means we discard the subtree state as soon as `reloadTree` is called and we ignore cache, matches `{ discardState: true, force: true }`.

saskliutas: Cleaned up options.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -94,6 +94,28 @@ interface UseTreeProps {
...
 }
 
+interface ReloadTreeCommonOptions {
+  /**
+   * Specifies how current tree state should be handled:
+   * - `keep` - try to keep current tree state (expanded/collapsed nodes, instance filters, etc.).
+   * - `discard` - do not try to keep current tree state. Tree model will be update after nodes are reloaded.
+   * - `reset` - remove subtree from the model before reloading and reload nodes ignoring cache.
+   */
+  state?: "keep" | "discard" | "reset";
```
grigasp, line 104:
```suggestion
  /**
   * Specifies how current tree state should be handled:
   * - `keep` - try to keep current tree state (expanded/collapsed nodes, instance filters, etc.).
   * - `discard` - do not try to keep current tree state. Tree model will be update after nodes are reloaded.
   * - `reset` - remove subtree from the model before reloading and reload nodes ignoring cache.
   * 
   * Defaults to `"keep"`.
   */
  state?: "keep" | "discard" | "reset";
```

grigasp (approved):

PR #640, opened 2024-06-20
Hierarchies-react: fix keyboard navigation issues

jasdom:
Reverted changes made with https://github.com/iTwin/presentation/pull/615/files as passing `onKeyDown` to `TreeNode` does not work as expected and causes issues with keyboard navigation as well as visual bugs.


packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
```diff
@@ -933,13 +933,16 @@ describe("UniquePropertyValuesSelector", () => {
...
-      expect(getDistinctValuesIteratorStub).to.be.calledThrice;
+      await waitFor(
+        () => {
+          expect(getDistinctValuesIteratorStub).to.be.calledThrice;
+          expect(queryByText("SearchedValue1")).to.not.be.null;
+          expect(queryByText("SearchedValue2")).to.not.be.null;
+          expect(queryByText("SearchedValue3")).to.not.be.null;
+          expect(queryAllByText(/SkippedValue/)).to.be.empty;
+        },
+        { timeout: 2000 },
```
saskliutas, line 944: Why is this needed?

jasdom: This test was occasionally failing. Did some debugging and it seems like the pages are requested correctly but sometimes it timeouts before the last page loaded.

saskliutas (approved):

PR #642, opened 2024-06-20
`hierarchies`: Add `onlyIfNotHandled` flag to class based hierarchy definitions

### LINKED ISSUE #596: `hierarchies`: Implement `onlyIfNotHandled` attribute support for `ClassBasedHierarchyDefinition`
`createClassBasedHierarchyDefinition` functions allows creating hierarchy definitions in a declarative way, but it's inconvenient to define fallback rules, e.g.:

1. For parent node class `bis.GeometricModel3d`, use definition `X`.
2. For parent node class `bis.GeometricModel2d`, use definition `Y`.
3. For parent node class `bis.Model`, only if it's neither 3d not 2d, use definition `Z`.

It seems the most convenient way to define this would be using an attribute similar to [`onlyIfNotHandled` in presentation rules](https://www.itwinjs.org/presentation/hierarchies/childnoderule/#attribute-onlyifnothandled).

Yato333:
Closes #596


.changeset/mean-zoos-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies": patch
+---
+
+Add `onlyIfNotHandled` flag to class based hierarchy definitions which allows to skip one if previous definitions are applied to a node.
```
grigasp, line 5: This should be a `minor`, since it's a public API addition. Also, I'd like the comment to mention which public API was changed (`createClassBasedHierarchyDefinition`), how and why, ideally with an example.

Yato333: Improved docs. Also when writing them, I've noticed that this flag is useless with custom definitions, since it never makes sense to have multiple definitions for the same node key.

grigasp (approved):

PR #643, opened 2024-06-20
Include APIs that are unexported but still public to the `extract-api` report

### LINKED ISSUE #545: Tooling to detect breaking changes
Our new strategy regarding API publishing is to avoid making various props and return types public by not exposing them through the barrel file. This means that the changes to the following example APIs may break public API:
- A non-exported `type` / `interface` that is exposed (directly or indirectly) through a public API.
- An exported `@internal` `type` / `interface` that is exposed (directly or indirectly) through a public API.
- A `type` / `interface` of another package that we expose (directly or indirectly) through our public API.

In all of the above cases the type in question is not exported through a barrel file, which means changes to it are not detected by the API extractor tool. Because of this, it's very easy to miss a breaking change, for example - by adding some required prop to a non-exported props interface that's exposed through a public API.

Need some tooling to detect such changes.

Yato333:
Closes #545 
Unexported APIs will now require a release tag which should match the release tag of exported APIs


apps/build-tools/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "clean": "rimraf lib build"
+  },
+  "dependencies": {
+    "@itwin/build-tools": "^4.7.1",
+    "@microsoft/api-extractor": "~7.47",
+    "yargs": "^17.7.2"
+  },
+  "devDependencies": {
+    "presentation-build-tools": "workspace:^"
+  }
```
grigasp, line 19: dependency on itself?

apps/build-tools/bin/extract-api.cjs
```diff
@@ -0,0 +1,119 @@
...
+const entryPointFileName = argv.entry;
+const missingTagsLevel = argv.missingTags ?? "error";
+const incompatibleTagsLevel = argv.ignoreIncompatibleTags ?? "error";
+
+const apiReportFolder = argv.apiReportFolder ?? "./api";
+const apiReportTempFolder = argv.apiReportTempFolder ?? "./api/temp";
+
+const config = {
+  $schema: "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+  projectFolder: "../../",
```
grigasp, line 25: Shouldn't project folder be `./`? This script assumes that `cwd` is package root, right?

apps/build-tools/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "description": "Presentation build tools",
+  "private": true,
+  "bin": {
+    "extract-api": "bin/extract-api.cjs"
+  },
+  "scripts": {
+    "clean": "rimraf lib build"
+  },
+  "dependencies": {
+    "@itwin/build-tools": "^4.7.1",
```
grigasp, line 13: we aren't using `@itwin/build-tools`, are we?

apps/build-tools/bin/extract-api.cjs
Yato333: This config is actually saved as a file in project's `<package_root>/lib/cjs` folder and the path is resolved from that.

apps/build-tools/package.json
Yato333: I am now only using it for the `simpleSpawn` utility (`extract-api.cjs`). Maybe it's not worth using that

grigasp: Missed that. I don't mind having the dependency, as long as it's used - let's keep it.

apps/build-tools/package.json
Yato333: Removed

packages/unified-selection/api/unified-selection.api.md
```diff
@@ -36,9 +82,27 @@ export interface CustomSelectable {
...
+interface ElementSelectionScopeProps {
+    ancestorLevel?: number;
+    id: "element" | "functional";
+}
+
 // @beta
 export function enableUnifiedSelectionSyncWithIModel(props: EnableUnifiedSelectionSyncWithIModelProps): () => void;
 
+// @internal
+interface EnableUnifiedSelectionSyncWithIModelProps {
```
grigasp, line 95: Here we have an `@internal` API exposed through a `@beta` one. Two questions:
1. Why doesn't `check-internal` script complain about `@internal` being exposed here in API report?
2. Why doesn't `extract-api` script complain about public API exposing internal one?

Yato333: > Why doesn't check-internal script complain about @internal being exposed here in API report?

I couldn't figure out a simple way to exclude this check only for APIs, exported from internal fields so I removed it for APIs that are not exported through barrel (no `export` keyword in API.md). This was needed because we have at least one instance of `@internal` interface member (in presentation-components
`ViewWithUnifiedSelectionProps.selectionHandler`). Judging from this comment in `check-internal` script I figured that this is something that we have to account for:
```
// only look for non-indented tags which target root level APIs - we still want to allow things
// like @internal class functions and interface attributes
```
In this particular case it looks like I missed this after disabling the internal check.

> Why doesn't extract-api script complain about public API exposing internal one?

It checks if release tags are "compatible" and not necessarily equal. This is supposed to guard against cases where creation of `public` objects would require to use `beta`, `alpha` or `internal` APIs. See [docs about this](https://api-extractor.com/pages/messages/ae-incompatible-release-tags/). Although, I am not sure why it allows for this. Maybe it handles unexported types in a different way, or maybe this is a bug. Unexported APIs is something that the authors are very reluctant to support or document about (they call this "forgotten exports").

grigasp: Added [another backlog item](https://github.com/iTwin/presentation/issues/652) to fix occurrences of `@internal` in the public API.

grigasp (approved):

PR #649, opened 2024-06-25
Hierarchies-react: switch focus to apply filter button when filter is cleared

jasdom:
Switch focus to "Apply filter" button when "Clear active filter" button is clicked.

### DIFF
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
@@ -5,7 +5,7 @@
 
 import "./TreeNodeRenderer.css";
 import cx from "classnames";
-import { ComponentPropsWithoutRef, ReactElement, useState } from "react";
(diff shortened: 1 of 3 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -57,6 +57,7 @@ export function TreeNodeRenderer({
   ...treeNodeProps
 }: TreeNodeRendererProps) {
   const { localizedStrings } = useLocalizationContext();
+  const [applyFilterButtonRef, setApplyFilterButtonRef] = useState<HTMLElement | null>(null);
```
saskliutas, line 60: Why not `useRef`?

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
jasdom: Changed to `useRef`.

saskliutas (approved):

PR #662, opened 2024-07-03
`components`: API promotions

grigasp:
Part of https://github.com/iTwin/itwinjs-backlog/issues/1103.

@saskliutas, can you please have a look at  `packages\components\src\presentation-components\properties\editors\NavigationPropertyEditorContext.ts`? `navigationPropertyEditorContext` is exposed as public, although `createContext` results are generally internal elsewhere - I wasn't sure we want to make it `public` or `internal`.


saskliutas (approved):
I think `navigationPropertyEditorContext` should be internal. There is no reason it should be used outside NavigationPropertEditor.

grigasp:
> I think `navigationPropertyEditorContext` should be internal. There is no reason it should be used outside NavigationPropertEditor.

Made the change, please review.

packages/components/src/presentation-components/properties/editors/NavigationPropertyEditorContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+export function NavigationPropertyEditorContextProvider({ children, ...props }: PropsWithChildren<NavigationPropertyEditorContextProviderProps>) {
+  return <navigationPropertyEditorContext.Provider value={props}>{children}</navigationPropertyEditorContext.Provider>;
+}
+
+/**
+ * Returns context provided by [[NavigationPropertyEditorContextProvider]]. The context is required for rendering [[NavigationPropertyEditor]] with editable
+ * navigation property values.
+ * @public
+ */
+export function useNavigationPropertyEditorContext() {
```
saskliutas, line 53: I don't think this needs to be exported. It should not be used outside NavigationPropertyEditor.

saskliutas (changes requested):

packages/components/src/presentation-components/properties/editors/NavigationPropertyEditorContext.tsx
grigasp: made `NavigationPropertyEditorContext` and `useNavigationPropertyEditorContext` internal

saskliutas (approved):

Issue #423, opened 2024-02-19
`hierarchies`: Training material


grigasp:
Add training material for using `@itwin/presentation-hierarchies` package to create hierarchies.

```[tasklist]
### Tasks
- [ ] https://github.com/iTwin/presentation/pull/667
- [ ] https://github.com/iTwin/presentation/pull/665
- [ ] https://github.com/iTwin/presentation/pull/672
- [ ] https://github.com/iTwin/presentation/pull/675
- [ ] https://github.com/iTwin/presentation/pull/678
- [ ] https://github.com/iTwin/presentation/pull/682
- [ ] https://github.com/iTwin/presentation/pull/679
- [ ] https://github.com/iTwin/presentation/pull/680
- [ ] https://github.com/iTwin/presentation/pull/685
- [ ] https://github.com/iTwin/presentation/pull/690
```

PR #663, opened 2024-07-04
Add ability to include compiled code in markdowns

grigasp:
Prerequisite for https://github.com/iTwin/presentation/issues/423.

The problem with including code into markdowns is that it can easily get obsolete. Even in case of the `README.md` we had an example that didn't really work... So it's much better if we extract the code from projects that compile and run. My approach adds some artifacts to markdown files, but they're not displayed in preview mode, so IMO that's okay.

The workflow would be like this:
1. You write a markdown and, instead of adding a piece of code, you add this:
   ```markdown
   <!-- [[include: Code.Piece.Identifier, ts]] -->
   ```
2. Then, in some package that gets compiled - ideally our full stack tests or the test app - wrap the piece of code like this:
   ```
   // __PUBLISH_EXTRACT_START__ Code.Piece.Identifier
   // a piece of code to be extracted
   // __PUBLISH_EXTRACT_END__
   ```
   Note: `__PUBLISH_EXTRACT_START__` and `__PUBLISH_EXTRACT_END__` tags are already used in `itwinjs-core`.
3. Run `pnpm lage docs` to get code pieces extracted.
4. Run `pnpm lage update-extractions` to inject those pieces in appropriate places. This modifies the target markdown files - review the changes and commit them.
5. The pipeline is set up to run `pnpm lage check-extractions`, which does pretty much the same as `update-extractions`, but fails if any changes to target files are detected. So CI will fail if you forget steps 3-4.

I understand this is not ideal and involves quite a few steps, but we're not adding / editing those code pieces to markdowns that often, so IMO a little bit of inconvenience is justified by the fact that we'll not have out-of-date code examples.

@iTwin/itwinjs-core-presentation, waiting for opinions about this approach.


scripts/updateExtractions.js
```diff
@@ -0,0 +1,147 @@
...
+const { execFileSync } = require("child_process");
+const yargs = require("yargs");
+const argv = yargs(process.argv).argv;
+
+// parse args
+const isCheck = "check" in argv;
+const targets = argv.targets;
+if (!targets) {
+  console.error(
+    `Fail! Please specify the "target" argument as a comma-separated list of paths to the target directories and files where extraction insertions need to be made.`,
```
Yato333, line 18:
```suggestion
    `Fail! Please specify the "targets" argument as a comma-separated list of paths to the target directories and files where extraction insertions need to be made.`,
```

scripts/updateExtractions.js
```diff
@@ -0,0 +1,147 @@
...
+// gather extractions from different packages into the workspace root
+console.log(`Gathering extractions from different packages...`);
+execFileSync("node", [path.join(__dirname, "gatherDocs.js")], { stdio: "inherit", cwd: path.join(__dirname, "..") });
+
+// set up constants
+const extractionsDir = path.join(__dirname, "..", "build/docs/extract");
+const extractionStart = "<!-- BEGIN EXTRACTION -->";
+const extractionEnd = "<!-- END EXTRACTION -->";
+const targetFileExtensions = [".ts", ".tsx", ".md"];
+const re = /^(\s*)(<!--|\/\/|\/\*)\s*\[\[include:\s*([\w\d\._-]+)(,[\s]*([\w\d_]+))?\]\]/;
```
Yato333, line 35: You can mark groups that you don't want captured `(?: ...)`, then capture indices would go in order:
```ts
const reIndentIndex = 1;
const reExtractionNameIndex = 2;
const reExtractionTypeIndex = 3;
```

scripts/updateExtractions.js
```diff
@@ -0,0 +1,147 @@
...
+      handleTargetFile(path.join(target, fileName));
+    });
+  } else if (targetStat.isFile(target)) {
+    handleTargetFile(target);
+  } else {
+    console.error(`Fail! Target "${target}" is not a valid directory or file.`);
+    process.exit(1);
+  }
+
+  if (isCheck) {
```
Yato333, line 57: What if code examples are modified, committed and not extracted? This doesn't seem to check that the extracted code matches the code in the documentation

grigasp: Not sure I understand what you're asking. 

You're not supposed to edit such snippets directly in the markdown - if you do that, the CI will fail as the snippet is going to be replaced with the extraction from compiled code. Then, since CI runs this script with `--check` flag, we're going to do `git status` and detect file modifications, which means the snippet in markdown doesn't match the extraction from compiled code.

scripts/updateExtractions.js
grigasp: good point, applied the change

scripts/updateExtractions.js
Yato333: Oh, I see now, I thought that `git status` was done before the changes

Yato333 (approved):

PR #683, opened 2024-07-18
Components: Reload tree on briefcase changes

jasdom:
Fixes https://github.com/iTwin/itwinjs-core/issues/6955.


packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
```diff
@@ -49,10 +49,18 @@ describe("usePresentationTreeState", () => {
...
+  const addListenerStub = sinon.stub<[() => void], () => void>();
 
   const imodel = {
     key: "test-imodel-key",
-  } as IModelConnection;
+    isBriefcaseConnection: isBriefcaseConnectionStub,
+    txns: {
+      onChangesApplied: {
+        addListener: addListenerStub,
+      },
```
saskliutas, line 61: Maybe change it into `BeEvent`? It will make test more readable as you will be able to raise event instead of invoking listener directly.

saskliutas (approved):

packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
jasdom: Changed.

saskliutas (approved):

packages/components/src/test/tree/controlled/UsePresentationTreeState.test.ts
```diff
@@ -280,15 +279,22 @@ describe("usePresentationTreeState", () => {
...
 
-      expect(addListenerStub).to.be.calledOnce;
-      const listener = addListenerStub.getCall(0).args[0];
+      onChangesPulledEvent.raiseEvent();
 
-      listener();
+      await waitFor(() => expect(result.current?.nodeLoader).to.not.eq(oldNodeLoader));
+    });
+
+    it("creates a new nodeLoader when briefcase `onChangesPushed` event is raised", async () => {
```
grigasp, line 292: why do we need to reload on push?

jasdom: When pushing, the changes are pulled, merged and applied. I originally reloaded on the `onChangesApplied` event, however that does not always work as the connection can be closed right after the event is raised, which leads to the same issue. The `onChangesPulled` and `onChangesPushed` events are raised after the changes are applied and the connection is reopened for the last time, so the reload occurs after the operation is fully completed.

PR #688, opened 2024-07-31
Hierarchies: Allow limiting auto-expansion to grouping nodes

jasdom:
Changes for https://github.com/iTwin/viewer-components-react/issues/947.

The API for allowing expansion up to grouping nodes is mostly done, however some cleanup is required. I am not entirely sure whether the filter attributes are correctly applied to custom nodes when multiple `matchingFilters` are returned as it currently takes the last `filterTarget` (though it seems that is also the case for instance nodes?). I've covered the changes with tests, however it would be wise to add integration tests to ensure everything works correctly when put together.


packages/hierarchies/src/hierarchies/internal/FilteringHierarchyDefinition.ts
```diff
@@ -42,41 +49,73 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+      // Expand grouping node if a grouping node below it is a filter target.
+      if (typeof child.filtering.filterTarget === "object") {
+        const nodeDepth = node.parentKeys.length;
+        const filterTargetDepth = child.filtering.filterTarget.parentKeysCount;
+        if (nodeDepth < filterTargetDepth) {
+          return true;
+        }
+      }
+
+      if (!child.filtering.filteredChildrenIdentifierPaths?.length) {
```
grigasp, line 85:
```suggestion
      if (!child.filtering.filteredChildrenIdentifierPaths) {
```

packages/hierarchies/api/presentation-hierarchies.api.md
```diff
@@ -26,7 +26,7 @@ interface BaseHierarchyNode {
         [key: string]: any;
     };
     filtering?: {
-        isFilterTarget?: boolean;
+        filterTarget?: FilterTarget;
```
grigasp, line 29: I feel like it's not completely correct to set grouping node's information to a "filter target" prop - filter target should always be the end of a filtering path. The grouping node is not a filter target, but instead and "auto expand target". What do thou think about keeping the `isFilterTarget` prop and additionally having the `autoExpand` one?

Yato333: Since this change was done by @jasdom, and you've said that everything looks OK, I didn't want to change it. Although, now that you've mentioned it, I agree with your suggestion. 

Come to think of it, why is `isFilterTarget` or even `filtering` property a part of `BaseHierarchyNode`, if grouping nodes can never be filter targets?

Yato333: What do you think about this?
```ts
filtering?: {
    filteredChildrenIdentifierPaths?: HierarchyFilteringPath[];
  } & (
    | { isFilterTarget: never }
    | {
        isFilterTarget: true;
        autoExpandUntil?: FilterTargetGroupingNodeInfo;
      }
  );
  ```

Comment (grigasp) [reply packages/hierarchies/api/presentation-hierarchies.api.md:29]: > why is `isFilterTarget` or even `filtering` property a part of `BaseHierarchyNode`, if grouping nodes can never be filter targets?

Good question - probably they shouldn't be.

> What do you think about this?

The idea looks good. Having `{ isFilterTarget: never }` may be inconvenient to use, e.g. I think you won't be able to do something like `isFilterTarget: x > 10`. Maybe change it to `undefined | false`?

Comment (Yato333) [reply packages/hierarchies/api/presentation-hierarchies.api.md]: Changed the API to how you've recommended

File: packages/hierarchies/src/hierarchies/HierarchyProvider.ts [hunk 3]
```diff
@@ -144,11 +145,16 @@ interface HierarchyProviderLocalizedStrings {
   other: string;
 }
 
+/** @beta */
+interface HierarchyFilteringPathOptions {
+  autoExpand?: boolean | FilterTargetGroupingNodeInfo;
```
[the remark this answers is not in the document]
Yato333: Fixed

.changeset/yellow-spies-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies": minor
+---
+
+Add ability to prevent auto-expanding of a grouping node in filtered hierarchies which target grouping nodes.
```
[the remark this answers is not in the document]
Yato333: Fixed

packages/hierarchies/src/hierarchies/HierarchyNode.ts
```diff
@@ -28,7 +28,20 @@ export interface FilterTargetGroupingNodeInfo {
...
-export type FilterTarget = boolean | FilterTargetGroupingNodeInfo;
+export type HierarchyNodeFilteringProps = {
+  /** If set to true, then one of the ancestor nodes in the hierarchy is the filter target. */
+  hasFilterTargetAncestor?: boolean;
+  /** Paths to node's children that are filter targets. */
+  filteredChildrenIdentifierPaths?: HierarchyFilteringPath[];
+} & (
+  | { isFilterTarget?: false }
+  | {
+      /** Whether or not this node is a filter target.. */
```
grigasp, line 39:
```suggestion
      /** Whether or not this node is a filter target. */
```

packages/hierarchies/src/presentation-hierarchies.ts
```diff
@@ -21,7 +21,18 @@ export {
...
+  GroupingNodeKey,
+  HierarchyNodeKey,
+  InstancesNodeKey,
+  ClassGroupingNodeKey,
+  LabelGroupingNodeKey,
+  PropertyGroupingNodeKey,
+  PropertyOtherValuesGroupingNodeKey,
+  PropertyValueGroupingNodeKey,
+  PropertyValueRangeGroupingNodeKey,
+  StandardHierarchyNodeKey,
```
grigasp, line 34: Do you think we need to export standard node key?

packages/hierarchies/src/hierarchies/internal/FilteringHierarchyDefinition.ts
```diff
@@ -27,6 +34,8 @@ export interface FilteringQueryBuilderProps {
   nodeIdentifierPaths: HierarchyFilteringPath[];
 }
 
+type FilterTarget = boolean | FilterTargetGroupingNodeInfo;
```
grigasp, line 37: You separated `isFilterTarget` and `autoExpandUntil` attributes in `HierarchyNode`, but them combined here. Is there a reason?

packages/hierarchies/src/hierarchies/internal/FilteringHierarchyDefinition.ts
```diff
@@ -50,33 +59,67 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+          return true;
+        }
+
+        // Expand grouping node if a grouping node below it is a filter target.
+        const nodeDepth = node.parentKeys.length;
+        const filterTargetDepth = child.filtering.autoExpandUntil.parentKeysCount;
+        if (nodeDepth < filterTargetDepth) {
+          return true;
+        }
+      }
```
grigasp, line 81: comments need update

packages/hierarchies/src/hierarchies/HierarchyNode.ts
```diff
@@ -18,6 +18,31 @@ import {
...
+      isFilterTarget: true;
+      /** Info about a grouping node that auto-expansion should be limited to.  */
```
grigasp, line 41:
```suggestion
      /**
       * Generally, when a hierarchy is filtered, it's auto-expanded up to filter targets or not 
       * expanded at all. This attribute may optionally be set to an instance node's grouping node
       * identifier to limit auto-expansion only up to that grouping node.
       */
```

packages/hierarchies/learning/HierarchyFiltering.md
```diff
@@ -236,3 +236,197 @@ The second step would be to create the node identifier paths. Let's consider two
...
+
+<!-- END EXTRACTION -->
+
+Additionally, hierarchies may contain grouping nodes.
+Grouping nodes can be displayed to the user as any other nodes, however they do not represent actual instances, so filtering cannot be applied to them.
+Nevertheless, it's possible to "target" grouping node in the hierarchy by filtering the hierarchy by node's grouped instance keys.
+However in some cases it may be useful to leave this grouping node collapsed in the resulting hierarchy.
+
+For this purpose, there's an option to limit auto-expansion of nodes to a given grouping node that is expected to be in the resulting hierarchy.
+In order to do that, there has to be provided a minimal set of information about the grouping node: node's key and depth in the hierarchy, which can be determined from `parentKeys` attribute.
```
grigasp, line 314:
```suggestion
In order to do that, there has to be provided a minimal set of information about the grouping node: node's key and depth in the hierarchy, which can be determined from the `parentKeys` attribute.
```

You say that we need to supply node's key and depth, but we call the latter `parentKeysCount`. Maybe the attribute should be renamed to `depth`? I think that would make it's meaning more obvious.

apps/full-stack-tests/src/hierarchies/learning-snippets/Utils.ts
```diff
@@ -4,17 +4,25 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { collect } from "presentation-test-utilities";
-import { HierarchyNode, HierarchyProvider } from "@itwin/presentation-hierarchies";
+import { HierarchyNode, HierarchyNodeKey, HierarchyProvider } from "@itwin/presentation-hierarchies";
 
 interface NodeDefinition {
+  nodeType: "custom" | Exclude<HierarchyNodeKey, string>["type"];
```
grigasp, line 10: why "custom" instead of the actual custom node key?

apps/full-stack-tests/src/hierarchies/learning-snippets/HierarchyFiltering.test.ts
```diff
@@ -49,6 +50,11 @@ describe("Hierarchies", () => {
         imodel = res.imodel;
         elementKeys = elements;
         elementIds = Object.entries(elements).reduce((acc, [name, instanceKey]) => ({ ...acc, [name]: instanceKey.id }), {} as { [name: string]: Id64String });
+
+        // Logger.initializeToConsole();
```
grigasp, line 54: needs cleanup

packages/hierarchies/src/hierarchies/HierarchyProvider.ts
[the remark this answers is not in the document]
grigasp: don't see it

apps/full-stack-tests/src/hierarchies/learning-snippets/HierarchyFiltering.test.ts
```diff
@@ -253,6 +259,169 @@ describe("Hierarchies", () => {
...
+                autoExpand: true,
+                children: [{ label: "C" }],
+              },
+            ],
+          },
+        ]);
+        // __PUBLISH_EXTRACT_END__
+      });
+
+      it("sets auto-expand flag to parent nodes of the filter target until a given groupingNode", async function () {
```
grigasp, line 300:
```suggestion
      it("sets auto-expand flag to parent nodes of the filter target until a given grouping node", async function () {
```

packages/hierarchies/src/presentation-hierarchies.ts
Yato333: I think we do, since we have `HiearchyNodeKey.isStandard()` function.

packages/hierarchies/src/hierarchies/internal/FilteringHierarchyDefinition.ts
Yato333: This greatly simplifies the amount of changes required to the logic. Also, there'd have to be 1 row added to the query in `applyECInstanceIdsFilter`.

packages/hierarchies/learning/HierarchyFiltering.md
Yato333: I think this is a double-edged sword. On the other hand, `depth` is more clear in regards to what this means in the internal logic, on the other hand `parentKeysCount` is more clear for the consumer in regards to what needs to be passed there. I am not sure what's better.

apps/full-stack-tests/src/hierarchies/learning-snippets/Utils.ts
Yato333: Because here, I wanted to store `node.key.type` and not the value of the key. But custom keys are just strings.

packages/hierarchies/learning/HierarchyFiltering.md
Yato333: Will rename to `depth`

packages/hierarchies/src/hierarchies/internal/FilteringHierarchyDefinition.ts
Yato333: Updated the comments

.changeset/yellow-spies-listen.md
```diff
@@ -2,4 +2,22 @@
 "@itwin/presentation-hierarchies": minor
 ---
 
-Add ability to prevent auto-expanding of a grouping node in filtered hierarchies which target grouping nodes.
+- Add hierarchy node key types to the barrel exports.
+- Add ability to prevent auto-expanding of a grouping node when filtering hierarchies:
```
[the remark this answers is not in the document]
Yato333: Separated 2 changelogs

packages/hierarchies/src/hierarchies/HierarchyProvider.ts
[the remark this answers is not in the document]
Yato333: Sorry about that. Added documentation

apps/full-stack-tests/src/hierarchies/learning-snippets/HierarchyFiltering.test.ts
Yato333: Removed

Yato333:
@grigasp awesome!
P.S. I've moved `FilterTargetGroupingNodeInfo` from HierarchyNode.ts to HierarchyProvider.ts where it's actually being used.

packages/hierarchies/src/hierarchies/internal/FilteringHierarchyDefinition.ts
grigasp: made the changes to remove ambiguity

grigasp (approved):

PR #690, opened 2024-08-07
Hierarchies: Add a guide for migrating from Presentation Rules system

grigasp:
Part of https://github.com/iTwin/presentation/issues/423.

### DIFF
--- a/packages/hierarchies/learning/PresentationRulesMigrationGuide.md
+++ b/packages/hierarchies/learning/PresentationRulesMigrationGuide.md
@@ -1,0 +1,779 @@
+# Migration guide from Presentation Rules
+
+This package replaces the part of the [Presentation Rules](https://www.itwinjs.org/presentation/) system that is intended for building hierarchies. The Presentation Rules are a convenient way to define a hierarchy in a declarative way, but the system has a number of limitations:
+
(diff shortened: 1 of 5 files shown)


packages/hierarchies/learning/PresentationRulesMigrationGuide.md
```diff
@@ -0,0 +1,779 @@
...
+    `,
+  },
+};
+```
+
+<!-- END EXTRACTION -->
+
+### Migrating same label grouping
+
+Same label grouping is different from other types of groupings, because it doesn't result in a grouping node being created. Instead, nodes with the same label get merged into a single instance node, which represents a merged set of instances. In the Presentation Rules system there are two options to choose from - grouping at query stage or at post-processing stage. The first one is more performance while the second one is more flexible.
```
Yato333, line 704:
```suggestion
Same label grouping is different from other types of groupings, because it doesn't result in a grouping node being created. Instead, nodes with the same label get merged into a single instance node, which represents a merged set of instances. In the Presentation Rules system there are two options to choose from - grouping at query stage or at post-processing stage. The first one is more performant while the second one is more flexible.
```

Yato333 (approved):

Issue #696, opened 2024-08-14
`components`: Unique values not filtered by selected class


grigasp:
In the PropertyFilterBuilderRenderer, for certain properties, the value selector would include results that doesn't exist for that class. For example, for the class VolBodies, when selecting the property HEIGHT, it also included values from other classes that also have the HEIGHT property (pc_ps_Volume_HEIGHT).

Current behavior:
![Image](https://github.com/user-attachments/assets/ab5b59a2-9162-4746-b790-9c3e802791de)
 
Desired behavior:
![Image](https://github.com/user-attachments/assets/5abaa717-7c62-48b7-aa3c-7bc325d0e730)

PR #704, opened 2024-09-24
[components]: Fix filtering by selected class

MartynasStrazdas:
Closes https://github.com/iTwin/presentation/issues/696


packages/components/src/presentation-components/instance-filter-builder/PresentationFilterBuilder.tsx
```diff
@@ -117,6 +117,8 @@ export interface PresentationFilterBuilderValueRendererProps extends PropertyFil
   descriptor: Descriptor;
   /** Keys used to create the descriptor. */
   descriptorInputKeys?: Keys;
+  /** Currently selected classes. */
```
grigasp, line 120: I think it would be great to provide more docs here. E.g.:
- How this prop affects the component's behavior?
- What's the default when prop is not provided?

Basically, try to think as an API consumer and make sure the docs are sufficient for you to understand the meaning of this prop.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -160,6 +162,7 @@ function useUniquePropertyValuesRuleset(descriptorRuleset?: Ruleset, field?: Fie
             specifications: [
               {
                 specType: "SelectedNodeInstances",
+                acceptableClassNames: selectedClasses && selectedClasses.map(({ name }) => name.split(":")[1]),
```
grigasp, line 165: I believe you should also specify the [acceptablePolymorphically](https://www.itwinjs.org/presentation/content/selectednodeinstances/#attribute-acceptablepolymorphically) attribute.

grigasp (commented):
There're no tests that prove your changes are working.

packages/components/src/presentation-components/instance-filter-builder/PresentationFilterBuilder.tsx
MartynasStrazdas: Changed

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
MartynasStrazdas: As I understand now that we provide the exact class name there should not be any need for `acceptablePolymorphically`. I added it with the default false flag, unless you have an idea why we would need it.

apps/full-stack-tests/src/components/properties/PresentationFilterBuilderValueRenderer.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      testCategory,
+      "PropertyName",
+      "PropertyName",
+      { valueFormat: PropertyValueFormat.Primitive, typeName: "string" },
+      false,
+      0,
+      testClassInfos.map((classInfo) => ({ property: { classInfo, name: "PropertyName", type: "string" } })),
+    );
+
+    const testDescriptor = new Descriptor({
```
saskliutas, line 101: As this is full stack test I think we should not mock data.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -160,6 +162,8 @@ function useUniquePropertyValuesRuleset(descriptorRuleset?: Ruleset, field?: Fie
             specifications: [
               {
                 specType: "SelectedNodeInstances",
+                acceptableClassNames: selectedClasses ? selectedClasses.map(({ name }) => name.split(":")[1]) : undefined,
+                acceptablePolymorphically: false,
```
saskliutas, line 166: I think this should be set to `true` as we will want to get results from classes that derive from selected classes.

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
grigasp: As @saskliutas mentioned in the other comment, we want to handle the classes polymorphically - this is generally how people expect classes to work, unless explicitly specifying that they're not going to be handled polymorphically. I don't see a reason we shouldn't do that.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
MartynasStrazdas: Changed

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
MartynasStrazdas: Seems I was under the wrong impression about how the rule would handle child classes in this case, changed to True.

saskliutas (changes requested):

apps/full-stack-tests/src/components/properties/PresentationFilterBuilderValueRenderer.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+    const combobox = await waitFor(() => getByRole("combobox"));
+
+    await waitFor(async () => {
+      await user.click(combobox);
+      expect(queryByText(baseElement, "Value1")).to.not.be.null;
+      expect(queryByText(baseElement, "Value2")).to.be.null;
+    });
+  });
+
+  it("renders 'PresentationFilterBuilderValueRenderer' with correct property values when selected classes are provided without keys", async function () {
```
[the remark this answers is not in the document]
MartynasStrazdas: It looks like the descriptorKeys property removal from the test was not saved and the test was indeed a duplicate. For the descriptor it does not matter what ruleset it contains, because we only need to find the field from it, which rule is used to fill the value dropdown depends on whether `PresentationFilterBuilderValueRenderer` has descriptorKeys passed to it. Removed the keys property can rewrite the descriptor rule if that would make the test more readable in your opinion.

apps/full-stack-tests/src/components/properties/PresentationFilterBuilderValueRenderer.test.tsx
```diff
@@ -0,0 +1,261 @@
...
+        operator={"is-equal"}
+      />,
+    );
+
+    // trigger loadTargets function
+    const user = userEvent.setup();
+    const combobox = await waitFor(() => getByRole("combobox"));
+
+    await waitFor(async () => {
+      await user.click(combobox);
```
saskliutas, line 256: Is there a reason we click is done in `waitFor`?

saskliutas (changes requested):

apps/full-stack-tests/src/components/properties/PresentationFilterBuilderValueRenderer.test.tsx
MartynasStrazdas: Leaving outside the waitFor seems to cause an act warning. Moved it to the inside to avoid it flooding the console.

saskliutas: user-event library should add act wrappers to avoid act warnings.

MartynasStrazdas: Changed

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -160,6 +162,8 @@ function useUniquePropertyValuesRuleset(descriptorRuleset?: Ruleset, field?: Fie
             specifications: [
               {
                 specType: "SelectedNodeInstances",
+                acceptableClassNames: selectedClasses ? selectedClasses.map(({ name }) => name.split(":")[1]) : undefined,
```
saskliutas, line 165: I thinks it's better to use `split(/[\.:]/)` as this might not be coming from descriptor. So just to be safe.

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
MartynasStrazdas: changed

saskliutas (approved):

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -168,7 +172,12 @@ function useUniquePropertyValuesRuleset(descriptorRuleset?: Ruleset, field?: Fie
       return;
     }
 
-    const classInfos = getFieldClassInfos(field);
+    const classInfos = selectedClasses
+      ? selectedClasses?.map((classInfo) => ({
+          ...classInfo,
+          name: classInfo.name.replace(".", ":"),
```
saskliutas, line 178: I would suggest updating `createSchemaClasses` to use `split(/[\.:]/)`.

MartynasStrazdas: Thats a better solution, changed

PR #705, opened 2024-09-24
Calculate rounding error for numeric values

saskliutas:
Closes https://github.com/iTwin/appui/issues/990

Also bumped Itwinjs-core and AppUI versions.


packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
```diff
@@ -374,7 +374,14 @@ function parseGenericFilterRule(rule: GenericInstanceFilterRule, ctx: GenericFil
   return {
     field,
     operator: rule.operator,
-    value: rule.value ? { valueFormat: PropertyValueFormat.Primitive, displayValue: rule.value.displayValue, value: rule.value.rawValue } : undefined,
+    value: rule.value
+      ? {
+          valueFormat: PropertyValueFormat.Primitive,
+          displayValue: rule.value.displayValue,
+          value: rule.value.rawValue,
+          ...(rule.value && "roundingError" in rule.value ? { roundingError: rule.value.roundingError } : undefined),
```
grigasp, line 382: do we need the `rule.value` check?

```suggestion
          ...("roundingError" in rule.value ? { roundingError: rule.value.roundingError } : undefined),
```

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
```diff
@@ -520,7 +527,11 @@ function toGenericInstanceFilterRuleValue(primitiveValue?: PrimitiveValue): Gene
     return undefined;
   }
 
-  return { displayValue: primitiveValue.displayValue ?? "", rawValue: primitiveValue.value };
+  return {
+    displayValue: primitiveValue.displayValue ?? "",
+    rawValue: primitiveValue.value,
+    ...(primitiveValue.roundingError ? { roundingError: primitiveValue.roundingError } : undefined),
```
grigasp, line 533: `roundingError` is numeric, so I think it's better to check if it exists rather for truthiness

```suggestion
    ...("roundingError" in primitiveValue ? { roundingError: primitiveValue.roundingError } : undefined),
```

packages/components/src/presentation-components/instance-filter-builder/PresentationInstanceFilter.ts
```diff
@@ -520,7 +527,11 @@ function toGenericInstanceFilterRuleValue(primitiveValue?: PrimitiveValue): Gene
...
+  } as GenericInstanceFilterRuleValue;
```
grigasp, line 534:
```suggestion
  } satisfies GenericInstanceFilterRuleValue;
```

packages/components/src/presentation-components/properties/inputs/NumericPropertyInput.tsx
```diff
@@ -57,6 +50,16 @@ export const NumericPropertyInput = forwardRef<PropertyEditorAttributes, Numeric
 });
 NumericPropertyInput.displayName = "NumericPropertyInput";
 
+function getPrimitiveValue(value: string): PrimitiveValue {
```
grigasp, line 53: maybe rename to `parse...`?

packages/components/src/presentation-components/properties/inputs/Utils.ts
```diff
@@ -0,0 +1,123 @@
...
+  const matches = (12345.6789).toLocaleString().match(/345(.*)67/);
+  if (matches && matches.length > 1) {
+    localeSpecificDecimalSeparator = matches[1];
+  }
+
+  return localeSpecificDecimalSeparator;
+}
+
+function isFractionalUnit(unitLabel: string) {
+  return ["ft", "in", "yd"].includes(unitLabel.toLowerCase());
```
grigasp, line 82: I wonder if this list is all we need? Aren't there more units that use fractional precision?

grigasp (approved):

PR #711, opened 2024-10-01
Fix slow hierarchy filtering with large number of filtered paths

JonasDov:
Closes [#706](https://github.com/iTwin/presentation/issues/706)


apps/full-stack-tests/src/hierarchies/HierarchyFiltering.test.ts
```diff
@@ -214,7 +214,6 @@ describe("Hierarchies", () => {
               children: [
                 NodeValidators.createForGenericNode({
                   key: "custom2",
-                  autoExpand: false,
```
grigasp, line 217: why was this removed?

apps/full-stack-tests/src/hierarchies/HierarchyFiltering.test.ts
```diff
@@ -1528,19 +1531,24 @@ describe("Hierarchies", () => {
...
+                  const identifierIndexAdded = identifierIndex + 1;
                   return (
-                    path.length &&
-                    HierarchyNodeIdentifier.isGenericNodeIdentifier(path[0]) &&
-                    path[0].source === "custom-provider" &&
-                    path[0].id === myNode.key.id
+                    path.length > identifierIndexAdded &&
+                    HierarchyNodeIdentifier.isGenericNodeIdentifier(path[identifierIndexAdded]) &&
+                    path[identifierIndexAdded].source === "custom-provider" &&
+                    path[identifierIndexAdded].id === myNode.key.id
```
grigasp, line 1551: Looks like implementing hierarchy filtering in a custom hierarchy provider became much more complicated. We should not enforce any specific implementation requirements for custom providers - they should be able to choose the most convenient way to filter the hierarchy.

packages/hierarchies/api/presentation-hierarchies.api.md
```diff
@@ -159,7 +159,8 @@ interface ECSqlValueSelector {
 
 // @beta
 export function extractFilteringProps(rootLevelFilteringProps: HierarchyFilteringPath[], parentNode: Pick<NonGroupingHierarchyNode, "filtering"> | undefined): {
-    filteredNodePaths: HierarchyFilteringPath[];
+    filterPathsIdentifierPositions?: Array<[number, number]>;
+    nodeIdentifierPaths: HierarchyFilteringPath[];
```
grigasp, line 163: This is a breaking change that we can't make anymore

grigasp: The other changes in this file are also breaking

apps/performance-tests/src/hierarchies/Filtering.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  for (let i = 0; i < totalNumberOfFilteringPaths; ++i) {
+    const hundredsPosition = Math.floor(i / itemsPerGroup);
+    const id = `0x${(i + startingIndex).toString(16)}`;
+    const nearestParentId = `0x${parentIdsArr[i % Math.ceil(totalNumberOfFilteringPaths / maximumNumberOfPathsForSingleParent)].toString(16)}`;
+    filteringPaths.push([
+      { className: `${schemaName}.${defaultClassName}_0`, id: `0x${startingIndex.toString(16)}` },
+      { className: `${schemaName}.${defaultClassName}_0`, id: nearestParentId },
+      { className: `${schemaName}.${defaultClassName}_${hundredsPosition}`, id },
+    ]);
+  }
```
grigasp, line 156: please add some comments explaining what's happening here

grigasp (commented):
didn't look at the changes in `FilteringHierarchyDefinition` - need to resolve breaking changes first

apps/performance-tests/src/hierarchies/Filtering.test.ts
```diff
@@ -0,0 +1,157 @@
...
+import { createBisInstanceLabelSelectClauseFactory } from "@itwin/presentation-shared";
+
+describe("filtering", () => {
+  const totalNumberOfFilteringPaths = 10000;
+  const startingIndex = 20;
+  const maximumNumberOfPathsForSingleParent = 500;
+  const parentIdsArr = new Array<number>();
+  for (let i = maximumNumberOfPathsForSingleParent, j = startingIndex + 1; i <= totalNumberOfFilteringPaths; i += maximumNumberOfPathsForSingleParent, ++j) {
+    parentIdsArr.push(j);
+  }
```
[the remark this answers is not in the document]
JonasDov: Moved

apps/performance-tests/src/hierarchies/Filtering.test.ts
```diff
@@ -0,0 +1,157 @@
...
+                        extendedData: {
+                          ecInstanceId: { selector: `this.ECInstanceId` },
+                        },
+                      })}
+                      FROM ${fullClassName} AS this
+                      WHERE this.ECInstanceId = ${startingIndex}
+                    `,
+                  },
+                },
+              ];
```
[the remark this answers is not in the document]
JonasDov: Yes, added `createHierarchyLevelDefinition` function

apps/full-stack-tests/src/hierarchies/HierarchyFiltering.test.ts
JonasDov: Added back

packages/hierarchies/api/presentation-hierarchies.api.md
JonasDov: Changed

apps/performance-tests/src/hierarchies/Filtering.test.ts
JonasDov: Didn't add a comment, just simplified it.

.changeset/tall-comics-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies": minor
+---
+
+Increased the speed of hierarchy filtering with large number of filtered paths.
```
grigasp, line 5: should list all the api additions / changes you made in this pr

apps/performance-tests/src/hierarchies/Filtering.test.ts
```diff
@@ -0,0 +1,107 @@
...
+            { className: `${schemaName}.${defaultClassName}_0`, id: `0x${physicalElementsSmallestIndex.toString(16)}` },
+            { className: `${schemaName}.${defaultClassName}_${i / itemsPerGroup}`, id: `0x${(i + physicalElementsSmallestIndex).toString(16)}` },
+            { className: `${schemaName}.${defaultClassName}_${j / itemsPerGroup}`, id: `0x${(j + physicalElementsSmallestIndex).toString(16)}` },
+          ]);
+        }
+      }
+
+      const iModel = SnapshotDb.openFile(Datasets.getIModelPath("50k flat elements"));
+      const fullClassName = PhysicalElement.classFullName.replace(":", ".");
+      const createHierarchyLevelDefinition = async (imodelAccess: ECSchemaProvider & ECClassHierarchyInspector, eCInstanceIdCondition: string) => {
```
grigasp, line 40: - letter casing problem in `eCInstanceIdCondition`
- I suggest changing it to `whereClause: (alias: string) => string`

apps/performance-tests/src/hierarchies/Filtering.test.ts
```diff
@@ -0,0 +1,107 @@
...
+            //
+            // We need to split the hierarchy in 100 parts, because we are using 50000 paths and there is a limit of 500 filtering paths for a single parent.
+
+            if (!props.parentNode) {
+              return createHierarchyLevelDefinition(imodelAccess, `= ${physicalElementsSmallestIndex}`);
+            }
+
+            if (props.parentNode?.extendedData?.ecInstanceId === physicalElementsSmallestIndex) {
+              return createHierarchyLevelDefinition(imodelAccess, `IN (${parentIdsArr.join(", ")})`);
+            }
```
grigasp, line 86: Why use extended data for this? We could do this instead:
```ts
if (parentNode && HierarchyNode.isInstancesNode(parentNode) && parentNode.key.instanceKeys.some(({ id }) => Id64.getLocalId(id) === physicalElementsSmallestIndex)) {
  // ...
}
```

apps/performance-tests/src/hierarchies/Filtering.test.ts
```diff
@@ -0,0 +1,107 @@
...
+import { IModelDb, PhysicalElement, SnapshotDb } from "@itwin/core-backend";
+import { createNodesQueryClauseFactory, HierarchyFilteringPath } from "@itwin/presentation-hierarchies";
+import { Datasets } from "../util/Datasets";
+import { run } from "../util/TestUtilities";
+import { ProviderOptions, StatelessHierarchyProvider } from "./StatelessHierarchyProvider";
+import { createBisInstanceLabelSelectClauseFactory, ECClassHierarchyInspector, ECSchemaProvider } from "@itwin/presentation-shared";
+
+describe("filtering", () => {
+  const totalNumberOfFilteringPaths = 50000;
+  const physicalElementsSmallestIndex = 20;
```
grigasp, line 16: I believe this is not an index, but an id in decimal form.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -21,6 +21,7 @@ import {
 } from "./IModelHierarchyDefinition";
 import { ProcessedGroupingHierarchyNode, ProcessedHierarchyNode, SourceHierarchyNode, SourceInstanceHierarchyNode } from "./IModelHierarchyNode";
 import { defaultNodesParser } from "./TreeNodesReader";
+import { Dictionary } from "@itwin/core-bentley";
```
grigasp, line 24: Imports from other packages should be listed before imports from local modules. I suggest using https://marketplace.visualstudio.com/items?itemName=mike-co.import-sorter for that.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -151,7 +239,7 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
             this._imodelAccess,
             (def, matchingFilters) => {
               const filteredChildrenIdentifierPaths = matchingFilters.reduce(
-                (r, c) => [...r, ...c.childrenIdentifierPaths],
+                (r, c) => [...r, ...(c.childrenIdentifierPaths ? c.childrenIdentifierPaths : [])],
```
grigasp, line 242:
```suggestion
                (r, c) => [...r, ...(c.childrenIdentifierPaths ?? [])],
```

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -219,14 +307,18 @@ async function matchFilters<
...
           isFilterTarget: false,
         };
         matchingFilters.push(entry);
       }
-      const remainingPath = path.slice(1);
-      if (remainingPath.length > 0) {
-        entry.childrenIdentifierPaths.push(options ? { path: remainingPath, options } : remainingPath);
+
+      if (path.length > 1) {
+        if (HierarchyNodeIdentifier.isGenericNodeIdentifier(nodeId)) {
```
grigasp, line 316: why do we only need to do this for generic nodes from now on?

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -323,12 +419,12 @@ export function applyECInstanceIdsFilter(
...
           ${matchingFilters
             .map((mc) =>
               mc.isFilterTarget
-                ? `VALUES (${mc.id.id}, 1, ${mc.filterTargetOptions ? `'${JSON.stringify(mc.filterTargetOptions)}'` : "CAST(NULL AS TEXT)"}, '${JSON.stringify(mc.childrenIdentifierPaths)}')`
-                : `VALUES (${mc.id.id}, 0, CAST(NULL AS TEXT), '${JSON.stringify(mc.childrenIdentifierPaths)}')`,
+                ? `VALUES (${mc.id.id}, 1, ${mc.filterTargetOptions ? `'${JSON.stringify(mc.filterTargetOptions)}'` : "CAST(NULL AS TEXT)"}, '${mc.id.className}')`
+                : `VALUES (${mc.id.id}, 0, CAST(NULL AS TEXT), '${mc.id.className}')`,
             )
             .join(" UNION ALL ")}
         )`,
```
grigasp, line 430: I'm now thinking that we could group all instance keys by class name and change the CTE from 

```sql
VALUES(...) UNION ALL VALUES(...) UNION ALL ...
```
to 
```sql
SELECT ... FROM ONLY {class name 1} WHERE ECInstanceId IN (?,?,?,...)
UNION ALL
SELECT ... FROM ONLY {class name 2} WHERE ECInstanceId IN (?,?,?,...)
UNION ALL 
...
```

This should further reduce the size of SQL statement and possibly even help for https://github.com/iTwin/presentation/issues/722

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -106,16 +152,58 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
         : undefined;
-      const filteredChildrenIdentifierPaths: HierarchyFilteringPath[] | undefined = row[ECSQL_COLUMN_NAME_FilteredChildrenPaths]
-        ? JSON.parse(row[ECSQL_COLUMN_NAME_FilteredChildrenPaths])
-        : undefined;
-      const defaultNode = (this._source.parseNode ?? defaultNodesParser)(row);
+
+      // istanbul ignore next
+      const allFilterPathsIdentifierPositions: Array<[number, number]> | undefined = row[ECSQL_COLUMN_NAME_FilterECInstanceId]
+        ? (this._pathsIdentifierPositions.get({
+            id: `0x${row[ECSQL_COLUMN_NAME_FilterECInstanceId].toString(16)}`,
```
grigasp, line 165: we should select the id as id64 string, otherwise there's a risk for it to exceed javascript's number

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -106,16 +152,58 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+      for (const [pathIndex, identifierIndex] of allFilterPathsIdentifierPositions) {
+        // istanbul ignore if
+        if (this._nodeIdentifierPaths.length <= pathIndex) {
+          continue;
+        }
+        const path = HierarchyFilteringPath.normalize(this._nodeIdentifierPaths[pathIndex]);
+        // istanbul ignore if
+        if (path.path.length <= identifierIndex) {
+          continue;
+        }
```
grigasp, line 184: I'd like tests to cover these two edge cases.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -106,16 +152,58 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
       const isFilterTarget: boolean = !!row[ECSQL_COLUMN_NAME_IsFilterTarget];
       const filterTargetOptions: HierarchyFilteringPathOptions | undefined = row[ECSQL_COLUMN_NAME_FilterTargetOptions]
         ? JSON.parse(row[ECSQL_COLUMN_NAME_FilterTargetOptions])
         : undefined;
-      const filteredChildrenIdentifierPaths: HierarchyFilteringPath[] | undefined = row[ECSQL_COLUMN_NAME_FilteredChildrenPaths]
-        ? JSON.parse(row[ECSQL_COLUMN_NAME_FilteredChildrenPaths])
-        : undefined;
-      const defaultNode = (this._source.parseNode ?? defaultNodesParser)(row);
+
+      // istanbul ignore next
```
grigasp, line 162: There's a lot of code for doing one thing - it makes `parseNode` hard to read. Could we move it into a separate function?

.changeset/tall-comics-compare.md
JonasDov: Added

apps/performance-tests/src/hierarchies/Filtering.test.ts
JonasDov: Changed

apps/performance-tests/src/hierarchies/Filtering.test.ts
JonasDov: Removed extendedData

apps/performance-tests/src/hierarchies/Filtering.test.ts
JonasDov: Renamed

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: I had opened presentation repo in co-space, for some reason import sorter does not work in co-space. Fixed the order

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: Because we are not using `childrenIdentifierPaths` for `InstanceNodeIdentifiers`. Changed the implementation, now `matchFilters` function takes in another argument: `extractChildrenIdentifierPaths` which determines whether or not to retrieve `childrenIdentifierPaths`

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: I don't think this suggestion would work as keys with the same className and different id's can have different `IsFilterTarget` and `FilterTargetOptions` values.

JonasDov: Keeping as is

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: Changed the query to return id64 string instead.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: These two edge cases should never happen because:
We create `_pathsIdentifierPositions` array based on `_nodeIdentifierPaths`. So `_pathsIdentifierPositions` always has indexes pointing to an existing place in the `_nodeIdentifierPaths` array.
These two checks were added just a precaution. Removed them.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: Addded `getChildrenIdentifierPathsFromIdentifier` function.

.changeset/tall-comics-compare.md
```diff
@@ -0,0 +1,9 @@
+---
+"@itwin/presentation-hierarchies": minor
+---
+
+- Increased the speed of hierarchy filtering with large number of filtered paths.
+
+- Added `compare` function to `HierarchyNodeIdentifier`.
+
+- Changed `NodeParser`, it now returns `SourceInstanceHierarchyNode | Promise<SourceInstanceHierarchyNode>`.
```
grigasp, line 9: The first row should not be a bullet point, otherwise the final changelog renders weirdly. Also made API addition descriptions more specific.

```suggestion
Increased the speed of hierarchy filtering with large number of filtered paths.

- Added `HierarchyNodeIdentifier.compare` function to compare two identifiers.
- Changed `NodeParser` (return type of `HierarchyDefinition.parseNode`), to additionally allow returning a promise, so instead of just `SourceInstanceHierarchyNode` it can now also return `Promise<SourceInstanceHierarchyNode>`.
```

grigasp (commented):
I'd like a few tests to be added to `FilteringHierarchyDefinition.test.ts` to cover the edge cases we discussed earlier:
- filtering by multiple paths where ids are the same, but classes are different,
- filtering by multiple paths where the same identifier is at different positions of the path.

grigasp:
any idea why "Hierarchies performance benchmark" comment now contains unified selection performance numbers?

JonasDov:
> FilteringHierarchyDefinition.test.ts

Added 3 tests:
- `sets correct filteredChildrenIdentifierPaths when nodes have same id's and different classNames that don't derive from one another`
- `sets correct filteredChildrenIdentifierPaths when same identifier is in different positions of the path`
- `sets correct filteredChildrenIdentifierPaths when nodes have same id's and different classNames that derive from one another`

JonasDov:
> any idea why "Hierarchies performance benchmark" comment now contains unified selection performance numbers?

This was due to adding `spec: ["./lib/**/*.test.js"]` to `.mocharc.json` file.
I tried a few things to fix this:
1. Tried specifying `--spec` option in package.json, but it did not work. It seems from mocha docs, that they merge options that can be merged: https://mochajs.org/#merging
2. Tried using `MOCHA_OPTIONS`, but it did not work.
3. Tried to find a way to split  `.mocharc.json` file into two, but did not find a way to do so.
Decided to revert the change, and add the path to `launch.json`, by doing it this way the debugger works.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -86,6 +102,70 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+        }
+        ++similarityLength;
+        continue;
+      }
+      // istanbul ignore next
+      break;
+    }
+    entry = similarityLength;
+    this._pathsSimilarityLengthCache.set({ lhsIndex, rhsIndex }, entry);
+    return entry;
```
grigasp, line 166: what's the point of assigning this to `entry`?

```suggestion
    this._pathsSimilarityLengthCache.set({ lhsIndex, rhsIndex }, similarityLength);
    return similarityLength;
```

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -188,12 +276,30 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+      if (identifierIndex !== validPath.path.length - identifiersCountAfter - 1) {
+        continue;
+      }
+
+      if (pathIndex !== validPathIndex) {
+        const similarityLength = await this.getPathsSimilarityLength(pathIndex, validPathIndex);
+        // istanbul ignore if
+        if (similarityLength < identifierIndex) {
+          continue;
+        }
```
grigasp, line 298: should be covered with tests

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -86,6 +102,70 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+      }
+
+      // istanbul ignore next
+      if (HierarchyNodeIdentifier.isGenericNodeIdentifier(lhsIdentifier) && HierarchyNodeIdentifier.isGenericNodeIdentifier(rhsIdentifier)) {
+        if (lhsIdentifier.source !== rhsIdentifier.source || lhsIdentifier.id !== rhsIdentifier.id) {
+          break;
+        }
+        ++similarityLength;
+        continue;
+      }
```
grigasp, line 160: should be covered with tests

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -86,6 +102,70 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+      // istanbul ignore next
+      break;
```
grigasp, line 162: should be covered with tests

packages/hierarchies/src/hierarchies/HierarchyNode.ts
```diff
@@ -26,6 +26,8 @@ export type HierarchyNodeFilteringProps = {
   hasFilterTargetAncestor?: boolean;
   /** Paths to node's children that are filter targets. */
   filteredChildrenIdentifierPaths?: HierarchyFilteringPath[];
+  /** Paths index in all filter paths */
+  filteredChildrenIdentifierPathsIndex?: number[];
```
grigasp, line 30: I don't understand the meaning of this. Can you explain in more detail?

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -35,12 +35,28 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+   * This dictionary would return: 3, since both paths start with a, b, c identifiers (which are similar)
+   */
+  private _pathsSimilarityLengthCache: Dictionary<{ lhsIndex: number; rhsIndex: number }, number>;
 
   public constructor(props: FilteringHierarchyDefinitionProps) {
     this._imodelAccess = props.imodelAccess;
     this._source = props.source;
     this._nodeIdentifierPaths = props.nodeIdentifierPaths;
     this._pathsIdentifierPositions = this.createPathsIdentifierPositions(props.nodeIdentifierPaths);
+    this._pathsSimilarityLengthCache = this.createPathsSimilarityLengthCache();
```
grigasp: At this point we have class 3 members + a ton of code for handling those paths. Maybe it would be worth moving all of that into a separate class responsible just for that?

apps/full-stack-tests/src/hierarchies/HierarchyFiltering.test.ts
```diff
@@ -223,6 +223,106 @@ describe("Hierarchies", () => {
...
+              fullClassName: subjectClassName,
+              query: {
+                ecsql: `
+                  SELECT ${await selectQueryFactory.createSelectClause({
+                    ecClassId: { selector: `this.ECClassId` },
+                    ecInstanceId: { selector: `this.ECInstanceId` },
+                    nodeLabel: { selector: `this.CodeValue` },
+                    extendedData: {
+                      level: extendedDataLevel,
+                    },
```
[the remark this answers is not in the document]
JonasDov: Removed

apps/full-stack-tests/src/hierarchies/HierarchyFiltering.test.ts
```diff
@@ -223,6 +223,106 @@ describe("Hierarchies", () => {
...
+            }
+            if (HierarchyNode.isInstancesNode(parentNode) && parentNode.label === "test subject 4") {
+              return createHierarchyLevelDefinition("3", (alias: string) => `WHERE ${alias}.Parent.Id = ${keys.rootSubject.id}`);
+            }
+            if (HierarchyNode.isInstancesNode(parentNode) && parentNode.label === "test subject 1" && parentNode.extendedData?.level === "3") {
+              return createHierarchyLevelDefinition("4", (alias: string) => `WHERE ${alias}.Parent.Id = ${keys.childSubject1.id}`);
+            }
+            return [];
+          },
+        };
```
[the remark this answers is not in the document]
JonasDov: Applied changes

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -86,6 +102,70 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+      }
+      if (lhs.lhsIndex !== rhs.lhsIndex) {
+        // istanbul ignore next
+        return lhs.lhsIndex > rhs.lhsIndex ? 1 : -1;
+      }
+      // istanbul ignore next
+      return lhs.rhsIndex > rhs.rhsIndex ? 1 : -1;
+    };
+    return new Dictionary<{ lhsIndex: number; rhsIndex: number }, number>(compare);
+  }
```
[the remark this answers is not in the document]
JonasDov: Moved outside of the class

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: It is now covered

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: Covered

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: Covered

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: Created a separate class for handling positions

packages/hierarchies/src/hierarchies/HierarchyNode.ts
JonasDov: This tells which original nodeIdentifierPath is filteredChildrenIdentifierPaths pointing to.
For example if we create a `FilteringHierarchyDefinition` with these `nodeIdentifierPaths`: 
```ts
nodeIdentifierPaths: [
  [a, b, c],
  [a, c, d],
  [a, b, d]
]
```
And try to get `filteredChildrenIdentifierPaths` for node b, 
The following `filteredChildrenIdentifierPaths` would be created:
```ts
filteredChildrenIdentifierPaths: [
  [c],
  [d]
]
```
And the following `filteredChildrenIdentifierPathsIndex` would be created:
```ts
filteredChildrenIdentifierPathsIndex: [0, 2]
```
filteredChildrenIdentifierPathsIndex tells us that first path ([c]) in `filteredChildrenIdentifierPaths` is created from 0th path ([a, b, c]) of `nodeIdentifierPaths`. And the the second path ([d]) is created from 2nd path ([a, b, d]).

JonasDov: After talking in private, changed the implementation, this is no longer needed.

apps/full-stack-tests/src/hierarchies/HierarchyFiltering.test.ts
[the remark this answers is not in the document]
grigasp: shouldn't the root "test subject 1" node have 2 children - "test subject 2" and "test subject 3"?

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -199,10 +246,11 @@ async function matchFilters<
   TDefinition = TIdentifier extends InstanceKey ? InstanceNodesQueryDefinition : GenericHierarchyNodeDefinition,
 >(
   definition: TDefinition,
-  filteringProps: { filteredNodePaths: HierarchyFilteringPath[]; hasFilterTargetAncestor?: boolean },
+  filteringProps: { filteredNodePaths: HierarchyFilteringPath[]; hasFilterTargetAncestor?: boolean; filteredNodePathsIndex?: number[] },
```
grigasp, line 249: is `filteredNodePathsIndex` needed anymore?

.changeset/tall-comics-compare.md
grigasp: I'm now thinking it would be nice to also specify the scope of improvement, e.g.:

```
Increased the speed of hierarchy filtering with large number of filtered paths.

| Amount of paths | Before the change | After the change |
| --- | --- | --- |
| 1k | ...
| 10k | ...
| 50k | ...

In addition, changed `NodeParser` (return type of `HierarchyDefinition.parseNode`):
- It now can return a promise, so instead of just `SourceInstanceHierarchyNode` it can now also return `Promise<SourceInstanceHierarchyNode>`.
- Additionally, it now accepts an optional `parentNode` argument of `HierarchyDefinitionParentNode` type.
```

grigasp (approved):
Looks great now!

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: No, removed

.changeset/tall-comics-compare.md
JonasDov: Updated

Issue #701, opened 2024-09-10
Replace `react-select` usage with `ComboBox` from iTwinUI


saskliutas:
`react-select` usage should be replace with `ComboBox` from iTwinUI to make sure that styling is consistent. `react-select` is usage in cases where we might encounter huge amount of options and we chose to implement infinite scroll. The new suggested behavior is to limit dropdown menu options to some predetermined size and instruct user to use filtering to get relevant result. 

The new behavior removes the need for infinite scroll and should allow to use `ComboBox` from iTwinUI.

grigasp:
https://github.com/iTwin/iTwinUI/discussions/2276

morganadrales:
hey @grigasp, what version will this fix be pulled into?

grigasp:
> hey [@grigasp](https://github.com/grigasp), what version will this fix be pulled into?

Hi @morganadrales. I think we can make a release tomorrow - the change will land into `@itwin/presentation-components` version `5.6.0`

grigasp:
@morganadrales, just released: https://github.com/iTwin/presentation/releases/tag/%40itwin%2Fpresentation-components%405.6.0

morganadrales:
@grigasp thank you so much!

grigasp:
> [@grigasp](https://github.com/grigasp) thank you so much!

kudos to @MartynasStrazdas who made the change, I'm just the messenger :)

PR #739, opened 2024-10-18
Replace react-select with itwinUI combobox

MartynasStrazdas:
Closes https://github.com/iTwin/presentation/issues/701


.changeset/nine-ducks-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": minor
+---
+
+Replaced `react-select` with `iTwinUI` comboboxes, limited visible options to 100, and added a non-pickable option prompting users to filter beyond the limit.
```
grigasp, line 5: I think a change of this size requires a more detailed description. E.g. something like this:

```markdown
Replaced `react-select` with [iTwinUI's ComboBox](https://itwinui.bentley.com/docs/combobox).

This affects [list the public components that this affects] components.

Changes:

- The number of select options is limited to 100. When more items exist, a non-selectable option is displayed at the bottom of the list, prompting users to provide an items filter. Previously, additional pages of select options was loaded when user scrolled to the bottom of the list.

- Did the style change? A before/after screenshot would be great.

- Maybe something else?
```

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
```diff
@@ -1395,4 +1283,84 @@ describe("UniquePropertyValuesSelector", () => {
...
+          hasMore: false,
+        };
+      });
+      await itemsLoader.loadInitialItems();
+      await itemsLoader.loadItems("filterText");
+      expect(spy.calledOnce);
+    });
+
+    it("does not load items when items loader is disposed", async () => {
+      const getItemsStub = sinon.stub();
```
saskliutas, line 1344:
```suggestion
        spy
```

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/UseUniquePropertyValuesLoader.ts
```diff
@@ -0,0 +1,302 @@
...
+}
+
+function formatOptionLabel(displayValue: string, type: string): string {
+  if (displayValue === "") {
+    return translate("unique-values-property-editor.empty-value");
+  }
+
+  switch (type) {
+    case "dateTime":
+      return new Date(displayValue).toLocaleString(undefined, {
```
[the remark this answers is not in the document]
MartynasStrazdas: If the dates differ by less than a second, multiple options will have the same display value which causes unwanted behavior (ex. when one option is selected all options with the same label are selected)

packages/components/src/test/properties/editors/NavigationPropertyEditor.test.tsx
```diff
@@ -101,14 +113,14 @@ describe("<NavigationPropertyTargetEditor />", () => {
...
-    const select = await waitFor(() => getByRole("combobox"));
+    const select = await waitFor(() => getByPlaceholderText("navigation-property-editor.select-target-instance"));
     await user.click(select);
 
     // select option from dropdown
     const target = await waitFor(() => getByText(contentItem.label.displayValue));
     await user.click(target);
 
-    await waitFor(() => expect(queryByDisplayValue(contentItem.label.displayValue)).to.not.be.null);
+    await waitFor(() => getByDisplayValue(contentItem.label.displayValue));
```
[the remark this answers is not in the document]
MartynasStrazdas: I find `getBy` calls cleaner as you don't need to double check for the `.not` when you expect the item to exist in DOM.

apps/full-stack-tests/src/components/properties/PresentationFilterBuilderValueRenderer.test.tsx
```diff
@@ -18,6 +19,18 @@ describe("Presentation filter builder value renderer", () => {
   before(async () => {
     await initialize();
     await UiComponents.initialize(new EmptyLocalization());
+
+    sinon.stub(window.Element.prototype, "getBoundingClientRect").returns({
```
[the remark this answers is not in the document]
MartynasStrazdas: Fixed

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -88,59 +63,51 @@ export function UniquePropertyValuesSelector(props: UniquePropertyValuesSelector
...
 
   return (
-    <AsyncSelect
-      inputValue={searchInput}
+    <ComboBox
+      multiple={true}
+      enableVirtualization={true}
+      options={selectOptions}
+      onChange={(newValue) => {
+        const newSelectedValues = loadedOptions.filter((opt) => newValue.includes(opt.displayValue));
```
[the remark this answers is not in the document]
MartynasStrazdas: Moved

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -88,59 +63,51 @@ export function UniquePropertyValuesSelector(props: UniquePropertyValuesSelector
...
+        const filteredOptions = options
+          .filter((option) => option.label.toLowerCase().includes(inputValue.toLowerCase()) && option.value !== FILTER_WARNING_OPTION.value)
+          .slice(0, UNIQUE_PROPERTY_VALUES_BATCH_SIZE);
+
+        if (filteredOptions.length >= UNIQUE_PROPERTY_VALUES_BATCH_SIZE) {
+          filteredOptions.push(FILTER_WARNING_OPTION);
+        }
+        return filteredOptions;
+      }}
+      emptyStateMessage={translate("unique-values-property-editor.no-values")}
```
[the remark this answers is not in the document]
MartynasStrazdas: Added

packages/components/src/presentation-components/properties/inputs/UseNavigationPropertyTargetsLoader.ts
```diff
@@ -32,6 +34,12 @@ export interface NavigationPropertyTarget {
   key: InstanceKey;
 }
 
+/** @internal */
+export interface NavigationPropertyValuesLoaderState {
```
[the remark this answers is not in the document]
MartynasStrazdas: Considered moving it to another file as well as testing it through the custom hook is annoying when forcing the useEffect into edge cases.

packages/components/src/presentation-components/properties/inputs/UseNavigationPropertyTargetsLoader.ts
```diff
@@ -42,53 +50,78 @@ export interface NavigationPropertyTargetsResult {
...
+    };
+  }, [imodel, initialSelectedTarget, ruleset]);
+
+  // On filter text change, check if more items need to be loaded
+  useEffect(() => {
+    const timeout = setTimeout(() => {
+      if (itemsLoader?.needsLoadingItems(filterText)) {
+        void itemsLoader?.loadItems(filterText);
       }
+    }, 250);
```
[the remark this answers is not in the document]
MartynasStrazdas: Added a check at the top

packages/components/src/presentation-components/properties/inputs/UseUniquePropertyValuesLoader.ts
```diff
@@ -0,0 +1,302 @@
...
+    void loader.loadInitialItems();
+    setItemsLoader(loader);
+    return () => {
+      loader.dispose();
+    };
+  }, [imodel, ruleset, field, descriptorInputKeys, initialSelectedValues]);
+
+  // On filter text change, check if more items need to be loaded
+  useEffect(() => {
+    if (!filterText) {
```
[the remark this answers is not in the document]
MartynasStrazdas: Added

packages/components/src/presentation-components/properties/inputs/UseUniquePropertyValuesLoader.ts
```diff
@@ -0,0 +1,302 @@
...
+      return;
+    }
+
+    const timeout = setTimeout(() => {
+      if (itemsLoader?.needsLoadingItems(filterText)) {
```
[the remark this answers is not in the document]
MartynasStrazdas: Removed needsLoadingItems

packages/components/src/presentation-components/properties/inputs/UseUniquePropertyValuesLoader.ts
```diff
@@ -0,0 +1,302 @@
...
+  private _hasMore = true;
+  private _disposed = false;
+  private _offset = 0;
+
+  constructor(
+    private _beforeLoad: () => void,
+    private _onItemsLoaded: (newItems: T[]) => void,
+    private _loadItems: (offSet: number) => Promise<LoadedItems<T>>,
+    private _getOptionLabel: (option: T) => string,
+    private _initialSelectedValues?: string[],
```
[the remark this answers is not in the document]
MartynasStrazdas: Removed needsLoadingItems

packages/components/src/presentation-components/properties/inputs/UseUniquePropertyValuesLoader.ts
```diff
@@ -0,0 +1,302 @@
...
+import { SelectOption } from "@itwin/itwinui-react";
+import { DisplayValue, DisplayValueGroup, Field, FieldDescriptor, Keys, KeySet, Ruleset } from "@itwin/presentation-common";
+import { Presentation } from "@itwin/presentation-frontend";
+import { translate, UniqueValue } from "../../common/Utils";
+
+/** @internal */
+export const UNIQUE_PROPERTY_VALUES_BATCH_SIZE = 100;
+
+/** @internal */
+export const FILTER_WARNING_OPTION = { label: "Too many values please use filter", value: "__filter__", disabled: true };
```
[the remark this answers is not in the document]
MartynasStrazdas: Moved it into another file, it still relies on the `UNIQUE_PROPERTY_VALUES_BATCH_SIZE`, if we decide to reuse it later we can add a private batch property.

packages/components/src/presentation-components/properties/inputs/UseNavigationPropertyTargetsLoader.ts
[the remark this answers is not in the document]
saskliutas: Does not look like it is used anywhere outside this file.

packages/components/src/presentation-components/properties/inputs/UseNavigationPropertyTargetsLoader.ts
```diff
@@ -42,53 +50,82 @@ export interface NavigationPropertyTargetsResult {
...
+  }, [imodel, initialSelectedTarget, ruleset]);
+
+  // On filter text change, check if more items need to be loaded
+  useEffect(() => {
+    if (!itemsLoader) {
+      return;
+    }
+
+    const timeout = setTimeout(() => {
+      if (itemsLoader?.needsLoadingItems(filterText)) {
```
saskliutas, line 102: I think `?` is not needed anymore. Also `needsLoadingItems` could be moved into `loadItems` same as with unique values loader.

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
```diff
@@ -49,98 +35,84 @@ export interface UniquePropertyValuesSelectorProps {
...
+  }, []);
 
   return (
-    <AsyncSelect
-      inputValue={searchInput}
+    <ComboBox
+      multiple={true}
+      enableVirtualization={true}
+      options={selectOptions}
+      onChange={(newValues) => onValueChange(newValues)}
```
saskliutas, line 90:
```suggestion
      onChange={onValueChange}
```

packages/components/src/presentation-components/properties/inputs/NavigationPropertyTargetSelector.tsx
```diff
@@ -58,46 +67,37 @@ export const NavigationPropertyTargetSelector = forwardRef<PropertyEditorAttribu
...
-        }}
-        menuPortalTarget={portalTarget}
-        {...menuProps}
-      />
-    </div>
+    <ComboBox
+      multiple={false}
+      enableVirtualization={true}
+      options={selectOptions}
+      onChange={(newValue) => onChange(newValue)}
```
saskliutas, line 79:
```suggestion
      onChange={onChange}
```

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
```diff
@@ -0,0 +1,113 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { UNIQUE_PROPERTY_VALUES_BATCH_SIZE } from "./UseUniquePropertyValuesLoader";
```
saskliutas, line 6: I think items batch size should be defined in this file and used in unique value and navigation target loader/combobox

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
MartynasStrazdas: Moved

packages/components/src/presentation-components/properties/inputs/UseNavigationPropertyTargetsLoader.ts
MartynasStrazdas: Removed `needsLoadingItems`

packages/components/src/presentation-components/properties/inputs/UniquePropertyValuesSelector.tsx
MartynasStrazdas: changed

packages/components/src/presentation-components/properties/inputs/NavigationPropertyTargetSelector.tsx
MartynasStrazdas: changed

packages/components/src/presentation-components/properties/inputs/UseNavigationPropertyTargetsLoader.ts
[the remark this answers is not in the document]
MartynasStrazdas: Misread the comment, removed.

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
MartynasStrazdas: not sure what you mean by this change :D?

saskliutas: Somehow this comment ended up on the wrong line. 😕  It was meant for `ItemsLoader` constructo call were you are passing `() => spy` or `() => getItemsStub()` you can pass `spy` or `getItemsStub` directly. Same for other tests in this file and `NavigationPropertyTargetsLoader` tests:

```ts
const itemsLoader = new ItemsLoader<string>(
        () => {},
        spy,
        getItemsStub,
        (option) => option,
        [],
      );
```

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
```diff
@@ -0,0 +1,114 @@
...
+    this._beforeLoad();
+
+    const { options, hasMore, length } = await this.loadFilteredItems(filterText);
+
+    if (this._disposed) {
+      return;
+    }
+
+    this._loadedItems.push(...options);
+    this._offset += length;
```
saskliutas, line 87: Looks like there is an issue with offset tracking. `loadFilteredItems` starts with `this._offset` and increases it based on new loaded items. Later it returns that computed offset as `length` and it is added here. It will double `this._offset`.

packages/components/src/presentation-components/properties/inputs/UseNavigationPropertyTargetsLoader.ts
```diff
@@ -42,53 +46,80 @@ export interface NavigationPropertyTargetsResult {
...
+  }, [imodel, initialSelectedTarget, ruleset]);
+
+  // On filter text change, check if more items need to be loaded
+  useEffect(() => {
+    if (!itemsLoader) {
+      return;
+    }
+
+    const timeout = setTimeout(() => {
+      void itemsLoader?.loadItems(filterText);
```
saskliutas, line 98:
```suggestion
      void itemsLoader.loadItems(filterText);
```

packages/components/src/presentation-components/properties/inputs/UseUniquePropertyValuesLoader.ts
```diff
@@ -0,0 +1,185 @@
...
+  }, [imodel, ruleset, field, descriptorInputKeys, initialSelectedValues]);
+
+  // On filter text change, check if more items need to be loaded
+  useEffect(() => {
+    if (!filterText || !itemsLoader) {
+      return;
+    }
+
+    const timeout = setTimeout(() => {
+      void itemsLoader?.loadItems(filterText);
```
saskliutas, line 82:
```suggestion
      void itemsLoader.loadItems(filterText);
```

saskliutas (changes requested):

packages/components/src/presentation-components/common/PortalTargetContext.tsx
```diff
@@ -33,11 +33,3 @@ export interface PortalTargetContext {
 }
 
 const portalTargetContext = createContext<PortalTargetContext>({ portalTarget: null });
-
-/**
- * Returns context provided by [[PortalTargetContextProvider]].
- * @internal
- */
-export function usePortalTargetContext() {
```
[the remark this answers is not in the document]
MartynasStrazdas: Deprecated, removed from test-app

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
MartynasStrazdas: changed

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
MartynasStrazdas: fixed

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const needsMoreItems = (options: T[]): boolean => {
+      matchingItems.push(...options.filter((option) => this._getOptionLabel(option).toLowerCase().includes(filterText.toLowerCase())));
+      return matchingItems.length < VALUE_BATCH_SIZE;
+    };
+
+    await this.loadUniqueItems(needItems, needsMoreItems);
+  }
+
+  private async loadUniqueItems(needsItems: () => boolean, needsMoreItems: (options: T[]) => boolean) {
```
saskliutas, line 84: Looks like instead of `needsItems` you can call `needsMoreItems(this._loadedItems)`.

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const needsItems = (): boolean => {
+      if (this._loadedItems.length > 0) {
+        filterInitialSelectedValues(this._loadedItems);
+      }
+
+      return this._loadedItems.length < VALUE_BATCH_SIZE && (initialSelectedValues?.length === 0 || initialSelectedValues?.length !== matchingItems.length);
+    };
+
+    const filterInitialSelectedValues = (items: T[]) => {
+      matchingItems.push(
```
saskliutas, line 49: This can be moved to `needsMoreItems`

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
MartynasStrazdas: It is also being used in `needsItems`

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
MartynasStrazdas: Changed it to pass needsMoreItems instead and removed `needsItems` from `loadInitialItems`

saskliutas: That's not exactly what I meant. I think `loadUniqueItems` function can take only `needsMoreItems` callback and call it instead of `needsItems()` like `needsMoreItems(this._loadedItems)`

MartynasStrazdas: `loadInitialItems` also calls loadUniqueItems and merging `needsItems` with `needsMoreItems` would be pretty hard because their conditions don't match (needsItems check that if for some reason some items are loaded perhaps we don't need to load more, which would not work with `needsMoreItems`)

saskliutas: Wouldn't something like this work:
```ts
const matchingItems: T[] = [];
const needsItemsLoaded = (items: T[]) => {
  if (initialSelectedValues && initialSelectedValues.length > 0) {
    matchingItems.push(
      ...items.filter((item) => {
        return initialSelectedValues.some((initialSelectedValue) => initialSelectedValue === this._getOptionLabel(item));
      }),
    );
    return matchingItems.length === initialSelectedValues.length;
  }
  matchingItems.push(...items);
  return matchingItems.length < VALUE_BATCH_SIZE;
}
```

MartynasStrazdas: I don't think so, you would only want to have the exact items that have the same label in the matching items as you check for ===, but in the else we would just push all of them, we would also need to check against already loaded items if we want to have the initial load. I think it's possible to merge `needsItems` and `needsMoreItems`, but it would be inefitient as some if statements which do not matter for while cycle would be hit and the merged `needsMoreItems` would be unnecessary complicated.

MartynasStrazdas: changed

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
```diff
@@ -0,0 +1,97 @@
...
+        return matchingItems.length < initialSelectedValues.length;
+      }
+      return items.length < VALUE_BATCH_SIZE;
+    };
+
+    await this.loadUniqueItems(needsItemsLoaded);
+  }
+
+  public async loadItems(filterText?: string) {
+    if (!filterText) {
```
saskliutas, line 52: I would move this into `needsItemsLoaded` to make sure that if `filterText` is not specified and no items are loaded we would load atleast one batch of items.

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
```diff
@@ -0,0 +1,97 @@
...
+    let currOffset = this._offset;
+    let hasMore = this._hasMore;
+
+    if (!this._hasMore || this._isLoading) {
+      return;
+    }
+
+    if (!needsItemsLoaded(this._loadedItems)) {
+      return;
+    }
```
saskliutas, line 75: These could be merged

packages/components/src/test/properties/inputs/UniquePropertyValuesSelector.test.tsx
```diff
@@ -1395,4 +1253,107 @@ describe("UniquePropertyValuesSelector", () => {
       });
     });
   });
+  describe("Items loader", () => {
```
saskliutas, line 1256:
```suggestion
  describe("ItemsLoader", () => {
```

saskliutas (approved):

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
MartynasStrazdas: changed

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
MartynasStrazdas: changed

.changeset/nine-ducks-wink.md
MartynasStrazdas: Rewritten.

packages/components/src/presentation-components/properties/inputs/ItemsLoader.tsx
```diff
@@ -0,0 +1,93 @@
...
+      return items.length < VALUE_BATCH_SIZE;
+    };
+
+    await this.loadUniqueItems(needsItemsLoaded);
+  }
+
+  public async loadItems(filterText?: string) {
+    const needsItemsLoaded = (options: T[]): boolean => {
+      if (!filterText) {
+        return options.length === 0;
```
saskliutas, line 54:
```suggestion
        return options.length < VALUE_BATCH_SIZE;
```

MartynasStrazdas: If there are already loaded values we probably don't want to load more, because it already loaded up to the VALUE_BATCH_SIZE or there are not enough items to load.

saskliutas: If there are not enough items to load this won't be called as `ItemsLoader` is tracking if there are more items to load. This changes does not change behavior. In my opinion it just makes both cases (wiht\without filterTest) the same - load items if there are less than batch size.

.changeset/nine-ducks-wink.md
```diff
@@ -2,4 +2,17 @@
...
+This affects [`PresentationInstanceFilterBuilder`, `NavigationPropertyEditor`] components.
+
+Changes:
+
+- The number of select options is limited to 100. When more items exist, a non-selectable option is displayed at the bottom of the list, prompting users to provide an items filter. Previously, additional pages of select options was loaded when user scrolled to the bottom of the list.
+
+- Old style:
+    ![alt text](image-1.png)
+    New style:
+    ![alt text](image.png)
```
grigasp, line 16: - Formatting of this won't look good when the markdown is rendered. I think we could show this as a table (e.g. like [this](https://github.com/iTwin/viewer-components-react/blob/dc644c1d291c175cad84bb04513efc84c54413d2/packages/itwin/tree-widget/README.md?plain=1#L23-L25)).

- Do we know what will happen with the images upon release, when these changesets are merged into the changelog? I don't think we have any code in place to make sure the images are copied-over, or even know where they should land.

MartynasStrazdas: Removed images altogether

Issue #737, opened 2024-10-11
`components`: Lots of clicks required to edit a numeric value with units


grigasp:
At the moment a user has to make several clicks to edit a numeric KoQ value:
- first click selects the property row
- second click enters editing mode, but the editor is not focused
- third click puts the cursor
- user edits the value
- fourth click or Enter key press exits the edit mode

At the minimum, when entering editor mode, we should:
1. automatically focus the editor,
2. select all content to allow easily replacing it.

We should also consider entering edit mode on first click, if the click is on the property value and not the label.

Issue #736, opened 2024-10-11
`components`: Editor's placeholder value is confusing


grigasp:
At the moment our KoQ editor uses `12.34` as the placeholder value in persistence units. However, that value is formatted before it's shown as the placeholder, and the formatted value might be quite confusing, e.g.:
![Image](https://github.com/user-attachments/assets/4044e331-8870-4b74-9240-c84bc3ca678c)

One way to improve the UX could be to use the initial/persisted value as the placeholder, if there is one.

PR #747, opened 2024-10-24
KOQ and numeric editor improvements

MartynasStrazdas:
Closes https://github.com/iTwin/presentation/issues/737
Closes https://github.com/iTwin/presentation/issues/736
Closes https://github.com/iTwin/presentation/issues/748
After:
Before click
image
After click
image
Placeholder message is based on default input if it exists:
image


saskliutas:
Have you checked how this affects UX in filter builder dialog?

MartynasStrazdas:
> Have you checked how this affects UX in filter builder dialog?

image
Adding a new condition will auto-focus to the input.
Inputting value and saving, then opening would not auto-focus. 
Clicking (on focus) input will select all the text inside the input.

So far no unwanted behavior, will keep in mind to check filter after numeric input changes.

saskliutas:
> > Have you checked how this affects UX in filter builder dialog?
> 
> image Adding a new condition will auto-focus to the input. Inputting value and saving, then opening would not auto-focus. Clicking (on focus) input will select all the text inside the input.
> 
> So far no unwanted behavior, will keep in mind to check filter after numeric input changes.

What if you open filter dialog with initial filter that contains multiple rules with KOQ values?

grigasp:
> Adding a new condition will auto-focus to the input.

I don't understand this. When you add a new condition, the value input is not even rendered, right? So I'd expect either the whole row, or the property picker to be focused. After selecting the property, you get to select the operator, at which point the value input is rendered. I guess it would be okay for it to auto-focus at that point.

MartynasStrazdas:
> > Adding a new condition will auto-focus to the input.
> 
> I don't understand this. When you add a new condition, the value input is not even rendered, right? So I'd expect either the whole row, or the property picker to be focused. After selecting the property, you get to select the operator, at which point the value input is rendered. I guess it would be okay for it to auto-focus at that point.
I meant after selecting the operator :D sorry for the confusion

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
```diff
@@ -33,13 +33,14 @@ export interface UseQuantityValueInputProps {
...
   koqName: string;
+  isReadOnly?: boolean;
 }
 
 /**
  * Custom hook that manages state for quantity values input.
  * @internal
  */
-export function useQuantityValueInput({ initialRawValue, schemaContext, koqName }: UseQuantityValueInputProps) {
+export function useQuantityValueInput({ initialRawValue, schemaContext, koqName, isReadOnly }: UseQuantityValueInputProps) {
```
saskliutas, line 43: I don't think this hook need to take `isReadOnly` flag. `isReadOnly` flag should be handled where UI itself is rendered.

packages/components/src/presentation-components/properties/inputs/NumericPropertyInput.tsx
```diff
@@ -70,14 +56,25 @@ function getInputTargetFromPropertyRecord(propertyRecord: PropertyRecord) {
 }
 
 /** @internal */
-export interface NumericInputProps {
+export interface NumericInputProps extends PropertyEditorProps {
   onChange: (newValue: string) => void;
   onBlur?: React.FocusEventHandler;
   value: string;
+  isReadOnly?: boolean;
```
saskliutas, line 63: I would rename to `isDisabled` or `disabld`.

saskliutas (changes requested):

packages/components/src/presentation-components/properties/inputs/NumericPropertyInput.tsx
MartynasStrazdas: Renamed

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
MartynasStrazdas: Moved

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
```diff
@@ -75,7 +75,7 @@ export function useQuantityValueInput({ initialRawValue, schemaContext, koqName
         placeholder: newPlaceholder,
       };
     });
-  }, [formatter, parser]);
+  }, [formatter, initialRawValue, parser]);
```
saskliutas, line 78: I think we should store `initialRawValue` in `useRef` for this to avoid rerenders if different initial value is provided.

packages/components/src/presentation-components/properties/inputs/QuantityPropertyEditorInput.tsx
```diff
@@ -72,7 +72,24 @@ const QuantityPropertyValueInput = forwardRef<PropertyEditorAttributes, Quantity
...
+    }, [setFocus]);
+
+    return (
+      <Input
+        size="small"
+        {...inputProps}
+        disabled={propertyRecord.isReadonly || inputProps.disabled}
+        ref={inputRef}
+        onBlur={onBlur}
+        onFocus={() => inputRef.current?.setSelectionRange(0, 9999)}
```
saskliutas, line 90: Why not `inputRef.current?.setSelectionRange(0, inputProps.value.length)`?

packages/components/src/test/properties/inputs/NumericPropertyInput.test.tsx
```diff
@@ -267,6 +274,18 @@ describe("<NumericInput />", () => {
     expect(spy).to.be.be.calledOnce;
   });
 
+  it("should focus on input if setFocus is true", async () => {
+    const { getByRole } = render(<NumericInput onChange={() => {}} value="1" setFocus={true} />);
+    let inputFocused = false;
+
+    const input = await waitFor(() => getByRole("textbox"));
+    input.addEventListener("focusout", (event) => {
```
saskliutas, line 282: Would it be possible to use `document.activeElement` to check if input is focused instead of listening for this event?

saskliutas (changes requested):

packages/components/src/test/properties/inputs/NumericPropertyInput.test.tsx
MartynasStrazdas: While for numeric property input it does work, for the quantity input it returns it with body, which should not be the case.

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
MartynasStrazdas: Changed

packages/components/src/presentation-components/properties/inputs/QuantityPropertyEditorInput.tsx
MartynasStrazdas: Changed

packages/components/src/test/properties/inputs/NumericPropertyInput.test.tsx
saskliutas: If `activeElement` is set to `body` does that mean the input is not focused?

MartynasStrazdas: Adjusted.

packages/components/src/presentation-components/properties/inputs/QuantityPropertyEditorInput.tsx
MartynasStrazdas: Reverted changed for `QuantityPropertyEditor`, the value is provided from a custom hook meaning the value is 0 (length) on first load when we focus on the component, left the adjusted selection for the numerical value. I think it is fine to use constant value for selection range as trying to use raw value + const margin might couse bugs.

saskliutas (approved):

packages/components/src/presentation-components/properties/inputs/QuantityPropertyEditorInput.tsx
MartynasStrazdas: changed

Issue #745, opened 2024-10-24
Upgrade to TS 5.6


grigasp:
see https://github.com/iTwin/itwinjs-core/pull/7232

Issue #744, opened 2024-10-24
Upgrade ESLint to version 9


grigasp:
see https://github.com/iTwin/itwinjs-core/pull/7233

PR #753, opened 2024-10-25
Bump Eslint & Typescript

saskliutas:
Closes https://github.com/iTwin/presentation/issues/745
Closes https://github.com/iTwin/presentation/issues/744


apps/full-stack-tests/src/unified-selection/SelectionScope.test.ts
```diff
@@ -33,7 +33,6 @@ describe("SelectionScope", () => {
         cacheDir: path.join(__dirname, ".cache", `${process.pid}`),
       },
     });
-    // eslint-disable-next-line @itwin/no-internal
     RpcManager.registerImpl(ECSchemaRpcInterface, ECSchemaRpcImpl);
```
grigasp, line 36: I'm surprised ESLint says the rule is not necessary - both `ECSchemaRpcInterface` and `ECSchemaRpcImpl` are internal. @wgoehrig, have you noticed anything like this in the core repo? Any ideas what made the rule think this is not usage of internal API?

saskliutas: Looks like `no-internal` rule only catches internal function/method invocation or internal class object creation. Filled an issue in eslint-plugin repo: https://github.com/iTwin/eslint-plugin/issues/84

grigasp (approved):

PR #765, opened 2024-11-04
Unified selection: fix transient elements support

saskliutas:
Class name used in transient element instance key does not follow ECSchema class name formats: Schema:Class or Schema.Class.
Added exception for transient element instance key when normalizing class names.

### DIFF
--- a/packages/unified-selection/src/unified-selection/Selectable.ts
+++ b/packages/unified-selection/src/unified-selection/Selectable.ts
@@ -79,7 +79,10 @@ export interface Selectables {
   custom: Map<string, CustomSelectable>;
 }
 
-/** @internal */
+/**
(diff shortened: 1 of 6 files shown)


packages/unified-selection/src/unified-selection/Selectable.ts
```diff
@@ -79,6 +79,9 @@ export interface Selectables {
   custom: Map<string, CustomSelectable>;
 }
 
+/** @internal */
+export const TRANSIENT_ELEMENT_CLASSNAME = "/TRANSIENT";
```
saskliutas, line 83: @grigasp I think this should be exported, made public and used in `presentation-frontend`. What do you think?

grigasp: agree

grigasp (approved):

PR #770, opened 2024-11-07
Update modelsTreeDefinition from viewer-components-react

JonasDov:
Changes from viewer-components-react: https://github.com/iTwin/viewer-components-react/pull/1084


packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
```diff
@@ -611,120 +617,140 @@ function createCategoryInstanceKeyPaths(categoryId: Id64String, idsCache: Models
...
           pe.Model.Id,
-          json_insert(
-            ce.Path,
-            '$[#]', ${ECSql.createInstanceKeySelector({ alias: "pe" })},
-            '$[#]', IIF(pe.Parent.Id IS NULL, ${ECSql.createInstanceKeySelector({ alias: "c" })}, NULL),
-            '$[#]', IIF(pe.Parent.Id IS NULL, ${ECSql.createInstanceKeySelector({ alias: "m" })}, NULL)
+          ce.GroupingNodeIndex,
+          IIF(pe.Parent.Id IS NULL,
+            ec_classname([m].[ECClassId], 's.c') || '${separator}' || CAST(IdToHex([m].[ECInstanceId]) AS TEXT) || '${separator}' || ec_classname([c].[ECClassId], 's.c') || '${separator}' || CAST(IdToHex([c].[ECInstanceId]) AS TEXT) || '${separator}' || ec_classname([pe].[ECClassId], 's.c') || '${separator}' || CAST(IdToHex([pe].[ECInstanceId]) AS TEXT) || '${separator}' || ce.Path,
+            ec_classname([pe].[ECClassId], 's.c') || '${separator}' || CAST(IdToHex([pe].[ECInstanceId]) AS TEXT) || '${separator}' || ce.Path
```
grigasp, line 674: I'm not a big fan of such string concatenations... [Your comment](https://github.com/iTwin/viewer-components-react/pull/1084#issue-2641797375) says this way is slightly faster than JSON, but do we know why exactly? Maybe it's just because this is less verbose? In that case, we could make JSON less verbose too:

- Instead of using instance key objects in an array, we could use tuples - that would remove "className" and "id" strings from the output (`{"className":"x", "id":"y"}` becomes `["x", "y"]`).

- Maybe we don't need to exact full class name? I think it would be enough for us to know if it's model, category or element id. So instead of adding full class name, we could add "m", "c", "e" strings and, when parsing the path, convert them to "BisCore.Model", "BisCore.SpatialCategory", "BisCore.GeometricElement3d".

grigasp (commented):
Do you think it would be possible to add a performance test that highlights this improvement? We already have some "models tree" tests, so should be just a matter of using a larger dataset and coming up with a list of target instance keys (maybe just query all geometric element ids to get that list?)

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
```diff
@@ -711,15 +712,25 @@ function createGeometricElementInstanceKeyPaths(
   );
 }
 
-function parseQueryRow(row: ECSqlQueryRow, groupInfos: ElementsGroupInfo[], separator: string) {
+function parseQueryRow(row: ECSqlQueryRow, groupInfos: ElementsGroupInfo[], separator: string, eClassName: string) {
```
grigasp, line 708: I'd rename `eClassName` to `elementClassName`

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
JonasDov: Applied changes, they helped improve the performance. However, string concatenation still outperformed json_array version by about 1 second. Keeping string concatenation.

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
JonasDov: Renamed

JonasDov:
> Do you think it would be possible to add a performance test that highlights this improvement? We already have some "models tree" tests, so should be just a matter of using a larger dataset and coming up with a list of target instance keys (maybe just query all geometric element ids to get that list?)

Added a performance test and tested it with new implementation and old implementation. These are the results when running locally:
Number of target items | Original implementation (time for initial load of tree) | New implementation (time for initial load of tree)
-- | -- | --
500 | 33.38 s | 172 ms
50k | ? | 1.73 s

apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
```diff
@@ -16,17 +15,30 @@ import {
...
   getHierarchyFactory(imodelAccess: ECSchemaProvider & ECClassHierarchyInspector): HierarchyDefinition;
   filtering?: {
     paths: HierarchyFilteringPath[];
   };
 }
+export type ProviderOptionsWithIModel = { iModel: IModelDb } & ProviderOptionsBase;
+
+export type ProviderOptionsWithIModelAccess = { imodelAccess: IModelAccess } & ProviderOptionsBase;
+
+export type ProviderOptions = ProviderOptionsWithIModel | ProviderOptionsWithIModelAccess;
```
grigasp, line 41: do we even need these to be exported? from the first look it seems they should only be needed here (the `setup` functions probably don't need to specify the return type).

apps/performance-tests/src/hierarchies/ModelsTree.test.ts
```diff
@@ -37,4 +37,28 @@ describe("models tree", () => {
       expect(result).to.be.greaterThan(0);
     },
   });
+
+  run<{ iModel: SnapshotDb; imodelAccess: IModelAccess; targetItems: Array<InstanceKey> }>({
+    testName: "gets 50k filter paths using ModelsTreeDefinition and does initial load",
```
grigasp, line 42:
```suggestion
    testName: "creates initial filtered view for 50k target items",
```

apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
JonasDov: If return type of `setup` is not specified, then vs code complains:
image

JonasDov: It seems simpler to just export those types and specify the return type of `setup`, or do you still think we should not export them?

grigasp: Short term it may be simpler, but we're littering intellisense with a number of useless types, which will make it more complicated in the future.

The build problems can be fixed:
- remove `rowLimit` from the object returned by `setup`,
- define types for `getHierarchyFactory`:
  ```ts
  getHierarchyFactory: (imodelAccess: ECSchemaProvider & ECClassHierarchyInspector): HierarchyDefinition => ({ ... 
  ```
- supply `rowLimit` when creating `StatelessHierarchyProvider`:
  ```ts
  const provider = new StatelessHierarchyProvider({
    ...props,
    rowLimit: "unbounded",
  });
  ```

JonasDov: Applied suggestion

apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
```diff
@@ -71,30 +95,35 @@ export class StatelessHierarchyProvider {
     });
   }
 
-  private createECSchemaProvider() {
-    const iModel = this._props.iModel;
+  private static createNewECSchemaProvider(iModel: IModelDb) {
```
grigasp, line 98: why was this renamed? The "new" word seems redundant - "create" already says it's going to be new

apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
```diff
@@ -71,30 +95,35 @@ export class StatelessHierarchyProvider {
...
+      "iModel" in this._props ? StatelessHierarchyProvider.getIModelAccess(this._props.iModel, this._props.rowLimit) : this._props.imodelAccess;
     return createIModelHierarchyProvider({
       imodelAccess,
       hierarchyDefinition: this._props.getHierarchyFactory(imodelAccess),
       queryCacheSize: 0,
       filtering: this._props.filtering,
     });
   }
+
+  public static getIModelAccess(iModel: IModelDb, rowLimit?: number | "unbounded"): IModelAccess {
```
grigasp, line 116: IMO this should be named "create" rather than "get"

apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
JonasDov: Renamed back to `createECSchemaProvider`

apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
JonasDov: Renamed

grigasp (approved):

PR #783, opened 2024-11-18
Hierarchies: Custom hierarchy providers learning

grigasp:
Closes https://github.com/iTwin/presentation/issues/749

Additionally, added an API that helps with hierarchy filtering implementation. Refactored imodel hierarchy provider to use that API and filtering performance improved by ~25%:

| Benchmark suite | Current: 4f50cafc12c043cafa467d7341640cfff8a9eb2d | Previous: 9374049632acf91de7018d02b49660b11e53caa6 | Deviation | Status |
|-|-|-|-|-|
| `filtering filters with 50000 paths` | `9974.49` ms | `13770.6` ms | `-27.5668%` | ✅ |

### DIFF
--- a/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
+++ b/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
@@ -3,9 +3,10 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { NonGroupingHierarchyNode } from "./HierarchyNode.js";
-import { HierarchyNodeIdentifiersPath } from "./HierarchyNodeIdentifier.js";
-import { GroupingNodeKey } from "./HierarchyNodeKey.js";
+import { filter, firstValueFrom, from, map, mergeMap, toArray } from "rxjs";
+import { HierarchyNode, NonGroupingHierarchyNode } from "./HierarchyNode.js";
+import { HierarchyNodeIdentifier, HierarchyNodeIdentifiersPath } from "./HierarchyNodeIdentifier.js";
+import { GenericNodeKey, GroupingNodeKey, HierarchyNodeKey, InstancesNodeKey } from "./HierarchyNodeKey.js";
 
 /** @public */
 export interface FilterTargetGroupingNodeInfo {
@@ -45,7 +46,7 @@ export namespace HierarchyFilteringPath {
    */
   export function normalize(source: HierarchyFilteringPath): Exclude<HierarchyFilteringPath, HierarchyNodeIdentifiersPath> {
     if (Array.isArray(source)) {
-      return { path: source, options: undefined };
+      return { path: source };
     }
     return source;
   }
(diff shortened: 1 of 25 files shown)


packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -113,3 +128,154 @@ export function extractFilteringProps(
...
+type NormalizedFilteringPath = ReturnType<(typeof HierarchyFilteringPath)["normalize"]>;
+class MatchingFilteringPathsReducer {
+  private _filteredChildrenIdentifierPaths = new Array<NormalizedFilteringPath>();
+  private _isFilterTarget = false;
+  private _filterTargetOptions = undefined as HierarchyFilteringPathOptions | undefined;
+  private _needsAutoExpand = false;
+
+  public accept({ path, options }: NormalizedFilteringPath) {
+    if (path.length === 1) {
+      this._isFilterTarget ||= true;
```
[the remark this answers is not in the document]
grigasp: Definitely would - fixed.

packages/hierarchies/src/hierarchies/HierarchyProvider.ts
```diff
@@ -49,10 +49,19 @@ export interface HierarchyProvider {
    */
   readonly hierarchyChanged: Event<() => void>;
 
-  /** Gets nodes for the specified parent node. */
+  /**
+   * Gets nodes for the specified parent node. This is THE method to implement, otherwise
```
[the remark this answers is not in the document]
grigasp: Agree, changed.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -109,248 +111,149 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
-    for (const filteredChildrenNodeIdentifierPath of filteredChildrenNodeIdentifierPaths) {
-      const { path, options } = HierarchyFilteringPath.normalize(filteredChildrenNodeIdentifierPath);
-      /* c8 ignore next 3 */
-      if (path.length === 0) {
-        continue;
+          ? await (async () => {
+              const rowInstanceKey = { className: row[ECSQL_COLUMN_NAME_FilterClassName], id: row[ECSQL_COLUMN_NAME_FilterECInstanceId] };
+              return filteringHelper.createChildNodePropsAsync({
+                // eslint-disable-next-line @typescript-eslint/promise-function-async
+                pathMatcher: (identifier) => {
```
JonasDov, line 123: Can't we do this:
`pathMatcher: (identifier): boolean | Promise<boolean> => {`
instead of adding ignore comment?

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -109,248 +111,149 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+    const [genericNodeDefinitions, instanceNodeDefinitions] = partition(sourceDefinitions, HierarchyNodesDefinition.isGenericNode);
+    return firstValueFrom(
+      merge(
+        genericNodeDefinitions.pipe(
+          map((definition) => {
+            if (
+              filteringHelper.hasFilterTargetAncestor ||
+              childNodeFilteringIdentifiers
+                .filter((id) => HierarchyNodeIdentifier.isGenericNodeIdentifier(id) && (!id.source || id.source === this._imodelAccess.imodelKey))
+                .some(({ id }) => id === definition.node.key)
```
JonasDov, line 173: Why do we first filter, and then do `some`. I think we can immediatly do `some`. For example:
```ts
 if (
    filteringHelper.hasFilterTargetAncestor ||
    childNodeFilteringIdentifiers.some(
      (id) =>
        HierarchyNodeIdentifier.isGenericNodeIdentifier(id) &&
        (!id.source || id.source === this._imodelAccess.imodelKey) &&
        id.id === definition.node.key,
    )
  ) {
            ```

File: packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts [hunk 5]
```diff
@@ -109,248 +111,149 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+    };
   }
 
   public async defineHierarchyLevel(props: DefineHierarchyLevelProps): Promise<HierarchyLevelDefinition> {
     const sourceDefinitions = await this._source.defineHierarchyLevel(props);
-    const filteringProps = extractFilteringProps(this._nodeIdentifierPaths, props.parentNode);
-    if (!filteringProps) {
+
+    const filteringHelper = createHierarchyFilteringHelper(this._nodeIdentifierPaths, props.parentNode);
+    const childNodeFilteringIdentifiers = filteringHelper.getChildNodeFilteringIdentifiers();
```
JonasDov, line 159: We are getting `childNodeFilteringIdentifiers` and later we are iterating over this array twice: once for `genericNodeDefinitions`, and once for `instanceNodeDefinitions`. Maybe we could split this array into two after checking if it's not undefined. In that case we would iterate only once over this array.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -109,248 +111,149 @@ export class FilteringHierarchyDefinition implements HierarchyDefinition {
...
+                if (id.imodelKey && id.imodelKey !== imodelAccess.imodelKey) {
+                  continue;
+                }
+                if (id.className !== definition.fullClassName) {
+                  if (
+                    !(await Promise.all([
+                      imodelAccess.classDerivesFrom(id.className, definition.fullClassName),
+                      imodelAccess.classDerivesFrom(definition.fullClassName, id.className),
+                    ]).then(([isDerivedFrom, isDerivedTo]) => isDerivedFrom || isDerivedTo))
+                  ) {
```
JonasDov, line 233: Do you think it's better to have nested ifs instead of a single if:
```ts
if (
  id.className !== definition.fullClassName &&
  !(await Promise.all([
    imodelAccess.classDerivesFrom(id.className, definition.fullClassName),
    imodelAccess.classDerivesFrom(definition.fullClassName, id.className),
  ]).then(([isDerivedFrom, isDerivedTo]) => isDerivedFrom || isDerivedTo))
) {
  continue;
}
```

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
grigasp: I think you suggestion would slightly help if the split would divide the array in more-or-less equal parts. However, I predict that whenever we get many filtered node paths, absolute majority of them are going to be of one type. And in that case, iterating over it once per definition is better than iterating once to split + once per definition.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
grigasp: Yes, that's cleaner. Changed.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
grigasp: Agree, changed.

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
grigasp: Both aren't great, but maybe nested conditions are worse... Applied your suggestion.

JonasDov (approved):

grigasp:
@MartynasStrazdas @saskliutas, any comments?

saskliutas (approved):

PR #795, opened 2024-11-28
`hierarchies-react`: adjust tree node loading indicator in small tree

saskliutas:
Adjusted size of tree node loading indicator in small tree to match other icons.
Before:
image

After:
image

### DIFF
--- a/.changeset/afraid-ducks-listen.md
+++ b/.changeset/afraid-ducks-listen.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Fix tree node loading indicator to match other icons size.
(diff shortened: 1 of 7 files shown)


.changeset/afraid-ducks-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
```
MartynasStrazdas, line 2: Should it not be a minor version bump because a new property was added to `TreeNodeRenderer`?

MartynasStrazdas (approved):

.changeset/afraid-ducks-listen.md
saskliutas: Added change file for new prop.

PR #798, opened 2024-12-10
Reduce main thread blocking when filtering Models tree

JonasDov:
Changes copied from https://github.com/iTwin/viewer-components-react/pull/1123


packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
```diff
@@ -740,40 +774,72 @@ async function createInstanceKeyPathsFromTargetItems({
...
+          if (keyParsed.type === 2) {
+            ids.models.push(keyParsed.key);
+            continue;
+          }
+          if (keyParsed.type === 3) {
+            ids.categories.push(keyParsed.key);
+            continue;
+          }
+          ids.elements.push(keyParsed.key);
+        }
```
grigasp, line 827: why not have a `reduce` that creates the `ids` structure in the first place, instead of `toArray` to create an array and then convert it to `ids`?

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
```diff
@@ -740,40 +774,72 @@ async function createInstanceKeyPathsFromTargetItems({
...
+      mergeMap(async (key): Promise<{ key: string; type: number } | { key: ElementsGroupInfo; type: 0 }> => {
+        if ("parent" in key) {
+          return { key, type: 0 };
+        }
+
+        const subjectCheck = imodelAccess.classDerivesFrom(key.className, "BisCore.Subject");
+        const isSubject = subjectCheck instanceof Promise ? await subjectCheck : subjectCheck;
+        if (isSubject) {
+          return { key: key.id, type: 1 };
+        }
```
grigasp, line 789: Does the `instanceof Promise` check have any value over simply doing `const isSubject = await imodelAccess.classDerivesFrom(key.className, "BisCore.Subject");`?

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
```diff
@@ -740,40 +774,72 @@ async function createInstanceKeyPathsFromTargetItems({
...
+
+        const elementsLength = ids.elements.length;
+        return collect(
+          merge(
+            from(ids.subjects).pipe(mergeMap((id) => createSubjectInstanceKeysPath(id, idsCache))),
+            from(ids.models).pipe(mergeMap((id) => createModelInstanceKeyPaths(id, idsCache))),
+            from(ids.categories).pipe(mergeMap((id) => createCategoryInstanceKeyPaths(id, idsCache))),
+            from(ids.elements).pipe(
+              bufferCount(Math.ceil(elementsLength / Math.ceil(elementsLength / 5000))),
+              releaseMainThreadOnItemsCount(2),
```
grigasp, line 837: Does this `releaseMainThreadOnItemsCount(2)` give any value? As far as I can tell, `createGeometricElementInstanceKeyPaths` doesn't do anything expensive, so it shouldn't be a problem if we call it many times from the same microtask. 

What could be a problem though, is `createGeometricElementInstanceKeyPaths` sending out `queryRows` requests - we don't want to send out too many of them at the same time. Maybe `mergeMap` should have a concurrency limit of ~10 to avoid that?

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
JonasDov: Changed

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
JonasDov: Well, it's tricky to say. If it has then it's very minor. I ran the test 50 times (one with the `instanceof` check and one without) and compared results, implementation with `instanceof` check was about 100ms faster on average. Then i did it again and it seemed to be just about 10 ms faster.
Since its not obvious that this helps I've removed the `instanceof` check.

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
JonasDov: You were right, removed

grigasp (approved):

PR #800, opened 2024-12-12
`unified-selection`: Support 5.0 selection set

grigasp:
This adds support for enhancements made in https://github.com/iTwin/itwinjs-core/pull/7478, and, together with the latter, resolves https://github.com/iTwin/viewer-components-react/issues/1099.


packages/unified-selection/src/test/EnableUnifiedSelectionSyncWithIModel.test.ts
```diff
@@ -297,16 +396,55 @@ describe("IModelSelectionHandler", () => {
...
           await waitFor(() => {
-            expect(selectionStorageStub.removeFromSelection.calledWith({ imodelKey: imodelAccess.key, source: "Tool", selectables: removedKeys })).to.be.true;
+            expect(selectionStorageStub.removeFromSelection).to.be.calledWith({ imodelKey: imodelAccess.key, source: "Tool", selectables: removedKeys });
           });
         },
       );
     });
 
-    it("replaces selection", async () => {
+    it("removes models/subcategories/elements collection from selection", async () => {
```
JonasDov, line 405: Does this test really check if its removed, or does it check if `removeFromSelection` is called with correct arguments?
Same with other tests

grigasp: This is a unit test of `IModelSelectionHandler` - it's not supposed to be checking other components, like `SelectionStorage`. The job of the handler is to call appropriate storage's method with appropriate arguments - that's what these tests verify.

JonasDov: So shouldn't we rename the tests, since they don't do exactly what they say they do?

grigasp: I think they do. From the perspective of `IModelSelectionHandler`, selection is defined through `SelectionStorage`, and we're checking that the handler correctly calls `removeFromSelection` on it, effectively removing stuff.

JonasDov (approved):

PR #804, opened 2024-12-13
`components`: Deprecate tree-related APIs

grigasp:


packages/components/src/presentation-components/tree/DataProvider.ts
```diff
@@ -108,6 +111,8 @@ export interface PresentationTreeDataProviderProps extends DiagnosticsProps {
 /**
  * Definitions of methods used by [[PresentationTreeDataProvider]] to get nodes' data.
  * @public
+ * @deprecated in 5.7. All tree-related APIs have been deprecated in favor of the new generation hierarchy
+ * building APIs (see https://github.com/iTwin/presentation/blob/33e79ee8d77f30580a9bab81a72884bda008db25/README.md#the-packages).
```
MartynasStrazdas, line 115: Is it fine to show a link to a specific branch instead to main, wouldn't the README become outdated after some time.

grigasp: That's the point - the README might change over time, e.g. the section that we're pointing to might be removed.

MartynasStrazdas (approved):

saskliutas (approved):
Should we deprecate `ControlledTree` in AppUI also or should we leave for a while as it might be used without presentation stuff?

grigasp:
> Should we deprecate `ControlledTree` in AppUI also or should we leave for a while as it might be used without presentation stuff?

Good point, I'd vote for deprecating it. @GerardasB, thoughts?

Issue #796, opened 2024-12-04
`hierarchies-react`: Tree hooks fail to load root nodes


MartynasStrazdas:
Providing `useUnifiedSelectionTree` or `useTree` with a provider that only knows how to `getNodes` will not load root nodes without providing it with a `setHierarchyFilter` that raises `hierarchyChanged` event .

related to https://github.com/iTwin/viewer-components-react/pull/1104

Issue #789, opened 2024-11-21
`hierarchies-react`: Consider debouncing `hierarchyChanged` events in `useTree` hooks


grigasp:
A hierarchy provider may raise a number of `hierarchyChange` events one after another, especially in case of merged hierarchy provider. `useTree` hooks should avoid triggering expensive operations after each of them and instead handle it in debounced manner.

PR #812, opened 2024-12-17
[Hierarchies-react]: fix initial root loading when no change event is raised

MartynasStrazdas:
closes https://github.com/iTwin/presentation/issues/796, closes https://github.com/iTwin/presentation/issues/789


packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -20,6 +20,11 @@ export class TreeActions {
   private _currentModel: TreeModel;
   private _reset = new Subject<void>();
 
+  private _nodeLoader: Subject<LoadNodesOptions>;
+  private _parentId?: string;
+  private _timeTracker?: TimeTracker;
+  private _initialRootNode?: TreeModelRootNode;
+
```
MartynasStrazdas, line 27: @grigasp, @saskliutas, @JonasDov could not find a better solution to provide the next/complete functions their props while keeping the `debounceWrapper` generic any suggestions?

MartynasStrazdas (commented):
Still working on figuring out how to reliably check results in tests (when the inner pipeline completes).

.changeset/smooth-tips-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Fixed `UseTree` to return initial root nodes when changed hierarchy event is never raised.
```
grigasp, line 5:
```suggestion
Fixed tree state hooks not returning root nodes when hierarchy provider doesn't raise the `onHierarchyChanged` event upon setting a hierarchy filter.
```

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -108,14 +108,15 @@ describe("TreeActions", () => {
       });
 
       const actions = createActions(model);
-      actions.reloadTree();
+      actions.reloadTree({ state: "keep" });
```
grigasp, line 111: unnecessary change?

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -725,6 +731,48 @@ describe("TreeActions", () => {
...
+        return createAsyncIterator([]);
+      });
+
+      const actions = createActions(model);
+
+      actions.reloadTree({ state: "discard" });
+      actions.reloadTree({ state: "keep" });
+      actions.reloadTree({ state: "discard" });
+
+      await new Promise((r) => setTimeout(r, 50));
```
grigasp, line 768: this needs to go away - relying to specific timeout will result in a flaky test sooner or later

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -66,26 +94,12 @@ export class TreeActions {
...
-        complete: () => {
-          this.onLoadingComplete(parentId);
-          timeTracker[Symbol.dispose]();
-        },
-      });
+    this._timeTracker = new TimeTracker((time) => this._onLoad(loadAction, time));
+    this._parentId = options.parent.id;
+    this._initialRootNode = initialRootNode;
+
+    this._nodeLoader.next(options);
```
grigasp, line 101: `loadSubTree` is used not just to reload the whole tree, but also when expanding nodes. It looks like the change might cause only the last requested node to expand if we call `expandNode` multiple times with different `parentId` arguments.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/Utils.ts
```diff
@@ -63,3 +64,23 @@ export function safeDispose(disposable: {} | { [Symbol.dispose]: () => void } |
     disposable[Symbol.dispose]();
   }
 }
+
+/** @internal */
+export function debounceWrapper<T, R>(pipeline: (value: T) => Observable<R>, next: (value: R) => void, complete: () => void): Subject<T> {
+  const subject = new Subject<T>();
+  subject
+    .pipe(
+      debounceTime(50),
```
grigasp, line 73: Would changing to `0` here stop it from working?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: I think, instead of setting the arguments as class members, you should send them to `_nodeLoader`. That way, the debouncing loader could ensure that only similar requests are debounced (e.g. requests for different parents aren't dropped).

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
MartynasStrazdas: Yes, reverted.

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
MartynasStrazdas: The intention was to have the tests, before figuring out how to consistently check the results. Removed now.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/Utils.ts
MartynasStrazdas: It seems to work (no double reload at initial load, tests pass). My only problem with it would be that if there are enough reload calls this would fail and reload multiple time as there is some code before this, but I guess with smaller wait debounce time it should improve overall performance, changed to 0. PS. could not find if the 0 defaults to 1ms or something else.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: Moved inside TreeActions.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -67,25 +118,10 @@ export class TreeActions {
...
-          if (!(firstChildNode && isTreeModelInfoNode(firstChildNode))) {
-            timeTracker.finish();
-          }
-        },
-        complete: () => {
-          this.onLoadingComplete(parentId);
-          timeTracker[Symbol.dispose]();
-        },
-      });
+    const isComplete = new Subject<void>();
```
grigasp, line 121: Maybe it would make more sense to return a `Promise<void>` for this? Then, in the tests we could do `await actions.reloadTree();`, which looks cleaner than this:

```ts
const isComplete = actions.expandNode("root-1", true);
await waitFor(() => {
  expect(isComplete?.closed).to.be.true;
});
```

The implementation wouldn't get more complicated either:

```ts
return new Promise((resolve) => {
    // the loader should call `onComplete` callback instead of subscribing to and unsubscribing from `isComplete` subject
    this._nodeLoader.next({ loadOptions: options, onComplete: resolve, timeTracker, parentId: options.parent.id, initialRootNode });
});
```

I think that would make tests easier to read.

MartynasStrazdas: I do agree it would be easier to read but all the functions that return this would need to become async.

grigasp: > We'd need to add void to all the reload calls, I don't know how I feel about that.

Yeah, that's not great. What if we return `{ completed: Promise<void> }` object? We could still `await reload().complete`, but simply calling `reload()` wouldn't require an await.

MartynasStrazdas: I think that works, changed.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -233,7 +232,7 @@ export class TreeActions {
       return;
```
grigasp, line 232: maybe return `{ complete: Promise.resolve() }` here and in other places where we do an early return?

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -108,13 +109,15 @@ describe("TreeActions", () => {
...
       const actions = createActions(model);
+
       actions.reloadTree();
 
       await waitFor(() => {
         expect(onModelChangedStub).to.be.calledOnce;
         const newModel = onModelChangedStub.firstCall.args[0];
         expect(newModel.rootNode.isLoading).to.be.true;
         expect(getHierarchyNode(newModel, "root-1")).to.not.be.undefined;
+        expect(provider.getNodes).to.be.calledOnce;
```
grigasp, line 120: why is this `getNodes` check needed? do we even need `await waitFor(() => { ... })` if we do `await actions.reloadTree().complete`?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: Empty return should resolve pretty much immediately, I prefer basic return as it is cleaner.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -67,35 +115,22 @@ export class TreeActions {
...
-        },
-        complete: () => {
-          this.onLoadingComplete(parentId);
-          timeTracker[Symbol.dispose]();
-        },
-      });
+
+    return {
+      complete: new Promise<void>((resolve) => {
+        this._nodeLoader.next({ loadOptions: options, onComplete: resolve, timeTracker, parentId: options.parent.id, initialRootNode });
```
grigasp, line 121: the promise should be resolved and timetracker should be disposed for every request, including the ones that are debounced - does that happen?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: I'm wondering if it's enough to group by `parentId` and just take the last request from the group. Wouldn't that cause problems in cases like the following?

```
actions.reloadTree({ parentNodeId, state: "discard" });
actions.reloadTree({ parentNodeId, state: "keep" });
```

The first call would be ignored and the state would not be discarded, right?

MartynasStrazdas: The first request would be ignored, in this case do we want the state to be discarded? If we do this might be a bit tricky as the options passed to the pipe are the same (the difference is the return for shouldLoadChildren)

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: I think you could use `bufferTime` instead of `debounce`. That way you'd get all the requests, which would allow you to:
- merge their props (e.g. `state: "discard"` and `state: "keep"`),
- process the dropped ones.

grigasp: > bufferTime constantly emits values

Oh wow, never noticed that. I guess we could filter those out, but would be emitting those empty arrays would be a waste.

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
MartynasStrazdas: just removed getNodes.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: Fixed. Used a combination of buffer and debounceTime.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: Made it so if multiple reloads where made and atleast one of them was discard it would use that request.

packages/hierarchies-react/src/test/UseTree.test.tsx
```diff
@@ -84,6 +85,29 @@ describe("useTree", () => {
     });
   });
 
+  it("loads root nodes without raising hierarchy changed event", async () => {
```
JonasDov, line 88: How does this test check if event was raised or not?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -27,6 +27,7 @@ export interface LoadNodesOptions {
   shouldLoadChildren: (node: TreeModelHierarchyNode) => boolean;
   buildNode?: (node: TreeModelHierarchyNode) => TreeModelHierarchyNode;
   ignoreCache?: boolean;
+  discardState?: boolean;
```
grigasp, line 30: It doesn't look like it belongs here. Why not place the attribute on `LoadDebouncedNodesOptions` instead?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -36,6 +45,61 @@ export class TreeActions {
...
+              let returnIndex = last;
+
+              groupArr.forEach((member, index) => {
+                if (member.loadOptions.discardState) {
+                  returnIndex = index;
+                  return;
+                }
+                if (index === last && last === returnIndex) {
+                  return;
+                }
```
grigasp, line 71: what's this for?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -36,6 +45,61 @@ export class TreeActions {
...
+                member.onComplete();
```
grigasp, line 72: should we also dispose `timeTracker` here?

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -257,6 +259,58 @@ describe("TreeActions", () => {
       expect(getHierarchyNode(newModel, "root-1")?.isExpanded).to.be.false;
     });
 
+    it("Loads nodes correctly when different nodes are being expanded at the same time", async () => {
```
grigasp, line 262:
```suggestion
    it("loads nodes correctly when different nodes are being expanded at the same time", async () => {
```

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -525,57 +579,61 @@ describe("TreeActions", () => {
 
       const actions = createActions(model);
 
-      actions.setInstanceFilter("root-1", filter);
-
-      expect(onModelChangedStub).to.be.calledOnce;
+      await actions.setInstanceFilter("root-1", filter)?.complete;
+      await waitFor(() => expect(onModelChangedStub).to.be.calledTwice);
```
grigasp, line 583: is this `waitFor` needed?

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -525,57 +579,61 @@ describe("TreeActions", () => {
...
       const newModel = onModelChangedStub.firstCall.args[0];
       expect(getHierarchyNode(newModel, "root-1")?.instanceFilter).to.be.eq(filter);
       expect(getHierarchyNode(newModel, "child-1")).to.be.undefined;
+      await waitFor(() => {
+        expect(provider.getNodes).to.be.calledWith(createGetNodesProps({ parentNode: getHierarchyNode(newModel, "root-1")?.nodeData, instanceFilter: filter }));
+      });
```
grigasp, line 589: I don't think `getNodes` check is needed for this test

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -525,57 +579,61 @@ describe("TreeActions", () => {
...
+      expect(getHierarchyNode(newModel, "root-A")?.instanceFilter).to.be.eq(filter);
+      expect(getHierarchyNode(newModel, "root-A")?.isLoading).to.be.true;
 
       await waitFor(() => {
-        expect(provider.getNodes).to.be.calledWith(createGetNodesProps({ parentNode: getHierarchyNode(newModel, "root-1")?.nodeData, instanceFilter: filter }));
+        expect(provider.getNodes).to.be.calledWith(createGetNodesProps({ parentNode: getHierarchyNode(newModel, "root-A")?.nodeData, instanceFilter: filter }));
       });
 
-      expect(onModelChangedStub).to.be.calledTwice;
+      await waitFor(() => {
```
grigasp, line 630: is this `waitFor` needed?

packages/hierarchies-react/src/test/UseTree.test.tsx
MartynasStrazdas: Perhaps naming can be improved, but the provider we create only includes getNodes while other functions are empty, previously such a provider would not return root node. ( The event would come in from hierarchy filter or formatter setting functions in the formater, in this case these functions are empty)

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: We don't want to resolve the promise of last member within the array or if a discard/reset was passed then we do.

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
MartynasStrazdas: removed waitFors that go after `?.complete`

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: added the dispose, it just stops the clock which we don't need right now but might be useful in the future?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
MartynasStrazdas: moved.

packages/hierarchies-react/src/test/UseTree.test.tsx
MartynasStrazdas: changed the name

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: I was about to suggest simply using `index === returnIndex` check, but now I think the only appropriate way is to have two loops:
1. iterate through all members to find the one we're going to use,
2. iterate through all members and call `onComplete` on each of them except the one we got from step 1.

Without that, we may be not calling `onComplete` on some of the members. E.g. see what happens if you call `reloadTree({ state: "discard" })` many times.

MartynasStrazdas: You're right changed to have two loops.

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
MartynasStrazdas: I think this was fixed

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
MartynasStrazdas: I think this was fixed.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: It was added to support performance telemetry reporting. We want to measure the time it takes for us to handle requests.

In case of one request that's fairly simple, but now with us merging the requests, it's more complicated. I think we want both requests to report handling time. And the loading time of dropped requests should be the same as of the non-dropped one.

packages/hierarchies-react/src/test/UseTree.test.tsx
JonasDov: I think the naming is still not very good. Maybe rename to something like:
`loads root nodes by only using providers' getNodes function`.

Or you can add a check that tests if event was raised or not. then the current name would be fine.

MartynasStrazdas: changed the name, `by only using providers' getNodes function` might not be informative enough as all providers get root nodes from `getNodes` function. Let me know if this one any better.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: What purpose that that prop serve to consumers?

MartynasStrazdas: My thoughts were that if you are measuring performance with `onPerformanceMeasured` you might want to filter completed actions that actually did something or simply know that the performance measure is for an action that did not do something.

If we make the time of debounced requests and completed request the same I think we are getting skewed measure as then we are trying to prolong debounced request time and I don't see a point in doing so.

grigasp: IMO, we shouldn't add this burden on consumers. All they care is for us to tell them how long an action took.

Now, what we report is up to debate. I see 3 options:

1. Report everything, debounced requests are reported with ~0 request times.
   ```ts
   actions.reload();
   actions.reload();
   actions.reload(); // takes 5 seconds
   // we'd report [0, 0, 5s]
   ```

2. Report everything, debounced requests are reported with request time of the request they were merged with.
   ```ts
   actions.reload();
   actions.reload();
   actions.reload(); // takes 5 seconds
   // we'd report [5s, 5s, 5s]
   ```

3. Debounced requests aren't reported at all.
   ```ts
   actions.reload();
   actions.reload();
   actions.reload(); // takes 5 seconds
   // we'd report [5s]
   ```

We want the reported numbers to represent what users are experiencing, and since we know these numbers are used for telemetry, they should be good for calculating statistics: count, average, etc. Based on that, I think the 1st option is bad - it makes the average wrong. The second and third will have the same average, but the counts are different. Initially, I thought that we should use the 2nd option, but now that I think of it, 3rd is probably the correct one. From the API perspective we have 3 calls and caller will have to wait for all of them to complete to see result of the first one. However from the user's perspective there's just one reload action, and it takes as long as the last call. 

In that case, the initial solution was fine. Sorry for the confusion.

MartynasStrazdas: reverted

grigasp (approved):

PR #824, opened 2024-12-20
Increase performance consistency when filtering Models tree

JonasDov:


apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
```diff
@@ -24,15 +26,15 @@ import {
   ECSqlQueryExecutor,
   ECSqlQueryReaderOptions,
 } from "@itwin/presentation-shared";
-import { asyncScheduler, expand, filter, finalize, from, observeOn, of, tap } from "rxjs";
 import { LOGGER } from "../util/Logging";
 
 interface ProviderOptionsBase {
   rowLimit?: number | "unbounded";
-  getHierarchyFactory(imodelAccess: ECSchemaProvider & ECClassHierarchyInspector): HierarchyDefinition;
+  getHierarchyFactory(imodelAccess: ECSchemaProvider & ECClassHierarchyInspector, idsCache?: ModelsTreeIdsCache): HierarchyDefinition;
```
grigasp, line 33: I don't think we need to add `idsCache` in here - it can be passed to `getHierarchyFactory` through context. E.g. in `ModelsTree.test.ts`:

```ts
const provider = new StatelessHierarchyProvider({
  imodelAccess,
  getHierarchyFactory: (imodelAccess) => new ModelsTreeDefinition({ imodelAccess, idsCache }),
  filtering,
});
```

packages/models-tree/src/models-tree/ModelsTreeIdsCache.ts
```diff
@@ -28,11 +29,18 @@ export class ModelsTreeIdsCache {
   private _subjectInfos: Promise<Map<Id64String, SubjectInfo>> | undefined;
   private _parentSubjectIds: Promise<Id64Array> | undefined; // the list should contain a subject id if its node should be shown as having children
   private _modelInfos: Promise<Map<Id64String, ModelInfo>> | undefined;
+  private _modelKeyPaths: Map<Id64String, HierarchyNodeIdentifiersPath[]>;
+  private _subjectKeyPaths: Map<Id64String, HierarchyNodeIdentifiersPath>;
+  private _categoryKeyPaths: Map<Id64String, HierarchyNodeIdentifiersPath[]>;
```
grigasp, line 34: These maps should store a Promise to value, otherwise we could end up with unnecessary calculations. E.g.:

```ts
async function getSubjectPath(id) {
    const curr = this._subjectKeyPaths.get(id);
    if (curr) {
        return curr;
    }
    const value = await doExpensiveCalculation(id);
    this._subjectKeyPaths.set(id, value);
    return value;
}

const [value1, value2] = await Promise.all(getSubjectPath(""), getSubjectPath("")));
// doExpensiveCalculation is now called twice!
```

apps/performance-tests/src/hierarchies/StatelessHierarchyProvider.ts
JonasDov: Removed

packages/models-tree/src/models-tree/ModelsTreeIdsCache.ts
JonasDov: Changed, didn't think of that

grigasp (approved):

packages/models-tree/src/models-tree/ModelsTreeDefinition.ts
```diff
@@ -702,10 +677,13 @@ function createGeometricElementInstanceKeyPaths(
...
     mergeMap(({ modelId, elementHierarchyPath, groupingNode }) =>
-      createModelInstanceKeyPaths(modelId, idsCache).pipe(
+      from(idsCache.createModelInstanceKeyPaths(modelId)).pipe(
+        mergeAll(),
         map((modelPath) => {
-          modelPath.pop(); // model is already included in the element hierarchy path
-          const path = [...modelPath, ...elementHierarchyPath];
+          // We dont want to modify the original path, we create a copy that we can modify
+          const newModelPath = [...modelPath];
+          newModelPath.pop(); // model is already included in the element hierarchy path
```
grigasp, line 685: fancier way to achieve this would be:
```ts
const [_, ...newModelPath] = modelPath;
```

feel free to resolve this comment yourself if you don't like it

JonasDov: Didn't know that we can remove first element like that. But for this case it wouldn't work since `pop` removes the last element not the first

PR #827, opened 2025-01-03
Add error logging

JonasDov:


packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -41,7 +41,7 @@ export class TreeLoader implements ITreeLoader {
   constructor(
     private _hierarchyProvider: HierarchyProvider,
     private _onHierarchyLimitExceeded: (props: { parentId?: string; filter?: GenericInstanceFilter; limit?: number | "unbounded" }) => void,
-    private _onHierarchyLoadError: (props: { parentId?: string; type: "timeout" | "unknown" }) => void,
+    private _onHierarchyLoadError: (props: { parentId?: string; type: "timeout" | "unknown"; error: any }) => void,
```
grigasp, line 44: I think type of `error` should be changed to `Error`

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -168,7 +168,10 @@ function useTreeInternal({
   });
   const onPerformanceMeasuredRef = useLatest(onPerformanceMeasured);
   const onHierarchyLimitExceededRef = useLatest(onHierarchyLimitExceeded);
-  const onHierarchyLoadErrorRef = useLatest(onHierarchyLoadError);
+  const defaultOnHierarchyLoadError = (props: { parentId?: string; type: "timeout" | "unknown"; error: any }) => {
+    getLogger().logWarning("Hierarch load error", props.error);
+  };
```
grigasp, line 173: I don't think we should do this here - it should be completely up to the consuming component.

apps/test-app/frontend/src/components/tree-widget/MultiDataSourceTree.tsx
```diff
@@ -102,6 +102,9 @@ function Tree({ imodelAccess, height, width }: { imodelAccess: IModelAccess; hei
       [imodelAccess],
     ),
     getFilteredPaths,
+    onHierarchyLoadError: (props) => {
+      Logger.logWarning("loading error", props.error);
```
grigasp, line 106: IMO we should just use `console.error` to log the error

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
JonasDov: After talking to Saulius I decided to make this an `unknown` type.
The type of `err` in `catchError` (TreeLoader.ts) is actually `unknown`. This means that we can't assume that we will always get an error there which is of `Error` type. And because of this, in order to return an error of type `Error` in `onHierarchyLoadError` we would need to do custom handling of different types that error can actually take.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
JonasDov: Removed

apps/test-app/frontend/src/components/tree-widget/MultiDataSourceTree.tsx
JonasDov: Changed, but why do you suggest to use `console.error` instead?

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
JonasDov: Updated the changelog.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -61,6 +61,17 @@ export class TreeLoader implements ITreeLoader {
     ).pipe(
       toArray(),
       catchError((err) => {
+        if (!(err instanceof Error)) {
```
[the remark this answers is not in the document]
JonasDov: Changed the structure to be similar to what you suggested

grigasp (approved):

Issue #826, opened 2025-01-03
Symbol.dispose is not defined runtime error in Firefox


linasburneika:
I'm getting "Symbol.dispose is not defined" when running in Firefox locally in devserver. Works fine in Chrome.

```
Symbol.dispose is not defined.
./node_modules/@itwin/unified-selection/lib/esm/unified-selection/EnableUnifiedSelectionSyncWithIModel.js/__addDisposableResource<@http://localhost:3000/static/js/bundle.js:620770:40
applyCurrentHiliteSet@http://localhost:3000/static/js/bundle.js:620920:57
IModelSelectionHandler@http://localhost:3000/static/js/bundle.js:620860:14
enableUnifiedSelectionSyncWithIModel@http://localhost:3000/static/js/bundle.js:620825:30
./node_modules/@itwin/viewer-react/lib/esm/hooks/useUnifiedSelectionSync.js/useUnifiedSelectionSync/<@http://localhost:3000/static/js/bundle.js:419543:110
commitHookEffectListMount@http://localhost:3000/static/js/bundle.js:477909:26
commitPassiveMountOnFiber@http://localhost:3000/static/js/bundle.js:479685:38
commitPassiveMountEffects_complete@http://localhost:3000/static/js/bundle.js:479650:34
commitPassiveMountEffects_begin@http://localhost:3000/static/js/bundle.js:479637:41
commitPassiveMountEffects@http://localhost:3000/static/js/bundle.js:479625:34
flushPassiveEffectsImpl@http://localhost:3000/static/js/bundle.js:481798:28
flushPassiveEffects@http://localhost:3000/static/js/bundle.js:481743:14
performSyncWorkOnRoot@http://localhost:3000/static/js/bundle.js:480835:3
flushSyncCallbacks@http://localhost:3000/static/js/bundle.js:466767:22
commitRootImpl@http://localhost:3000/static/js/bundle.js:481718:3
commitRoot@http://localhost:3000/static/js/bundle.js:481441:19
finishConcurrentRender@http://localhost:3000/static/js/bundle.js:480740:19
performConcurrentWorkOnRoot@http://localhost:3000/static/js/bundle.js:480568:29
workLoop@http://localhost:3000/static/js/bundle.js:504139:42
flushWork@http://localhost:3000/static/js/bundle.js:504112:14
performWorkUntilDeadline@http://localhost:3000/static/js/bundle.js:504406:21
```

Version of @itwin/unified-selection installed: 1.2.0

To reproduce:
[...]

PR #828, opened 2025-01-07
Polyfill `Symbol.dispose` and `Symbol.asyncDispose`

saskliutas:
Closes https://github.com/iTwin/presentation/issues/826

Because JS explicit resource management feature is still in proposal state APIs associated with it are not available in all environments. Everything works fine in chromium based browsers and node.js but firefox does not define `Symbol.dispose` and `Symbol.asyncDispose`. This leads to the runtime errors. 

In order to use this feature while it is in proposal state `typescript` suggest to polyfill those symbols in case they are not present during runtime: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management. For this I have added module that is imported in everywhere where `Symbol.dispose` is used and it makes sure that `Symbol.dispose` is defined before first usage.

### DIFF
--- a/.changeset/silver-ears-change.md
+++ b/.changeset/silver-ears-change.md
@@ -1,0 +1,8 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+"@itwin/unified-selection": patch
+"@itwin/presentation-hierarchies": patch
+"@itwin/presentation-components": patch
+---
+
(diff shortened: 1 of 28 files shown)


MartynasStrazdas (approved):

grigasp (approved):

grigasp:
@wgoehrig FYI regarding https://github.com/iTwin/itwinjs-core/pull/7309

PR #831, opened 2025-01-08
[Tree Node Renderer]: Filter button visibility fix

MartynasStrazdas:
Closes https://github.com/iTwin/viewer-components-react/issues/1117

https://github.com/user-attachments/assets/528a1ee2-c5a7-42ee-b322-c2af697bd456

Still need to check if visibility button also causes the same bug.

### DIFF
--- a/.changeset/pink-paws-refuse.md
+++ b/.changeset/pink-paws-refuse.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Clicking on tree node buttons now visually show focus on node.
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.css
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.css
@@ -17,6 +17,12 @@
   visibility: visible;
 }
 
+.stateless-tree-node:focus-within {
+  outline: 1px solid var(--iui-color-border-accent);
+  outline-offset: -1px;
+  border-radius: var(--iui-border-radius-1);
+}
+
 .stateless-tree-node-small-spinner {
   height: var(--iui-size-s);
   width: var(--iui-size-s);
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
@@ -101,9 +101,6 @@ export const TreeNodeRenderer: React.ForwardRefExoticComponent<TreeNodeRendererP
           onExpanded={(_, isExpanded) => {
             expandNode(node.id, isExpanded);
           }}
-          expanderProps={{
-            onMouseDown: (e) => e.preventDefault(),
-          }}
           icon={getIcon ? getIcon(node) : undefined}
           label={getLabel ? getLabel(node) : node.label}
           sublabel={getSublabel ? getSublabel(node) : undefined}
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
@@ -101,6 +101,9 @@ export const TreeNodeRenderer: React.ForwardRefExoticComponent<TreeNodeRendererP
           onExpanded={(_, isExpanded) => {
             expandNode(node.id, isExpanded);
           }}
+          expanderProps={{
+            onMouseDown: (e) => e.preventDefault(),
+          }}
           icon={getIcon ? getIcon(node) : undefined}
           label={getLabel ? getLabel(node) : node.label}
           sublabel={getSublabel ? getSublabel(node) : undefined}


grigasp (commented):
needs a change file

saskliutas:
Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.

MartynasStrazdas:
> Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.

If you mean by clicking the expand with mouse and then using keys, then it seems it would not be posible with this implementation, but using keys to expand the node and continuing to use them is still good.

saskliutas:
> > Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.
> 
> If you mean by clicking the expand with mouse and then using keys, then it seems it would not be posible with this implementation, but using keys to expand the node and continuing to use them is still good.

Yes I am talking about expanding node using mouse and then continue with arrow keys.

grigasp:
> > > Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.
> > 
> > 
> > If you mean by clicking the expand with mouse and then using keys, then it seems it would not be posible with this implementation, but using keys to expand the node and continuing to use them is still good.
> 
> Yes I am talking about expanding node using mouse and then continue with arrow keys.

I agree - clicking on the expander should focus the node the expander is on. That should also take care of the issue (visible in the demo) where filtering button doesn't go away when expanding child nodes.

MartynasStrazdas:
> > > > Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.
> > > 
> > > 
> > > If you mean by clicking the expand with mouse and then using keys, then it seems it would not be posible with this implementation, but using keys to expand the node and continuing to use them is still good.
> > 
> > 
> > Yes I am talking about expanding node using mouse and then continue with arrow keys.
> 
> I agree - clicking on the expander should probably focus the node the expander is on, or expander itself. That should also take care of the issue (visible in the demo) where filtering button doesn't go away when expanding child nodes.

Originally the click on expander would focus it, which would mean we'd still have the issue of filter button being visible after expander click.

grigasp:
> > > > > Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.
> > > > 
> > > > 
> > > > If you mean by clicking the expand with mouse and then using keys, then it seems it would not be posible with this implementation, but using keys to expand the node and continuing to use them is still good.
> > > 
> > > 
> > > Yes I am talking about expanding node using mouse and then continue with arrow keys.
> > 
> > 
> > I agree - clicking on the expander should probably focus the node the expander is on, or expander itself. That should also take care of the issue (visible in the demo) where filtering button doesn't go away when expanding child nodes.
> 
> Originally the click on expander would focus it, which would mean we'd still have the issue of filter button being visible after expander click.

Could we show the filtering button when the focus is **within** the node rather than **directly on** the node?

MartynasStrazdas:
> > > > > > Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.
> > > > > 
> > > > > 
> > > > > If you mean by clicking the expand with mouse and then using keys, then it seems it would not be posible with this implementation, but using keys to expand the node and continuing to use them is still good.
> > > > 
> > > > 
> > > > Yes I am talking about expanding node using mouse and then continue with arrow keys.
> > > 
> > > 
> > > I agree - clicking on the expander should probably focus the node the expander is on, or expander itself. That should also take care of the issue (visible in the demo) where filtering button doesn't go away when expanding child nodes.
> > 
> > 
> > Originally the click on expander would focus it, which would mean we'd still have the issue of filter button being visible after expander click.
> 
> Could we show the filtering button when the focus is **within** the node rather than **directly on** the node?

That how it originally works, thats why clicking on expander kept the filter even when hovering on other nodes.

grigasp:
> > > > > > > Not sure if it's an issue but in previous version you could expand node and then continue navigating tree using keyboard arrows. With current implementation that is not a case.
> > > > > > 
> > > > > > 
> > > > > > If you mean by clicking the expand with mouse and then using keys, then it seems it would not be posible with this implementation, but using keys to expand the node and continuing to use them is still good.
> > > > > 
> > > > > 
> > > > > Yes I am talking about expanding node using mouse and then continue with arrow keys.
> > > > 
> > > > 
> > > > I agree - clicking on the expander should probably focus the node the expander is on, or expander itself. That should also take care of the issue (visible in the demo) where filtering button doesn't go away when expanding child nodes.
> > > 
> > > 
> > > Originally the click on expander would focus it, which would mean we'd still have the issue of filter button being visible after expander click.
> > 
> > 
> > Could we show the filtering button when the focus is **within** the node rather than **directly on** the node?
> 
> That how it originally works, thats why clicking on expander kept the filter even when hovering on other nodes.

LOL, yeah, you're right, we're back to the original issue.

I clicked around the "Source Control" tree in VSCode. The tree has similar stuff to ours: expanders, action buttons, keyboard navigation, etc. And I think it's pretty intuitive to use, so we should try to make ours similar. Here's what I think is needed:

1. Make expander non-focusable at all. When user clicks on it, the whole node should get focused.
2. When a node is focused, it should always have an outline. **This is the key to this issue.** It solves two problems that our tree currently has:
   - It's not clear why the filtering button is visible, because focused node has no outline.
   - It's not clear what'll happen when user starts keyboard navigation, because it's not clear what's focused. 

Can we do that?

grigasp:
> Maybe we should to get in touch with iTwinUI team regarding missing outline when node becomes focused after expander is clicked?

I think you're better suited to decide whether it's a bug in iTwinUI or something that should be implemented on our end. Contacting them and asking what they thing seems reasonable.

MartynasStrazdas:
> > Maybe we should to get in touch with iTwinUI team regarding missing outline when node becomes focused after expander is clicked?
> 
> I think you're better suited to decide whether it's a bug in iTwinUI or something that should be implemented on our end. Contacting them and asking what they think seems reasonable.

Tried the solution provided by ItwinUI with some tweaks, this solution would still be affected by the double border (like we see with multi select). This would be a css only solution.

https://github.com/user-attachments/assets/7f795d6e-b87d-46dd-b48b-6e52879a03c2

@grigasp @saskliutas

grigasp:
> > > Maybe we should to get in touch with iTwinUI team regarding missing outline when node becomes focused after expander is clicked?
> > 
> > 
> > I think you're better suited to decide whether it's a bug in iTwinUI or something that should be implemented on our end. Contacting them and asking what they think seems reasonable.
> 
> Tried the solution provided by ItwinUI with some tweaks, this solution would still be affected by the double border (like we see with multi select). This would be a css only solution.
> 
> @grigasp @saskliutas

LGTM

saskliutas:
> > > Maybe we should to get in touch with iTwinUI team regarding missing outline when node becomes focused after expander is clicked?
> > 
> > 
> > I think you're better suited to decide whether it's a bug in iTwinUI or something that should be implemented on our end. Contacting them and asking what they think seems reasonable.
> 
> Tried the solution provided by ItwinUI with some tweaks, this solution would still be affected by the double border (like we see with multi select). This would be a css only solution.
> 
>  navigation.mp4 
> @grigasp @saskliutas

Much better.

grigasp (changes requested):
Need a change file

grigasp (approved):

Issue #815, opened 2024-12-18
[hierarchies-react]: Add ability to disable filter button render on hover


saskliutas:
There might be use cases where hierarchy level limit is reached rarely and in those cases showing filter button on hover might by undesirable. For those cases we should add a simple way to disable filter button render on hover. 

It would still be possible to apply filter through info node if hierarchy level limit is exceeded. If filter is applied `Clear` and `Filter` should be rendered regardless.

MartynasStrazdas:
Perhaps we could allow user to choose to disable button-based filtering, leave it on hover or allow display of all filter buttons?

grigasp:
> Perhaps we could allow user to choose to disable button-based filtering, leave it on hover or allow display of all filter buttons?

By "user" you mean "consumer", right?

Also, just to confirm, are you suggesting that we provide 3 options: "no filtering buttons", "show filtering buttons on hover", "always show filtering buttons"?

MartynasStrazdas:
> > Perhaps we could allow user to choose to disable button-based filtering, leave it on hover or allow display of all filter buttons?
> 
> By "user" you mean "consumer", right?
> 
> Also, just to confirm, are you suggesting that we provide 3 options: "no filtering buttons", "show filtering buttons on hover", "always show filtering buttons"?

Yes and yes.

grigasp:
> > > Perhaps we could allow user to choose to disable button-based filtering, leave it on hover or allow display of all filter buttons?
> > 
> > 
> > By "user" you mean "consumer", right?
> > Also, just to confirm, are you suggesting that we provide 3 options: "no filtering buttons", "show filtering buttons on hover", "always show filtering buttons"?
> 
> Yes and yes.

I can't think of a reason anyone would want to see those buttons all the time. But I guess to future-proof ourselves we could use a string union prop rather than a boolean for this. Initially that could be `"hide" | "show-on-hover"`, and if there's a need, we could add `| "show-always"` in the future. What do you think?

saskliutas:
> > > > Perhaps we could allow user to choose to disable button-based filtering, leave it on hover or allow display of all filter buttons?
> > > 
> > > 
> > > By "user" you mean "consumer", right?
> > > Also, just to confirm, are you suggesting that we provide 3 options: "no filtering buttons", "show filtering buttons on hover", "always show filtering buttons"?
> > 
> > 
> > Yes and yes.
> 
[...]

PR #840, opened 2025-01-16
Add option to disable button based filtering

MartynasStrazdas:
closes https://github.com/iTwin/presentation/issues/815

Demo:
https://github.com/user-attachments/assets/5813eb99-5c30-4c1e-9732-5529b12372f2

### DIFF
--- a/.changeset/little-vans-sell.md
+++ b/.changeset/little-vans-sell.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-hierarchies-react": minor
 ---
 
-Added `disableFiltering` to `treeNodeRenderer`, which when set to true hides filter buttons, defaults to false.
+Added `filterButtonsVisibility` to `treeNodeRenderer`, which when set to `hide` hides filter buttons, defaults to `show-on-hover`.
--- a/.changeset/little-vans-sell.md
+++ b/.changeset/little-vans-sell.md
@@ -2,4 +2,6 @@
 "@itwin/presentation-hierarchies-react": minor
 ---
 
-Added `filterButtonsVisibility` to `treeNodeRenderer`, which when set to `hide` hides filter buttons, defaults to `show-on-hover`.
+Added `filterButtonsVisibility` to `treeNodeRenderer`. Which allows configuring filter buttons visibility for the whole tree.
+    `show-on-hover` - default value, shows filter buttons on node hover or focus.
+    `hide` - hides filter buttons on focus and hover, but will continue to show buttons on nodes in which filter is applied. Reaching hierarchy limit will continue to provide a way to filter nodes.
--- a/.changeset/little-vans-sell.md
+++ b/.changeset/little-vans-sell.md
@@ -3,5 +3,6 @@
 ---
 
 Added `filterButtonsVisibility` to `treeNodeRenderer`. Which allows configuring filter buttons visibility for the whole tree.
-    `show-on-hover` - default value, shows filter buttons on node hover or focus.
-    `hide` - hides filter buttons on focus and hover, but will continue to show buttons on nodes in which filter is applied. Reaching hierarchy limit will continue to provide a way to filter nodes.
+
+- `show-on-hover` - default value, shows filter buttons on node hover or focus.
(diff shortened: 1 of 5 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -39,6 +39,8 @@ interface TreeNodeRendererOwnProps {
   actionButtonsClassName?: string;
   /** Tree node size. Should match the size passed to `TreeRenderer` component. */
   size?: "default" | "small";
+  /** Disables button-based filtering and hides filter buttons. Defaults to `false`*/
+  disableFiltering?: boolean;
```
saskliutas, line 43: I think as suggested in the issue it would be better to use string literal union instead of boolean. Something like `filterButtonStyleType?: "show-on-hover" | "hide"`.

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: was unsure about that, changed.

saskliutas (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -39,6 +39,8 @@ interface TreeNodeRendererOwnProps {
   actionButtonsClassName?: string;
   /** Tree node size. Should match the size passed to `TreeRenderer` component. */
   size?: "default" | "small";
+  /** Configures filter buttons visibility. Defaults to `show-on-hover`*/
```
grigasp, line 42: I think we could do better with the docs - explain in more detail what each option means. Things worth mentioning:
- setting to `hide` will still show the buttons if the node is filtered,
- setting to `show-on-hover` will also show on focus.

Similarly, I'd like a more detailed explanation in the changeset file.

packages/hierarchies-react/src/test/itwinui/Tree.test.tsx
```diff
@@ -353,6 +353,25 @@ describe("Tree", () => {
     expect(onFilterClick).to.be.calledOnceWith(hierarchyLevelDetails);
   });
 
+  it("renders filter button when filter buttons are hiden, but node is filtered", async () => {
```
grigasp, line 356:
```suggestion
  it("renders filter button when filter buttons are hidden, but node is filtered", async () => {
```

.changeset/little-vans-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": minor
+---
+
+Added `filterButtonsVisibility` to `treeNodeRenderer`, which when set to `hide` hides filter buttons, defaults to `show-on-hover`.
```
grigasp, line 5: I'd change this to:

```md
Added a `filterButtonsVisibility` prop to control filtering action buttons' display in trees and tree nodes:
  
  <explain available options (see the other comment about docs)>
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: Changed documentation.

.changeset/little-vans-sell.md
MartynasStrazdas: Changed change log.

MartynasStrazdas: 3rd time's the charm

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: 3rd time's the charm. Changed

grigasp (approved):

Issue #830, opened 2025-01-08
`presentation-components`: Deprecate unified selection related APIs


grigasp:
Prerequisite: https://github.com/iTwin/presentation/issues/829

The `@itwin/unified-selection` package replaces the selection-related APIs from `@itwin/core-frontend`, that are currently used by `UnifiedSelectionContext` in `@itwin/presentation-components`. All these APIs in `@itwin/presentation-frontend` and `@itwin/presentation-components` should be deprecated in favor of the ones in `@itwin/unified-selection`.

Issue #829, opened 2025-01-08
`unified-selection`: Supply APIs to provide and use React context for accessing selection storage


grigasp:
1. Add an optional peer dependency of react to @itwin/unified-selection.
2. Provide React context provider and hook to provide/use selection storage.

For consumers this means:

- Applications that create selection storage can provide it to underlying components using the context provider. They'd have a direct dependency on the `@itwin/unified-selection` package.
- Components that need to access the storage, without having it in props, can use the `useContext` hook. They'd have a peer dependency on the `@itwin/unified-selection` package.
- Components that don't use the context, but use selection storage in some way, can continue to have direct dependency on the `@itwin/unified-selection` package, like they do today.

Need to ensure documentation is very clear about how to set up dependencies on this package.

saskliutas:
Should we consider introducing `@itwin/unified-selection-react` package that would have APIs for storing and getting selection storage through React context? Libraries that want to use unified selection storage through context would have peer dependency on it and App would add that context to the components tree.

It probably will be a package with single file but that way we could keep all unified selection stuff decoupled from React (similar to hierarchies and hierarchies-react).

PR #841, opened 2025-01-17
Add `unified-selection-react` package

grigasp:
Closes https://github.com/iTwin/presentation/issues/829
Closes https://github.com/iTwin/presentation/issues/830

This probably should've been at least 2 PRs, but there was a lot of back and forth, and I didn't want to waste time splitting them afterwards.

The new package now provides a React context that other packages can use, so consumers don't need to have a separate provider for each package. However, I also made changes in `hierarchies-react` and `components` to consume selection storage through a prop rather than context, so the context is only used by `InstanceKeyValueRenderer`, which has no way of getting the selection storage through prop.


packages/components/src/presentation-components/common/UnifiedSelection.tsx
```diff
@@ -0,0 +1,39 @@
...
+const noopContext = createContext<ReturnType<typeof useUnifiedSelectionContext>>(undefined);
+
+async function getUnifiedSelectionContextFn() {
+  try {
+    const optionalPkg = await import("@itwin/unified-selection-react");
+    return optionalPkg.useUnifiedSelectionContext;
+    /* c8 ignore next 5 */
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(e);
```
saskliutas, line 36: Should we log this error to the console while `@itwin/unified-selection-react` is optional and there is a valid use case where this package will not be available?

packages/components/package.json
```diff
@@ -96,6 +105,8 @@
     "@itwin/itwinui-react": "catalog:itwinui",
     "@itwin/presentation-common": "catalog:itwinjs-core",
     "@itwin/presentation-frontend": "catalog:itwinjs-core",
+    "@itwin/unified-selection": "workspace:^",
```
saskliutas, line 108: It should be enough to have only in dependencies `@itwin/unified-selection`

saskliutas (changes requested):

.changeset/thin-windows-decide.md
```diff
@@ -0,0 +1,24 @@
+---
```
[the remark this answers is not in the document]
grigasp: removed :confused:

packages/components/src/presentation-components/common/UnifiedSelection.tsx
grigasp: agree, makes no sense to log an error here - removed

packages/components/package.json
grigasp: removed

saskliutas (approved):

PR #847, opened 2025-01-21
Kiwi tree

MartynasStrazdas:
Early preview of how the kiwi tree looks.

There are plenty of things that are not yet released from kiwi side, that we want:
1. Separate error nodes: will be provided eventually
2. Action items wrapper: so far can't provide use case or check if we need it as actions don't seem to work as intended with icons.
3. Multiline label: for now only label is available, sublabel should be added by kiwi.
4. Size for tree items: size is removed, <root> component has property `density` to control size of all components.
6. ...

Tree renderer property name changes so far:
IsSelected => selected
isDisabled => aria-disabled (need to figure out if we want a separate prop)

General:
1. Icons are now provided through Href, but it seems all tree will continue to support basic svg (actions/node icons).
2. There is no to little changes for using new TreeRenderer.

Tree preview:

https://github.com/user-attachments/assets/5a157d8c-c9a2-4410-9c74-eb331e7ae8d8


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -3,28 +3,26 @@
...
-type TreeNodeProps = ComponentPropsWithoutRef<typeof TreeNode>;
+/** @alpha */
+type TreeNodeProps = Omit<ComponentPropsWithoutRef<typeof Tree.Item>, "actions">;
 
-/** @public */
-interface TreeNodeRendererOwnProps {
+/** @alpha */
+export interface TreeNodeRendererOwnProps {
   /** Node that is rendered. */
   node: RenderedTreeNode;
```
saskliutas, line 21: Looks like this renderer does not receive `RenderedTreeNode` anymore.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -37,24 +35,14 @@ interface TreeNodeRendererOwnProps {
...
-   *
-   * Options:
-   * - `show-on-hover` - show filter buttons when hovering over node, or the node is in focus.
-   * - `hide` - hide filter buttons, but will show them if the filter is applied.
-   *
-   * Default value: `show-on-hover`
-   */
-  filterButtonsVisibility?: "show-on-hover" | "hide";
+  /** Renderer for additional tree items actions */
+  actionRenderers?: (node: PresentationHierarchyNode) => ReactElement;
```
saskliutas, line 39:
```suggestion
  actionsRenderer?: (node: PresentationHierarchyNode) => ReactElement;
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -71,250 +59,73 @@ export const TreeNodeRenderer: React.ForwardRefExoticComponent<TreeNodeRendererP
...
-              : undefined
-          }
-        />
-      );
-    }
-
-    if (node.type === "NoFilterMatches") {
-      return <TreeNode {...treeNodeProps} ref={ref} label={localizedStrings.noFilteredChildren} isDisabled={true} onExpanded={/* c8 ignore next */ () => {}} />;
-    }
+    const isDisabled = false;
```
saskliutas, line 87: Why is this always `false`?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButtons.tsx
```diff
@@ -0,0 +1,62 @@
...
+      }}
+      icon={node.isFiltered ? filterIcon : filterIcon} // currently base filter icon is hollow
+    />
+  ) : undefined;
+}
+
+/** @internal */
+export function RemoveFilterActionButton({ node, getHierarchyLevelDetails }: ActionButtonProps) {
+  const { localizedStrings } = useLocalizationContext();
+  const applyFilterButtonRef = useRef<HTMLButtonElement>(null);
```
saskliutas, line 47: Looks like refs in these buttons are not useful anymore.

apps/test-app/frontend/vite.config.mts
```diff
@@ -44,4 +44,10 @@ export default defineConfig({
       },
     ],
   },
+  build: {
+    assetsInlineLimit: (filePath) => {
+      if (filePath.includes("@itwin/itwinui-icons/")) return false;
```
grigasp, line 49:
```suggestion
      if (filePath.includes("@itwin/itwinui-icons/")) {
        return false;
      }
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -37,24 +35,14 @@ interface TreeNodeRendererOwnProps {
   reloadTree?: (options: { parentNodeId: string | undefined; state: "reset" }) => void;
   /** CSS class name for the action buttons. */
   actionButtonsClassName?: string;
```
grigasp: if we allow consumers to render anything through `actionsRenderer`, does `actionButtonsClassName` have any value anymore?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButtons.tsx
```diff
@@ -0,0 +1,62 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { useRef } from "react";
+import { IconButton } from "@itwin/itwinui-react/bricks";
+import { PresentationHierarchyNode } from "../TreeNode.js";
+import { HierarchyLevelDetails, useTree } from "../UseTree.js";
+import { useLocalizationContext } from "./LocalizationContext.js";
+
+const filterIcon = new URL("@itwin/itwinui-icons/filter.svg", import.meta.url).href;
+const SvgRemove = new URL("@itwin/itwinui-icons/dismiss.svg", import.meta.url).href;
```
grigasp, line 12: please make the naming consistent

apps/test-app/frontend/src/components/app/App.tsx
```diff
@@ -125,22 +126,24 @@ export function App() {
   }, []);
 
   return (
-    <ThemeProvider theme="os">
-      <UnifiedSelectionContextProvider storage={MyAppFrontend.selectionStorage}>
-        <div className="app">
-          <div className="app-header">
-            <h2>{IModelApp.localization.getLocalizedString("Sample:welcome-message")}</h2>
+    <Root colorScheme="light" density="dense">
+      <ThemeProvider theme="light" future={{ themeBridge: true }}>
```
grigasp, line 130: why not "os" theme? would make it easier to test both cases

apps/test-app/frontend/src/components/app/App.tsx
```diff
@@ -236,9 +239,9 @@ function IModelComponents(props: IModelComponentsProps) {
               canPopout: true,
             },
             {
-              id: "stateless-models-tree",
-              label: "Stateless Models tree",
-              content: <StatelessModelsTreePanel imodel={imodel} />,
+              id: "stateless-models-kiwi-tree",
+              label: "Stateless Models Kiwi tree",
+              content: <StatelessModelsKiwiTreePanel imodel={imodel} />,
```
grigasp, line 244: I don't think we need these changes

apps/test-app/frontend/src/components/app/App.tsx
```diff
@@ -319,7 +322,7 @@ function RulesDrivenTreePanel(props: { imodel: IModelConnection; rulesetId?: str
   );
 }
 
-function StatelessModelsTreePanel(props: { imodel: IModelConnection }) {
+function StatelessModelsKiwiTreePanel(props: { imodel: IModelConnection }) {
```
grigasp, line 325: Let's keep the old name

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: We need to figure out if we need a callback to determine state. As right now there is no way to control it from treeRenderer. For now removed isDisabled and is using `aria-disabled`.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButtons.tsx
MartynasStrazdas: Removed.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: That was removed.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
saskliutas: Looks like in old implementation it was also not possible to control `isDisabled` through `TreeRenderer`. It was possible only when using `TreeNodeRenderer` directly. I think same approach could be used in new implementation. 

Or maybe we do not need to have an ability to disable nodes. Only disable error nodes to make then non-interactable.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButtons.tsx
MartynasStrazdas: Changed

apps/test-app/frontend/src/components/app/App.tsx
MartynasStrazdas: reverted

saskliutas (approved):
Looks good for an initial attempt. Should be enough to integrate into tree-widget.

PR #860, opened 2025-01-30
`hierarchies-react`: Cleanup filtering action buttons

MartynasStrazdas:
Filter icon style fix
image

Removed filter reset button.

### DIFF
--- a/.changeset/famous-bags-pull.md
+++ b/.changeset/famous-bags-pull.md
@@ -2,5 +2,6 @@
 "@itwin/presentation-hierarchies-react": patch
 ---
 
-Fix tree filter button style.
-Removed filter reset button.
+Changes to tree node action buttons related to filtering:
+- Fix filter button style.
+- Removed "reset filter" button. The filter can be reset through the filter editing dialog.
--- a/.changeset/famous-bags-pull.md
+++ b/.changeset/famous-bags-pull.md
@@ -1,0 +1,6 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Fix tree filter button style.
+Removed filter reset button.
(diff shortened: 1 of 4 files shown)


saskliutas (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButtons.tsx
```diff
@@ -9,7 +9,7 @@ import { HierarchyLevelDetails, useTree } from "../UseTree.js";
 import { useLocalizationContext } from "./LocalizationContext.js";
 
 const filterIcon = new URL("@itwin/itwinui-icons/filter.svg", import.meta.url).href;
-const dismissIcon = new URL("@itwin/itwinui-icons/dismiss.svg", import.meta.url).href;
+const activeFilterIcon = new URL("@itwin/itwinui-icons/placeholder.svg", import.meta.url).href; // Placeholder
```
grigasp, line 12: does this mean we're missing an icon? if yes, please add it to "Icons status" list in https://github.com/iTwin/viewer-components-react/issues/1149#issue-2808932282

.changeset/famous-bags-pull.md
```diff
@@ -0,0 +1,6 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Fix tree filter button style.
+Removed filter reset button.
```
grigasp, line 6: This won't render well in changelog. Let's restructure like this:

```suggestion
Changes to tree node action buttons related to filtering:
- Fix filter button style.
- Removed "reset filter" button. The filter can be reset through the filter editing dialog.
```

grigasp (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButtons.tsx
MartynasStrazdas: I was under the impression that this includes it.
image

grigasp: yes, my bad (somehow I thought we need two icons - regular one and "active filter" one)

PR #870, opened 2025-02-05
[Hierarchies-react]: add dropdown button actions

MartynasStrazdas:
When multiple actions are provided:
image
image
When one action is provided:
image
When no actions are provided:
image

Keeping icon button position in the same place when other icon buttons are not displayed:
image
image
When dropdown actions exist but none are visible for a node.
image

### DIFF
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
@@ -30,13 +30,7 @@ export type FilterActionProps = {
 /** @internal */
 export function TreeActionButton({ show, label, action, icon }: TreeItemAction) {
   return (
-    <IconButton
-      className={`tree-action-item${!show ? "-invisible" : ""}`}
-      variant={"ghost"}
-      onClick={action}
-      label={label}
-      icon={icon ?? placeholderIcon}
-    />
+    <IconButton className={`tree-action-item${!show ? "-invisible" : ""}`} variant={"ghost"} onClick={action} label={label} icon={icon ?? placeholderIcon} />
   );
 }
 
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
@@ -7,17 +7,24 @@ import "./TreeActionButton.css";
 import { IconButton } from "@itwin/itwinui-react/bricks";
 import { PresentationHierarchyNode } from "../TreeNode.js";
 import { HierarchyLevelDetails, useTree } from "../UseTree.js";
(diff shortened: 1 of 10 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -51,6 +63,9 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
       <Tree.Root style={{ height: "100%", width: "100%" }}>
         <TreeLevelRenderer
           {...treeProps}
+          actions={[...(actions || []), getFilterAction({ onClick: onFilterClick, getHierarchyLevelDetails, label: localizedStrings?.filterHierarchyLevel })]}
```
saskliutas, line 66: I think we should leave for consumer to add `filterButton`. That way it will be possible to fully control buttons position. I think we should have something like:

```tsx
// in consumer component
return <TreeRenderer
  {...treeProps}
  actions=[
    (node) => createMyCustomAction(node),
    createFilterAction({ 
      onFilter: handleFilter // same as `onFilerClick` at the moment
      getHierarchyLevelDetails
    })
  ]
/>
```

saskliutas (approved):
Looks ok for now. We can revisit API for providing actions later.

I guess we could remove `isDropdownAction` and only show up to 3 buttons as discussed.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
MartynasStrazdas: Changed, the overflow into dropdown is not perfect and there could be cases where the overflow button will have one action, but I don't think there is any reason to add any more effort when itwinUI will change action logic. For filterButton exposed it with alpha tag, made label required for now consumers should have localizedStrings anyway should circle back to it when final version of itwinUI actions is available.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+  label: string;
+} & Partial<Pick<ReturnType<typeof useTree>, "getHierarchyLevelDetails">>;
+
+/** @internal */
+export function TreeActionButton({ show, label, action, icon }: TreeItemAction) {
+  return (
+    <IconButton
+      className={`tree-action-item${!show ? "-invisible" : ""}`}
+      variant={"ghost"}
+      onClick={() => action()}
```
saskliutas, line 36:
```suggestion
      onClick={action}
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
saskliutas, line 1: I guess these changes can be reverted.

saskliutas (approved):
Small cleanup.

PR #875, opened 2025-02-17
`unified-selection`: Add more learning material

grigasp:
As part of https://github.com/iTwin/itwinjs-core/issues/7527, I wanted to move the documentation material from `itwinjs-core` to the `unified-selection` package. Also added a migration guide which we can link from `NextVersion.md` in `itwinjs-core`.


packages/unified-selection/learning/SelectionScopes.md
```diff
@@ -0,0 +1,37 @@
...
+
+```ts
+import { computeSelection } from "@itwin/unified-selection";
+import { IModelConnection } from "@itwin/core-frontend";
+import { createECSqlQueryExecutor } from "@itwin/presentation-core-interop";
+const queryExecutor = createECSqlQueryExecutor(imodel);
+const selection = computeSelection({ queryExecutor, elementIds, scope: "element" });
+```
+
+`element` and `functional` scopes additionally allow selecting assembly elements by specifying the `ancestorLevel` property in the selection scope argument of `computeSelection` function. The `ancestorLevel` property specifies how far "up" we should walk to find the target element. When not specified or `0`, the target element matches the request element. When set to `1`, the target element matches the direct parent element. When `2`, the target element is the parent of the parent element, and so on. In all situations when this is `> 0`, we're not walking further than the last existing element, for example, when `ancestorLevel = 1` (direct parent element is requested), but the request element doesn't have a parent, the request element is returned as the result. A negative value would result in the top-most element to be returned.
```
saskliutas, line 24: This paragraph is the same as the previous one. Is this intentional?

grigasp: nm, this is a result of me moving things around... removed one of these paragraphs

saskliutas (approved):

PR #878, opened 2025-02-19
[Hierarchies-react]: flat tree

MartynasStrazdas:
Updated hierarchies react to consume  itwinUI alpha.6.


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -46,20 +45,47 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
...
-          nodes={rootNodes}
-          expandNode={expandNode}
-          onNodeClick={onNodeClick}
-          onNodeKeyDown={onNodeKeyDown}
-          isNodeSelected={isNodeSelected}
-        />
+        {flatNodes.map((flatNode) => (
+          <TreeNodeRenderer
+            {...treeProps}
+            aria-level={flatNode.level}
```
saskliutas, line 56: Is there a reason why we can't pass whole `flatNode` to `TreeNodeRenderer` and set these properties inside it?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -46,20 +45,47 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
...
+interface FlatPresentationTreeNode {
+  node: PresentationTreeNode;
+  level: number;
+  levelSize: number;
+  posInLevel: number;
+}
+
+export function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
+  const flatNodes: FlatPresentationTreeNode[] = [];
+  nodes.map((node, index) => {
```
saskliutas, line 83: `.forEach` is more applicable in this use case.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
```diff
@@ -24,26 +30,34 @@ export type FilterActionProps = {
...
+      }
+      if (node.isFiltered) {
+        return true;
+      }
+      return undefined;
+    };
+
     return {
-      label,
+      label: label ?? defaultLocalizedStrings.filterHierarchyLevel,
```
saskliutas, line 55: `defaultLocalizedStrings` should not be used here. I think this could be handled in `TreeActionButton` by checking if `label` mathes any of the localized strings that we support.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -37,16 +37,30 @@ export const TreeErrorRenderer: React.ForwardRefExoticComponent<TreeErrorRendere
                 }
               : undefined
           }
+          aria-level={0}
```
saskliutas, line 40: I think error nodes should use information from 'flattened node'. `aria-level={0}` does not look like correct approach. Also why only `NoFilterMatches` node has `aria-posinset` and `aria-setsize` set as 0?

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
MartynasStrazdas: Agree, changed

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
MartynasStrazdas: Changed

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
MartynasStrazdas: Changed to use level from flattened node, missed `NoFilterMatches`, changed set size and pos in set to 1

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
MartynasStrazdas: I guess there is none, just different ways to handle level and other flat tree props. Changed to level being provided from tree node as well tweaked how flat tree node type looks.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
```diff
@@ -15,35 +15,49 @@ const placeholderIcon = new URL("@itwin/itwinui-icons/placeholder.svg", import.m
...
+      }
+      if (node.isFiltered) {
+        return true;
+      }
+      return undefined;
+    };
+
     return {
-      label,
+      label: "filterHierarchyLevel",
```
saskliutas, line 55:
```suggestion
      label: "filterHierarchyLevel" satisfies keyof LocalizedStrings,
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -46,20 +45,43 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
...
 
 function noopSelectNodes() {}
+
+export type FlatPresentationTreeNode = {
+  level: number;
+  levelSize: number;
+  posInLevel: number;
+} & PresentationTreeNode;
+
+export function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
```
saskliutas, line 77: I think for public facing API it would be better to have something like:

```ts
/** @alpha */
export function flattenNodes(nodes: PresentationTreeNode[]) {
  return getFlatNodes(nodes, 0);
}

/** @internal */
function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
}
```

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
saskliutas: Why some type of nodes still have `setsize` and `posinset` values set to constants instead of using information from `FlatNode`?

MartynasStrazdas: I don't think there is a reason to clutter interfaces with `setsize` and `posinset` when we know that their pos and set size is a constant 1.

saskliutas: Can't we just pass whole `FlatNode` to those places?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
MartynasStrazdas: Changed.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -46,20 +45,49 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
     selectionMode: selectionMode ?? "single",
   });
 
+  const flatNodes = useMemo(() => getFlatNodes(rootNodes, 0), [rootNodes]);
```
saskliutas, line 48:
```suggestion
  const flatNodes = useMemo(() => flattenNodes(rootNodes), [rootNodes]);
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
MartynasStrazdas: We already check if its an info (error node), so we'd need to add flat properties inside it.

saskliutas: You could do something like:

```ts
type FlatTreeNode<TNode extends PresentationTreeNode = PresentationTreeNode> = {
  level: number;
  levelSize: number;
  posInLevel: number;
} & TNode;

interface TreeErrorRendererOwnProps {
  node: FlatTreeNode<PresentationInfoNode>;
}
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -77,79 +77,94 @@ export const TreeNodeRenderer: React.ForwardRefExoticComponent<TreeNodeRendererP
...
+          onKeyDown={(event) => {
+            // Ignore if it is called on the element inside, e.g. checkbox or expander
+            if (!treeItemProps["aria-disabled"] && event.target === nodeRef.current) {
+              onNodeKeyDown?.(node, !selected, event);
+            }
+          }}
+          icon={getIcon ? getIcon(node) : undefined}
+          actions={getActions()}
+        />
+        {node.isExpanded && node.children === true && <PlaceholderNode {...treeItemProps} level={node.level + 1} ref={ref} />}
```
saskliutas, line 148: I think this might cause problems with virtualization. I think we should add entry into flattened nodes array for placeholder nodes and render it as separate `Tree.Item` component instead of rendering two components from single expanded tree node without children loaded.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -28,7 +27,7 @@ interface TreeRendererOwnProps {
 /** @alpha */
 type TreeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode"> &
   Partial<Pick<ReturnType<typeof useTree>, "selectNodes" | "isNodeSelected" | "getHierarchyLevelDetails" | "reloadTree">> &
-  Omit<TreeNodeRendererProps, "node" | "reloadTree"> &
+  Omit<TreeNodeRendererProps, "node" | "reloadTree" | "aria-level" | "aria-posinset" | "aria-setsize"> &
```
saskliutas, line 30: I guess `aria` props are not needed.

MartynasStrazdas: They are being omitted.

saskliutas: Aren't they already emitted here? https://github.com/iTwin/presentation/blob/596bf2fac592f5da9c4ac896d74898aa8a5d3aa0/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx#L18

Also looking at that `TreeNodeProps` type. Can we move `Omit` parts to https://github.com/iTwin/presentation/blob/596bf2fac592f5da9c4ac896d74898aa8a5d3aa0/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx#L47 to avoid having multiple places omitting props from same base type.

MartynasStrazdas: Yes, you are right, forgot that node renderer needed to omit those props also. Changed both types

saskliutas: Can we move `Omit` part to the `TreeNodeRenderer` type definition and keep `type TreeNodeProps = ComponentPropsWithoutRef<typeof Tree.Item>`

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
MartynasStrazdas: Changed Flat tree node interface to the one suggested, seems more flexible.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: Moved placeholder rendering outside tree item, added a `placholder` boolean prop to flat tree node interface.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
MartynasStrazdas: Moved it out

saskliutas (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -46,20 +45,57 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
...
+    if (!isPresentationHierarchyNode(node) || !node.isExpanded) {
+      return;
+    }
+    if (node.children !== true) {
+      const childNodes = getFlatNodes(node.children, level + 1);
+      flatNodes.push(...childNodes);
+      return;
+    }
+
+    flatNodes.push({ ...node, level: level + 1, levelSize: 1, posInLevel: 1, placeholder: true });
```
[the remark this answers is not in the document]
MartynasStrazdas: We need to provide some kind of node or could add another type to `PresentationTreeNode` specifically for placeholder node.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -46,20 +45,57 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
...
-        />
+        {flatNodes.map((flatNode) => (
+          <TreeNodeRenderer
+            {...treeProps}
+            expandNode={expandNode}
+            onNodeClick={onNodeClick}
+            onNodeKeyDown={onNodeKeyDown}
+            node={flatNode}
+            key={flatNode.id}
+            selected={isNodeSelected?.(flatNode.id)}
```
saskliutas, line 61: Is there a reason why `TreeNodeRenderer` takes `selected` flag instead of `isNodeSelected`?

MartynasStrazdas: selected comes from itwinUI side, we could pass isNodeSelected and omit selected, then call it inside the `TreeNodeRenderer`, but I don't think  that's necessary as it will just add additional code lines we can avoid.

saskliutas: I think it makes sense handling `isNodeSelected` inside `TreeNodeRenderer` because at the moment we call `isNodeSelected` even for error nodes. Also looking into this more I think `TreeNodeRenderer` should not allow passing `expanded` flag as it is not used internally.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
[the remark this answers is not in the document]
MartynasStrazdas: Expanded placeholder logic.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!isPresentationHierarchyNode(node) || !node.isExpanded) {
+      return;
+    }
+    if (node.children !== true) {
+      const childNodes = getFlatNodes(node.children, level + 1);
+      flatNodes.push(...childNodes);
+      return;
+    }
+
+    flatNodes.push({ id: "placeholderNode", level: level + 1, placeholder: true } satisfies PlaceholderNode);
```
saskliutas, line 47: There might be multiple placeholders rendered.
```suggestion
    flatNodes.push({ id: `${node.id}-children-placeholder`, level: level + 1, placeholder: true } satisfies PlaceholderNode);
```

saskliutas (approved):

PR #880, opened 2025-02-20
Fix api-extraction

saskliutas:
Switched from `presentation-build-tools` to `@itwin/build-tools` for api extraction.
Bump typescript version


grigasp (commented):
what was wrong with it? I thought we made the necessary improvements in `betools` to suit our needs

saskliutas:
Noticed that `unified-selection-react` was the only package using `betools` so decided to unify that. If `betools` have everything needed can we switch to that and remove `presentation-build-tools`?

grigasp:
> Noticed that `unified-selection-react` was the only package using `betools` so decided to unify that. If `betools` have everything needed can we switch to that and remove `presentation-build-tools`?

yes, I believe so

PR #892, opened 2025-03-03
[Hierarchy-react]: focus outline on tree nodes

MartynasStrazdas:
Closes https://github.com/iTwin/viewer-components-react/issues/1215

Current version: 
image

Because kiwi tree uses outline with offset -1px, we have to use `before` to have outline not hidden behind action buttons.

Ps. Kiwi tree currently does not handle padding correctly and outline is not fully visible in some cases, added outline mimics the outline kiwi tree provides with keyboard navigation. The issue of not fully visible outlines should be fixed on kiwi's side

### DIFF
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
@@ -3,6 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+import "./TreeNodeRenderer.css";
 import {
   ComponentPropsWithoutRef,
   forwardRef,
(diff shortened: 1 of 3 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -135,6 +136,7 @@ export const TreeNodeRenderer: ForwardRefExoticComponent<TreeNodeRendererProps &
     return (
       <Tree.Item
         {...treeItemProps}
+        className={"stateless-tree-node"}
```
grigasp, line 139: I don't think we should use the term "stateless tree" anymore - mind changing this to `presentation-hierarchies-node`?

MartynasStrazdas: Changed

grigasp (commented):
LGTM. @saskliutas, any concerns?

mayank99:
This can be undone after updating to the latest alpha.

PR #895, opened 2025-03-04
Get rid of testdouble

grigasp:
Don't have an issue for this repo, but this is for fixing this: https://github.com/iTwin/viewer-components-react/issues/1190

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -43,13 +43,13 @@
     "clean": "rimraf lib build",
     "cover": "c8 npm run -s test",
     "lint": "eslint ./src/**/*.{ts,tsx}",
-    "test:dev": "cross-env NODE_OPTIONS=\"--import presentation-test-utilities/node-hooks/ignore-styles\" mocha --enable-source-maps --config ./.mocharc.json",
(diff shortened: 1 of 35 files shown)


packages/components/package.json
```diff
@@ -43,13 +43,13 @@
     "clean": "rimraf lib build",
     "cover": "c8 npm run -s test",
     "lint": "eslint ./src/**/*.{ts,tsx}",
-    "test:dev": "cross-env NODE_OPTIONS=\"--import presentation-test-utilities/node-hooks/ignore-styles\" mocha --enable-source-maps --config ./.mocharc.json",
+    "test:dev": "node --experimental-test-module-mocks --enable-source-maps --import presentation-test-utilities/node-hooks/ignore-styles ./node_modules/mocha/bin/mocha.js --config ./.mocharc.json",
```
saskliutas, line 46: Does it work like this?
```suggestion
    "test:dev": "cross-env NODE_OPTIONS=\"--experimental-test-module-mocks --enable-source-maps --import presentation-test-utilities/node-hooks/ignore-styles\" mocha --config ./.mocharc.json",
```

Would avoid pionting into `./node_modules` directly and leave it up to npm/pnpm to find mocha executable.

grigasp: no, node complains that `--experimental-test-module-mocks` can't be passed through `NODE_OPTIONS` env var :(

saskliutas (approved):
Nice job entirely removing requirement to mock modules in majority of packages. 🥳

PR #898, opened 2025-03-06
Make sure workspace packages are used for sub dependencies

saskliutas:
This is necessary to make sure that regression tests are working in release branch where workspace packages versions are already bumped but not yet in npm registry.

~~Additionally bumped `mocha` to the latest version.~~ Some tests are failing. Reverted back.

@grigasp Can we use this option as default or should we enable it only in regression pipeline. One change this causes that I am aware of is that `presentation-frontend` used by test-app will use local `unified-selection` version.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -3,6 +3,7 @@ registry=https://registry.npmjs.org/
 resolution-mode=lowest-direct
 auto-install-peers=false
 package-manager-strict=false
+link-workspace-packages=deep
 
 hoist-pattern[]=!*
 hoist-pattern[]=@types/react*
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -202,8 +202,8 @@ catalogs:
       specifier: ^25.0.1
       version: 25.0.1
     mocha:
-      specifier: ^10.7.3
-      version: 10.7.3
+      specifier: ^11.1.0
+      version: 11.1.0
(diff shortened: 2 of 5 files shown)


grigasp:
> @grigasp Can we use this option as default or should we enable it only in regression pipeline. One change this causes that I am aware of is that `presentation-frontend` used by test-app will use local `unified-selection` version.

Do we know if `link-workspace-packages=deep` setting makes pnpm always override dependencies to our local ones, or only if they're compatible. E.g. if `presentation-frontend` has a dependency on `unified-selection@0.x`, will it get overwritten  with our local `1.x`?

saskliutas:
> > @grigasp Can we use this option as default or should we enable it only in regression pipeline. One change this causes that I am aware of is that `presentation-frontend` used by test-app will use local `unified-selection` version.
> 
> Do we know if `link-workspace-packages=deep` setting makes pnpm always override dependencies to our local ones, or only if they're compatible. E.g. if `presentation-frontend` has a dependency on `unified-selection@0.x`, will it get overwritten with our local `1.x`?

Did some testing and it looks like it uses workspace package only if versions are compatible. It didn't use `unified-selection` from workspace when `presentation-frontend` had it's version set to `0.5.1` or `1.0.0`. Workspace version was used when it set to `^1.0.0` or `^1.3.0`.

PR #902, opened 2025-03-12
Bump `itwinjs-core` version

saskliutas:
Bumped `itwinjs-core` version and resolved breaking changes.
Added handling for the cases where we need to support 4.x API removed in 5.x.
Updated regression test patches.


grigasp (commented):
I think we need change files for components & core-interop

grigasp (approved):

packages/components/src/presentation-components/common/IUnifiedSelectionComponent.ts
grigasp, line 1: I think `IUnifiedSelectionComponent` should be deprecated with the following message:
> All unified selection-driven components should use `SelectionStorage` from `@itwin/unified-selection` rather than `SelectionHandler`

apps/test-app/frontend/src/index.tsx
```diff
@@ -80,10 +80,14 @@ async function initializePresentation() {
   });
   // __PUBLISH_EXTRACT_END__
 
+  /* eslint-disable @typescript-eslint/no-deprecated */
```
grigasp, line 83: Doesn't look like we're using this extraction anymore and we probably don't need the active scope here. Let's remove it.

.changeset/sweet-horses-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": minor
+---
+
+Deprecated `IUnifiedSelectionComponent`. `SelectionStorage` from `@itwin/unified-selection` should be directly instead.
```
grigasp, line 5:
```suggestion
Deprecated `IUnifiedSelectionComponent`. `SelectionStorage` from `@itwin/unified-selection` should be used directly instead.
```

grigasp (approved):

PR #903, opened 2025-03-12
[Hierarchies-react]: filter action active state

MartynasStrazdas:
Changes:
- updated itwinUI alpha to 10
- added `ActiveDescription` to action items definition, when a value is provided a dot will appear above action.

Preview (Filter is applied):
image
Ps. no changes when filter is not used.

### DIFF
--- a/.changeset/gold-days-end.md
+++ b/.changeset/gold-days-end.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Added `activeDescription` optional property to TreeItemAction interface, which when provided a value will display a dot above the button indicating an active state. The provided value is used to set accesible description and should explain why the action item is active.
--- a/.changeset/gold-days-end.md
+++ b/.changeset/gold-days-end.md
@@ -3,5 +3,3 @@
 ---
 
 Added `activeDescription` optional property to TreeItemAction interface, which when provided a value will display a dot above the button indicating an active state. The provided value is used to set accesible description and should explain why the action item is active.
-
-Renamed `createFilterAction` to `useFilterAction`.
--- a/.changeset/gold-days-end.md
+++ b/.changeset/gold-days-end.md
@@ -3,3 +3,5 @@
 ---
 
 Added `activeDescription` optional property to TreeItemAction interface, which when provided a value will display a dot above the button indicating an active state. The provided value is used to set accesible description and should explain why the action item is active.
+
+Renamed `createFilterAction` to `useFilterAction`.
--- a/.changeset/wise-plants-try.md
+++ b/.changeset/wise-plants-try.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-hierarchies-react": patch
 ---
 
-Renamed `createFilterAction` to `useFilterAction`
+Changed `createFilterAction` function to a React hook and renamed it to `useFilterAction`.
--- a/.changeset/wise-plants-try.md
+++ b/.changeset/wise-plants-try.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Renamed `createFilterAction` to `useFilterAction`
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
@@ -9,12 +9,12 @@ import { HierarchyLevelDetails, useTree } from "../UseTree.js";
 import { LocalizedStrings, useLocalizationContext } from "./LocalizationContext.js";
 
 const filterIcon = new URL("@itwin/itwinui-icons/filter.svg", import.meta.url).href;
-const placeholderIcon = new URL("@itwin/itwinui-icons/placeholder.svg", import.meta.url).href; // TODO: active filter icon/placeholder icon if button was not given an icon
 
 /** @alpha */
 export interface TreeItemAction {
   label: string;
   action: () => void;
+  icon: string;
   /**
    * Determines action items visibility:
    * - `False` - action item is hidden.
@@ -28,9 +28,6 @@ export interface TreeItemAction {
    * If left undefined the action item will be rendered normally.
    */
   activeDescription?: string;
-
-  state?: string;
-  icon?: string;
 }
 
 /** @alpha */
@@ -46,9 +43,9 @@ export function TreeActionButton({ show, label, action, icon, activeDescription 
   const localizedDescription: string | undefined = localizedStrings[activeDescription as keyof typeof localizedStrings];
   return (
     <Tree.ItemAction
-      render={<IconButton dot={localizedDescription ?? activeDescription} icon={icon ?? placeholderIcon} variant={"ghost"} label={localizedLabel ?? label} />}
+      render={<IconButton dot={localizedDescription ?? activeDescription} icon={icon} variant={"ghost"} label={localizedLabel ?? label} />}
       label={localizedLabel ?? label}
-      icon={icon ?? placeholderIcon}
+      icon={icon}
       visible={show}
       onClick={action}
     />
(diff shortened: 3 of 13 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
```diff
@@ -22,6 +22,14 @@ export interface TreeItemAction {
...
    */
   show?: boolean;
+  /**
+   * Provide a value when action button is in active state to display a dot above the button.
+   * Provided text value is used to set accessible description it should descibe why the action is in active state.
+   * If left undefined the action item will be rendered normally.
+   */
+  activeDescription?: string;
+
+  state?: string;
```
grigasp, line 32: what does `state` mean?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
```diff
@@ -32,10 +40,19 @@ export type FilterActionProps = {
 } & Partial<Pick<ReturnType<typeof useTree>, "getHierarchyLevelDetails">>;
 
 /** @internal */
-export function TreeActionButton({ show, label, action, icon }: TreeItemAction) {
+export function TreeActionButton({ show, label, action, icon, activeDescription }: TreeItemAction) {
   const { localizedStrings } = useLocalizationContext();
   const localizedLabel: string | undefined = localizedStrings[label as keyof typeof localizedStrings];
-  return <Tree.ItemAction visible={show} onClick={action} label={localizedLabel ?? label} icon={icon ?? placeholderIcon} />;
+  const localizedDescription: string | undefined = localizedStrings[activeDescription as keyof typeof localizedStrings];
```
grigasp, line 46: I don't understand this. The docs for `activeDescription` prop say that it's alreadya user-friendly value that we can show, but here it's being used as a key for `localizedStrings`. Can you explain?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
```diff
@@ -32,10 +40,19 @@ export type FilterActionProps = {
...
+  return (
+    <Tree.ItemAction
+      render={<IconButton dot={localizedDescription ?? activeDescription} icon={icon ?? placeholderIcon} variant={"ghost"} label={localizedLabel ?? label} />}
+      label={localizedLabel ?? label}
+      icon={icon ?? placeholderIcon}
```
grigasp, line 51: Do we need to set both - `render` and `icon` props? It seems `render` should override whatever you pass to `icon`.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
```diff
@@ -32,10 +40,19 @@ export type FilterActionProps = {
...
 /** @internal */
-export function TreeActionButton({ show, label, action, icon }: TreeItemAction) {
+export function TreeActionButton({ show, label, action, icon, activeDescription }: TreeItemAction) {
   const { localizedStrings } = useLocalizationContext();
   const localizedLabel: string | undefined = localizedStrings[label as keyof typeof localizedStrings];
-  return <Tree.ItemAction visible={show} onClick={action} label={localizedLabel ?? label} icon={icon ?? placeholderIcon} />;
+  const localizedDescription: string | undefined = localizedStrings[activeDescription as keyof typeof localizedStrings];
+  return (
+    <Tree.ItemAction
+      render={<IconButton dot={localizedDescription ?? activeDescription} icon={icon ?? placeholderIcon} variant={"ghost"} label={localizedLabel ?? label} />}
```
grigasp, line 49: Do we still need `placeholderIcon`? I think we should make `icon` required and stop using the placeholder.

grigasp (changes requested):
> changed text variant from `body-sm` to `body-md` as it more resembles the previous default text size

We should target the new design rather than trying to match what we had in 3.x. As far as I can tell, the new design says we should use "sm":
- Design system sandbox:
   image
- Figma:
   image

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
MartynasStrazdas: Forgot to remove this, removed now.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
MartynasStrazdas: We provide functions for some buttons (filter, visibility), they need descriptions we can allow passing in descriptions when creating the button or we can provide a localizedString value. Because you would only want to change the description on different language, I believe using localized strings is better.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
MartynasStrazdas: This will be fixed by kiwi, because dot is not available for items actions, we need to use render and both IconButton and item action require an icon.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
MartynasStrazdas: I guess we don't need it, made icon required.

MartynasStrazdas:
> > changed text variant from `body-sm` to `body-md` as it more resembles the previous default text size
> 
> We should target the new design rather than trying to match what we had in 3.x. As far as I can tell, the new design says we should use "sm":
> 
> * Design system sandbox:
>   image
> * Figma:
>   image

You're right was looking at the wrong value, reverted back to sm

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeActionButton.tsx
MartynasStrazdas: If we are sure consumers are going to use `createUiItemsProviders` or `getWidgets` in a way that allows hooks i'm fine with it.

grigasp: They will, most likely, not going to call them from render functions. However, they can always render their own custom component that can use hooks and then render our trees.

MartynasStrazdas: Changed

grigasp: Looks good. Please add a changefile mentioning the `createFilterAction` -> `useFilterAction` change.

.changeset/gold-days-end.md
```diff
@@ -3,3 +3,5 @@
 ---
 
 Added `activeDescription` optional property to TreeItemAction interface, which when provided a value will display a dot above the button indicating an active state. The provided value is used to set accesible description and should explain why the action item is active.
+
+Renamed `createFilterAction` to `useFilterAction`.
```
grigasp, line 7: This will create a second paragraph to the `activeDescription` change. Instead, we want the rename to appear as a separate change in the changelog - you'll need to create a separate change file for that.

MartynasStrazdas: Did not know you can have multiple change files for the same package, changed to a separate one.

.changeset/wise-plants-try.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Renamed `createFilterAction` to `useFilterAction`
```
grigasp, line 5:
```suggestion
Changed `createFilterAction` function to a React hook and renamed it to `useFilterAction`.
```

grigasp (approved):

PR #906, opened 2025-03-18
[Hierarchies-react]: decorators

MartynasStrazdas:
closes https://github.com/iTwin/viewer-components-react/issues/1203
Preview:
image
Could not find a color picker like component in kiwi.
image

### DIFF
--- a/.changeset/good-actors-dance.md
+++ b/.changeset/good-actors-dance.md
@@ -14,7 +14,7 @@ function myGetIcon(node: PresentationHierarchyNode): React.ReactElement {
 <TreeNodeRenderer getIcon={myGetIcon} />
 
 // now:
-<TreeNodeRenderer getDecorations={[myGetIcon]} />
+<TreeNodeRenderer getDecorations={myGetIcon} />
 
 // previously, when `getIcon` returned an icon URI:
 function myGetIconUri(node: PresentationHierarchyNode): string {
@@ -23,5 +23,5 @@ function myGetIconUri(node: PresentationHierarchyNode): string {
 <TreeNodeRenderer getIcon={myGetIconUri} />
 
 // now:
-<TreeNodeRenderer getDecorations={[(node) => <Icon href={myGetIconUri(node)} />]} />
+<TreeNodeRenderer getDecorations={(node) => <Icon href={myGetIconUri(node)} />} />
 ```
--- a/.changeset/good-actors-dance.md
+++ b/.changeset/good-actors-dance.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-hierarchies-react": patch
 ---
 
-Replaced `getIcons` with `getDecorations`. `getDecorations` takes in an array of callbacks that return an element, previously used `getIcons` should be used with kiwi provided `<Icon />` component.
+Replaced `getIcons` with `getDecorations`. `getDecorations` takes in an array of callbacks  each returning an element. Previously provided `getIcons` can be reused with kiwi `<Icon />` component to provide icons to the tree.
--- a/.changeset/good-actors-dance.md
+++ b/.changeset/good-actors-dance.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Replaced `getIcons` with `getDecorations`. `getDecorations` takes in an array of callbacks that return an element, previously used `getIcons` should be used with kiwi provided `<Icon />` component.
--- a/packages/hierarchies-react/api/presentation-hierarchies-react.api.md
+++ b/packages/hierarchies-react/api/presentation-hierarchies-react.api.md
@@ -206,7 +206,7 @@ export const TreeNodeRenderer: ForwardRefExoticComponent<TreeNodeRendererProps &
 // @alpha (undocumented)
 interface TreeNodeRendererOwnProps {
     actions?: Array<(node: PresentationHierarchyNode) => TreeItemAction>;
-    getIcon?: (node: PresentationHierarchyNode) => string | ReactElement | undefined;
+    getDecorations?: Array<(node: PresentationHierarchyNode) => ReactElement | undefined>;
     getLabel?: (node: PresentationHierarchyNode) => ReactElement | undefined;
     getSublabel?: (node: PresentationHierarchyNode) => ReactElement | undefined;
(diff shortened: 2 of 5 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -154,8 +158,8 @@ export const TreeNodeRenderer: ForwardRefExoticComponent<TreeNodeRendererProps &
             onNodeKeyDown?.(node, !selected, event);
           }
         }}
-        icon={getIcon ? getIcon(node) : undefined}
         actions={getActions()}
+        unstable_decorations={getDecorations && [getDecorations.map((getDecoration) => getDecoration(node))]}
```
grigasp, line 162: How does this work? As far as I can tell, `unstable_decorations` should be a `React.ReactNode`, but we're passing `React.ReactNode[][]`.

packages/hierarchies-react/api/presentation-hierarchies-react.api.md
```diff
@@ -206,7 +206,7 @@ export const TreeNodeRenderer: ForwardRefExoticComponent<TreeNodeRendererProps &
 // @alpha (undocumented)
 interface TreeNodeRendererOwnProps {
     actions?: Array<(node: PresentationHierarchyNode) => TreeItemAction>;
-    getIcon?: (node: PresentationHierarchyNode) => string | ReactElement | undefined;
+    getDecorations?: Array<(node: PresentationHierarchyNode) => ReactElement | undefined>;
```
saskliutas, line 209: Is there any benefits taking array of callback producing react elements instead of single callback that produces array of elements?

.changeset/good-actors-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Replaced `getIcons` with `getDecorations`. `getDecorations` takes in an array of callbacks  each returning an element. Previously provided `getIcons` can be reused with kiwi `<Icon />` component to provide icons to the tree.
```
grigasp, line 5:
````suggestion
Replaced `TreeNodeRenderer.getIcon` prop with `getDecorations` prop.

The `getIcon` prop took either an icon URI, or a `ReactElement`. The new `getDecorations` prop allows passing multiple React elements, allowing consumers to render not just the icon, but also addition components like a tag or a color swatch. Migration:

```ts
// previously, when `getIcon` returned a `ReactElement`:
function myGetIcon(node: PresentationHierarchyNode): React.ReactElement {
  // implementation...
}
<TreeNodeRenderer getIcon={myGetIcon} />

// now:
<TreeNodeRenderer getDecorations={[myGetIcon]} />

// previously, when `getIcon` returned an icon URI:
function myGetIconUri(node: PresentationHierarchyNode): string {
  // implementation...
}
<TreeNodeRenderer getIcon={myGetIconUri} />

// now:
<TreeNodeRenderer getDecorations={[(node) => <Icon href={myGetIconUri(node)} />]} />
```
````

File: packages/hierarchies-react/api/presentation-hierarchies-react.api.md (hunk 2 above)
Comment (MartynasStrazdas) [reply packages/hierarchies-react/api/presentation-hierarchies-react.api.md:209]: I guess there is no big difference, only one being is that if we decide to provide decorations it would be more readable implementation on consumer end. Not sure which one would be better

File: packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx (hunk 1 above)
Comment (MartynasStrazdas) [reply packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx]: Removed left over [] from testing. It should now return `ReactElement[]`, don't really want to allow `ReactNode` as it includes too much unnecessary things.

File: packages/hierarchies-react/api/presentation-hierarchies-react.api.md (hunk 2 above)
Comment (grigasp) [reply packages/hierarchies-react/api/presentation-hierarchies-react.api.md:209]: > if we decide to provide decorations it would be more readable implementation on consumer end

Not sure I understand what you mean here. I personally like the single callback approach better.

Comment (MartynasStrazdas) [reply packages/hierarchies-react/api/presentation-hierarchies-react.api.md:209]: Basically having one callbacks will in the end produce the same structure as multiple callbacks, e.g. you have (node) => [functionForColorPicker(node), functionForIcon(node)] ==> [functionForColorPicker, functionForIcon]. In the case for TreeWidget we might need to expose out icons function if the user wants to have icons and colorpicker position fluid not static in right for the icons at all times.

Comment (saskliutas) [reply packages/hierarchies-react/api/presentation-hierarchies-react.api.md:209]: I think we should make `getDecorations` API match `TreeItem.unstable_decorations`. E.g. `getDecorations?: (node) => ReactNode`. 

Also instead of functions for color picker and icon we could provide components:
```tsx
getDecorations: (node) => <>
  <ColorPicker node={node} {...otherProps} />
  <SomeTreeIcon node={node} />
</>
```

MartynasStrazdas: Yeah forgot that they would be components, changed to a single callback and react node so the consumer can decide if to use an array or fragments.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -47,12 +46,18 @@ export interface TreeNodeRendererOwnProps {
    * Actions for tree item.
    */
   actions?: Array<(node: PresentationHierarchyNode) => TreeItemAction>;
+  /**
+   * Used to display elements between expander and label.
+   * Elements are arranged from left to right, with the last array element positioned on the right.
```
saskliutas, line 51: Looks like this comment need to be updated after return type was changed to `ReactNode`

saskliutas (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: Removed the part about element arrangement.

Issue #908, opened 2025-03-18
@itwin/unified-selection 1.4.0 with 5.0 dev dependencies ?


raplemie:
Hi, we just updated and saw that @itwin/unified-selection 1.4.0 was published with iTwin.js 5.0.0-dev dependencies, was that intended ?

grigasp:
Yes, we have peer deps on `^4.x || ^5.0` and build against 5.0-dev versions. We have regression pipelines to ensure we stay compatible with `4.x`. Did you notice any issues with that?

raplemie:
No, I'm just surprised that we published a released package that directly dependended on dev packages and thought it may have been an error: 

![Image](https://github.com/user-attachments/assets/d2539e7c-f946-4245-b210-9ce560f67e9c)

grigasp:
Oh, that might actually be a problem. I thought you're asking about `devDependencies`, but now I see you have actual production `dependencies` in mind. We shouldn't use `5.0-dev` for those, thanks for noticing - will look into that.

raplemie:
itwin/presentation-hierarchies-react seems to have the same update, I will not report other if I find them, I suppose they are all handled the same way as it was also released yesterday.

grigasp:
@raplemie, the fixes have been released for all affected packages: https://github.com/iTwin/presentation/pull/910#issue-2930941970

raplemie:
Thanks!

PR #909, opened 2025-03-18
Avoid using `dev` versions of `itwinjs-core` packages

saskliutas:
Closes https://github.com/iTwin/presentation/issues/908


grigasp (commented):
what do you think about having 2 catalogs for `itwinjs-core`: one for dev dependencies, and one for direct ones?

grigasp (approved):

PR #916, opened 2025-03-24
[hierarchies-react]: Fix nodes overlapping

saskliutas:
Fixed tree nodes overlapping if nodes with and without sublabel are rendered alongside.

### DIFF
--- a/.changeset/four-shoes-repeat.md
+++ b/.changeset/four-shoes-repeat.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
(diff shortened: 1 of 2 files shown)


grigasp (approved):
Mind a short explanation how the change helped?

saskliutas:
> Mind a short explanation how the change helped?

Looks like virtualizer uses item key to cache measured size. It defaults to the item index. I haven't looked too much into this bus I guess it was using incorrect sizes when node is collapsed and all nodes' indices shift. 

Also pushed change to use id instead of index as key because there were some problems with sibling nodes expander animating when node is collapsed/expanded.

PR #919, opened 2025-03-27
Bump deps

grigasp:


apps/load-tests/tests/src/processors/models-tree-stateless.ts
```diff
@@ -103,7 +103,7 @@ function createModelsTreeProvider(context: ScenarioContext, events: EventEmitter
       }
       try {
         const schemaInfo = await Schema.startLoadingFromJson(schemaJson, schemaContext);
-        if (schemaInfo !== undefined && schemaInfo.schemaKey.matches(schemaKey, matchType)) {
+        if (schemaInfo !== undefined && schemaInfo.schemaKey.matches(schemaKey as SchemaKey, matchType)) {
```
saskliutas, line 106: Can we change function definition from `schemaKey: Readonly<SchemaKey>` to `schemaKey: SchemaKey`?

apps/full-stack-tests/src/IModelUtils.ts
```diff
@@ -240,15 +240,15 @@ export function createSchemaContext(imodel: IModelConnection | IModelDb | ECDb)
       async getSchemaInfo(schemaKey: Readonly<SchemaKey>, matchType: SchemaMatchType, schemaContext: SchemaContext): Promise<SchemaInfo | undefined> {
         const schemaJson = imodel.getSchemaProps(schemaKey.name);
         const schemaInfo = await Schema.startLoadingFromJson(schemaJson, schemaContext);
-        if (schemaInfo !== undefined && schemaInfo.schemaKey.matches(schemaKey, matchType)) {
+        if (schemaInfo !== undefined && schemaInfo.schemaKey.matches(schemaKey as SchemaKey, matchType)) {
```
saskliutas, line 243: Can we change function definition from `schemaKey: Readonly<SchemaKey>` to `schemaKey: SchemaKey`?

apps/load-tests/tests/src/processors/models-tree-stateless.ts
grigasp: `ecschema-metadata@4.x` defines `ISchemaLocater` with readonly schema keys, changing them to non-readonly will make our regression tests fail to build. This is test code, so I think it's fine to use `as` as a workaround.

saskliutas (approved):

PR #920, opened 2025-03-31
[Hierarchies-react]: error node states

MartynasStrazdas:
Preview:
image
image


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -13,14 +13,21 @@ interface PlaceholderNode {
...
   posInLevel: number;
-} & TNode;
+} & PresentationHierarchyNode;
 
 /** @alpha */
-export type FlatTreeNode<TNode extends PresentationTreeNode = PresentationTreeNode> = FlatNode<TNode> | PlaceholderNode;
+export interface ErrorType {
+  errorNode: PresentationInfoNode;
+  parentNode?: PresentationHierarchyNode;
+  expandToNode: (expandNode: (nodeId: string) => void) => void;
```
saskliutas, line 26: Whay do you think about these renames?

```suggestion
export interface ErrorNode {
  error: PresentationInfoNode;
  parent?: PresentationHierarchyNode;
  expandTo: (expand: (nodeId: string) => void) => void;
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -49,3 +59,38 @@ function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
...
+
+export function getErrors(rootNodes: PresentationTreeNode[]) {
+  return rootNodes.flatMap((rootNode) => getErrorNodes(rootNode, []));
+}
+
+function getErrorNodes(parentNode: PresentationTreeNode, path: string[]) {
+  const errorList: ErrorType[] = [];
+  const newPath = [...path];
+
+  if (!isPresentationHierarchyNode(parentNode)) {
```
saskliutas, line 71: Looks like this code path is can only be hit if one of the root nodes is error node?

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -40,14 +42,33 @@ type TreeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode"> &
...
     rootNodes,
     selectNodes: selectNodes ?? noopSelectNodes,
     selectionMode: selectionMode ?? "single",
   });
+  const scrollToNode = useRef<string | undefined>(undefined);
   const parentRef = useRef<HTMLDivElement>(null);
   const flatNodes = useMemo(() => flattenNodes(rootNodes), [rootNodes]);
+  const errorList = useMemo(() => getErrors(rootNodes), [rootNodes]);
+
+  const isErrorNode = useCallback(
```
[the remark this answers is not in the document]
MartynasStrazdas: Renamed, seem to be more appropriate.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
MartynasStrazdas: Adjusted to similar, `expandTo` in my mind should continue to take `expandNode` as it is the name of the actual expand function we have in `TreeRenderer`

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
MartynasStrazdas: This has been adjusted

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -68,53 +71,32 @@ type TreeNodeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode" | "is
...
         return undefined;
       }
 
       if (actions.length < 4) {
-        return actions.map((action, index) => {
-          const actionInfo = action(node);
-          return <TreeActionButton key={index} {...actionInfo} />;
-        });
+        return [
+          ...actions.map((action, index) => {
```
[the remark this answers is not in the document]
MartynasStrazdas: Actions are going to change as we need to add a reload action on error nodes.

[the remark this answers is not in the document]
saskliutas: But `map` already creates a new array.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -20,162 +17,114 @@ export interface TreeErrorItemProps {
...
-function ErrorNodeLabel({ message, onRetry }: { message: string; onRetry?: () => void }) {
-  const { localizedStrings } = useLocalizationContext();
+function MessageWithNode({ errorNode, scrollToElement, message }: LinkedNodeProps & { message: string }) {
+  const splitMessage = message.split("{{node}}", 2);
   return (
-    <div style={{ display: "flex" }}>
-      <Text variant={"body-sm"}>{message}</Text>
-      {onRetry ? <Anchor onClick={onRetry}>{localizedStrings?.retry}</Anchor> : null}
+    <div>
+      {splitMessage[0]}
```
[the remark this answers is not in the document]
MartynasStrazdas: Yes, still working on it

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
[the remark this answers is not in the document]
MartynasStrazdas: Yes, this has been changed, what I mean is that because of errors requiring an additional action the structure of it might change.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -59,8 +80,37 @@ export function TreeRenderer({ rootNodes, expandNode, localizedStrings, selectNo
...
+    }
+    virtualizer.scrollToIndex(index);
+  };
+
+  useEffect(() => {
+    if (scrollToNode.current === undefined) {
+      return;
+    }
+    const index = flatNodes.findIndex((flatNode) => flatNode.id === scrollToNode.current);
+    virtualizer.scrollToIndex(index);
```
saskliutas, line 98: This should be after `if` statement.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -49,3 +59,41 @@ function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
...
+  }
+
+  parent.children.forEach((node) => {
+    if (!isPresentationHierarchyNode(node)) {
+      errorList.push({ parent, error: node, expandTo: (expandNode) => expandTo(expandNode, path) });
+      return;
+    }
+
+    if (node.children !== true) {
+      const childErrorList = getErrorNodes(node, [...path, ...(!parent.isExpanded ? [parent.id] : [])]);
```
saskliutas, line 89: You can avoid copying array multiple times for each child node with error.
```suggestion
  const pathToChild = [...path, ...(!parent.isExpanded ? [parent.id] : [])];
  parent.children.forEach((node) => {
    if (!isPresentationHierarchyNode(node)) {
      errorList.push({ parent, error: node, expandTo: (expandNode) => expandTo(expandNode, path) });
      return;
    }

    if (node.children !== true) {
      const childErrorList = getErrorNodes(node, pathToChild);
```

saskliutas (changes requested):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -20,162 +18,114 @@ export interface TreeErrorItemProps {
...
+function MessageWithNode({ errorNode, scrollToElement, message }: LinkedNodeProps & { message: string }) {
+  const splitMessage = message.split("{{node}}", 2);
   return (
-    <div style={{ display: "flex" }}>
-      <Text variant={"body-sm"}>{message}</Text>
-      {onRetry ? <Anchor onClick={onRetry}>{localizedStrings?.retry}</Anchor> : null}
+    <div>
+      <Text variant={"body-sm"}>{splitMessage[0]}</Text>
+      <LinkedNode errorNode={errorNode} scrollToElement={scrollToElement} />
+      <Text variant={"body-sm"}>{splitMessage[1]}</Text>
```
saskliutas, line 128: How does DOM look if `splitMessage[1]` is `undefined`?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -68,68 +73,42 @@ type TreeNodeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode" | "is
...
-      ...treeItemProps
-    },
+    { node, expandNode, getLabel, getSublabel, onNodeClick, onNodeKeyDown, isNodeSelected, actions, getDecorations, hasError, reloadTree, ...treeItemProps },
     forwardedRef,
   ) => {
     const nodeRef = useRef<HTMLElement>(null);
     const ref = useMergedRefs(forwardedRef, nodeRef);
+    const { localizedStrings } = useLocalizationContext();
+
+    const childErrorNode = useMemo(() => {
```
saskliutas, line 83: Do we need this `useMemo`? Without it you won't need to do additional check for placeholder node.

MartynasStrazdas: I don't think we expect the node to change much after loading so I think having it as useMemo is good. Quick test with console.log seems to show that useMemo saves some calls.

PR #928, opened 2025-04-07
[hierarchies-react]: Memoize components

saskliutas:
Memoized tree components to reduce the amount of renders when interacting with the tree.

### DIFF
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
@@ -3,12 +3,13 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { ComponentPropsWithoutRef, useCallback, useEffect, useMemo, useRef } from "react";
+import { ComponentPropsWithoutRef, forwardRef, memo, useCallback, useEffect, useMemo, useRef } from "react";
 import { Tree } from "@itwin/itwinui-react/bricks";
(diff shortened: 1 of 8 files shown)


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -119,26 +126,26 @@ export function TreeRenderer({
       <div style={{ height: "100%", width: "100%", overflowY: "auto" }} ref={parentRef}>
         <Tree.Root style={{ height: virtualizer.getTotalSize(), minHeight: "100%", width: "100%", position: "relative", overflow: "hidden" }}>
           {items.map((virtualizedItem) => {
+            const selected = isNodeSelected?.(flatNodes[virtualizedItem.index].id) ?? false;
+            const error = hasError(flatNodes[virtualizedItem.index].id);
```
MartynasStrazdas, line 130: We could check if `flatNodes[virtualizedItem.index]` has children, if they don't, we could pass `undefined`, and return `undefined` from `hasError`, this should avoid checks for nodes we know don't have errors.

MartynasStrazdas (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
saskliutas: I don't think it has much value. I believe error list usually will have only a few elements ~2-5 so it shouldn't be expensive. Also it would require to do a typeguard check to make sure it is a hierarchy node before checking for children.

PR #930, opened 2025-04-08
[Hierarchies-react]: expose custom error rendering & transform flat nodes/error functions to hooks

MartynasStrazdas:
Preview:
image

Related issue:
https://github.com/orgs/iTwin/projects/20/views/14?pane=issue&itemId=98495356&issue=iTwin%7Cviewer-components-react%7C1194
image

### DIFF
--- a/.changeset/rude-rivers-post.md
+++ b/.changeset/rude-rivers-post.md
@@ -1,0 +1,6 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Exposed `TreeErrorRenderer`.
+`TreeRenderer` now takes `renderError` to render custom errors.
--- a/.changeset/tame-years-make.md
+++ b/.changeset/tame-years-make.md
@@ -1,0 +1,7 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Changed flat tree building functions to hooks.
+`flattenNodes` => `useFlattenNodes`.
+`getErrors` => `useErrorList`.
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
@@ -3,6 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+import { useMemo } from "react";
 import { isPresentationHierarchyNode, PresentationHierarchyNode, PresentationInfoNode, PresentationTreeNode } from "../TreeNode.js";
 
 /** @alpha */
@@ -35,8 +36,8 @@ export function isPlaceholderNode(node: FlatTreeNode): node is PlaceholderNode {
 }
 
 /** @alpha */
-export function flattenNodes(rootNodes: PresentationTreeNode[]) {
-  return getFlatNodes(rootNodes, 1);
+export function useFlattenNodes(rootNodes: PresentationTreeNode[]) {
+  return useMemo(() => getFlatNodes(rootNodes, 1), [rootNodes]);
 }
 
 function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
@@ -60,16 +61,21 @@ function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
   return flatNodes;
 }
 
-export function getErrors(rootNodes: PresentationTreeNode[]): ErrorNode[] {
-  return rootNodes.flatMap((rootNode) => {
-    if (!isPresentationHierarchyNode(rootNode)) {
-      return [{ parent: undefined, error: rootNode, expandTo: (expandNode) => expandTo(expandNode, []) }];
-    }
-    if (rootNode.children === true) {
-      return [];
-    }
-    return getErrorNodes(rootNode, !rootNode.isExpanded ? [rootNode.id] : []);
-  });
+/** @alpha */
+export function useErrorList(rootNodes: PresentationTreeNode[]): ErrorNode[] {
+  return useMemo(
+    () =>
+      rootNodes.flatMap((rootNode) => {
+        if (!isPresentationHierarchyNode(rootNode)) {
+          return [{ parent: undefined, error: rootNode, expandTo: (expandNode) => expandTo(expandNode, []) }];
+        }
+        if (rootNode.children === true) {
+          return [];
+        }
+        return getErrorNodes(rootNode, !rootNode.isExpanded ? [rootNode.id] : []);
+      }),
+    [rootNodes],
+  );
 }
 
 function getErrorNodes(parent: PresentationHierarchyNode, path: string[]) {
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
@@ -3,6 +3,7 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
+import { ReactElement } from "react";
 import { Anchor, unstable_ErrorRegion as ErrorRegion, Text } from "@itwin/itwinui-react/bricks";
 import { MAX_LIMIT_OVERRIDE } from "../internal/Utils.js";
 import { HierarchyLevelDetails, useTree } from "../UseTree.js";
@@ -10,25 +11,34 @@ import { ErrorNode } from "./FlatTreeNode.js";
 import { useLocalizationContext } from "./LocalizationContext.js";
 
 /** @alpha */
-export interface TreeErrorItemProps {
+interface TreeErrorItemProps {
   /** A callback to reload a hierarchy level when an error occurs and `retry` button is clicked. */
   reloadTree?: (options: { parentNodeId: string | undefined; state: "reset" }) => void;
   /** Action to perform when the filter button is clicked for this node. */
   onFilterClick?: (hierarchyLevelDetails: HierarchyLevelDetails) => void;
 }
-
+/** @alpha */
 interface TreeErrorRendererOwnProps {
+  /** List of errors to be displayed */
   errorList: ErrorNode[];
+  // Callback to render custom error messages. Component should be wrapped in `ErrorRegion.Item` from itwinUI.
+  renderError?: ({ error, scrollToElement }: { error: ErrorNode } & Pick<LinkedNodeProps, "scrollToElement">) => ReactElement;
 }
 
-type TreeErrorRendererProps = TreeErrorRendererOwnProps &
+/** @alpha */
+export type TreeErrorRendererProps = TreeErrorRendererOwnProps &
   TreeErrorItemProps &
   Partial<Pick<ReturnType<typeof useTree>, "getHierarchyLevelDetails">> &
   Pick<LinkedNodeProps, "scrollToElement">;
 
-export function TreeErrorRenderer({ errorList, reloadTree, scrollToElement, getHierarchyLevelDetails, onFilterClick }: TreeErrorRendererProps) {
+/** @alpha */
+export function TreeErrorRenderer({ errorList, reloadTree, scrollToElement, getHierarchyLevelDetails, onFilterClick, renderError }: TreeErrorRendererProps) {
   const { localizedStrings } = useLocalizationContext();
   const errorItems = errorList.map((errorNode) => {
+    if (renderError) {
+      return renderError({ error: errorNode, scrollToElement });
+    }
+
     if (errorNode.error.type === "ResultSetTooLarge") {
       return (
         <ResultSetTooLarge
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
@@ -10,9 +10,9 @@ import { PresentationTreeNode } from "../TreeNode.js";
 import { SelectionMode, useSelectionHandler } from "../UseSelectionHandler.js";
 import { useTree } from "../UseTree.js";
 import { useEvent } from "../Utils.js";
-import { ErrorNode, flattenNodes, getErrors } from "./FlatTreeNode.js";
+import { ErrorNode, useErrorList, useFlattenNodes } from "./FlatTreeNode.js";
 import { LocalizationContextProvider } from "./LocalizationContext.js";
-import { TreeErrorItemProps, TreeErrorRenderer } from "./TreeErrorRenderer.js";
+import { TreeErrorRenderer, TreeErrorRendererProps } from "./TreeErrorRenderer.js";
 import { TreeNodeRenderer } from "./TreeNodeRenderer.js";
 
 /** @alpha */
@@ -31,7 +31,7 @@ interface TreeRendererOwnProps {
 /** @alpha */
 type TreeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode"> &
   Partial<Pick<ReturnType<typeof useTree>, "selectNodes" | "isNodeSelected" | "getHierarchyLevelDetails" | "reloadTree">> &
-  Pick<TreeErrorItemProps, "onFilterClick"> &
+  Pick<TreeErrorRendererProps, "onFilterClick" | "renderError"> &
   Omit<TreeNodeRendererProps, "node" | "reloadTree" | "selected" | "error"> &
   TreeRendererOwnProps &
   ComponentPropsWithoutRef<typeof LocalizationContextProvider>;
@@ -57,6 +57,7 @@ export function TreeRenderer({
   getDecorations,
   getLabel,
   getSublabel,
+  renderError,
   ...treeProps
 }: TreeRendererProps) {
   const { onNodeClick, onNodeKeyDown } = useSelectionHandler({
@@ -66,8 +67,8 @@ export function TreeRenderer({
   });
   const scrollToNode = useRef<string | undefined>(undefined);
   const parentRef = useRef<HTMLDivElement>(null);
-  const flatNodes = useMemo(() => flattenNodes(rootNodes), [rootNodes]);
-  const errorList = useMemo(() => getErrors(rootNodes), [rootNodes]);
+  const flatNodes = useFlattenNodes(rootNodes);
+  const errorList = useErrorList(rootNodes);
 
   const handleNodeClick = useEvent(onNodeClick);
   const handleKeyDown = useEvent(onNodeKeyDown);
@@ -117,6 +118,7 @@ export function TreeRenderer({
   return (
     <LocalizationContextProvider localizedStrings={localizedStrings}>
       <TreeErrorRenderer
+        renderError={renderError}
         errorList={errorList}
         scrollToElement={scrollToElement}
         getHierarchyLevelDetails={getHierarchyLevelDetails}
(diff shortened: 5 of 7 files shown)


.changeset/tame-years-make.md
```diff
@@ -0,0 +1,7 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Changed flat tree building functions to hooks.
+`flattenNodes` => `useFlattenNodes`.
+`getErrors` => `useErrorList`.
```
grigasp, line 7:
```suggestion
---
"@itwin/presentation-hierarchies-react": minor
---

Changed flat tree building functions to hooks:
- `flattenNodes` => `useFlattenNodes`.
- `getErrors` => `useErrorList`.
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -35,8 +36,8 @@ export function isPlaceholderNode(node: FlatTreeNode): node is PlaceholderNode {
 }
 
 /** @alpha */
-export function flattenNodes(rootNodes: PresentationTreeNode[]) {
-  return getFlatNodes(rootNodes, 1);
+export function useFlattenNodes(rootNodes: PresentationTreeNode[]) {
```
grigasp, line 39: Maybe rename to `useFlatTreeNodesList`? Also, I think it's about time we start incrementally adding API docs to the new future-public functions like this.

.changeset/rude-rivers-post.md
```diff
@@ -0,0 +1,6 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
```
grigasp, line 3:
```suggestion
---
"@itwin/presentation-hierarchies-react": minor
---
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -3,32 +3,42 @@
...
   reloadTree?: (options: { parentNodeId: string | undefined; state: "reset" }) => void;
   /** Action to perform when the filter button is clicked for this node. */
   onFilterClick?: (hierarchyLevelDetails: HierarchyLevelDetails) => void;
 }
-
+/** @alpha */
 interface TreeErrorRendererOwnProps {
+  /** List of errors to be displayed */
   errorList: ErrorNode[];
+  // Callback to render custom error messages. Component should be wrapped in `ErrorRegion.Item` from itwinUI.
```
grigasp, line 24:
```suggestion
  // Callback to render custom error messages. Component should be wrapped in `ErrorRegion.Item` from `@itwin/itwinui-react` package.
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -57,6 +57,7 @@ export function TreeRenderer({
   getDecorations,
   getLabel,
   getSublabel,
+  renderError,
```
grigasp, line 60: Maybe `TreeRenderer`, instead of taking renderer of a single error item, could take renderer of the whole errors block, and then the default `TreeErrorRenderer` allows passing renderer for a single error item. That way, consumers would be able to supply custom renderers to both - the whole error block, or a single error item. What do you think?

grigasp: > this way they can set their custom message without rewriting the existing structure

Not sure what you mean by "rewriting the existing structure". With current changes, passing a custom error renderer looks like this:
```ts
<TreeRenderer
  renderError={(errorProps) => <MyErrorRenderer {...errorProps} />}
/>
```
With my suggestion that changes to this (if you want our errors' block renderer):
```ts
<TreeRenderer
  errorsRenderer={(errorsRendererProps) => <TreeErrorRenderer {...errorsRendererProps} renderError={(errorProps) => <MyErrorRenderer {...errorProps} />}
/>
```
This is a common practice used elsewhere in our components, it allows you to replace components at different levels without overly polluting the API.

> If the consumer would like to have their own `errorRenderer` they can reuse our `nodeRenderer` and build their own `treeRenderer`.

Yes, but you could say the same about renderer of a single error, too, right? 

> Allowing a custom `errorRenderer` would mean it still has to be some kind of dropdown which would be limiting and would make changing the error message to a custom one more tedious.

Why does it have to be a dropdown? I think replacing any kind of internal renderer is an "advanced" level stuff, and it should be expected that you may need to supply a few levels of default renderers to get to the one you want to replace.

MartynasStrazdas: I guess this would work. 

What I meant that it should be a dropdown is because the error renderer shares space with the tree. The shared space is not scrollable, because of that the number of different components that could hold errors are limited.

Will update to match the example.

MartynasStrazdas: Adjusted, updated change file.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
MartynasStrazdas: changed to `useFlatTreeNodeList`

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -6,37 +6,59 @@
...
 
 /** @alpha */
 export function isPlaceholderNode(node: FlatTreeNode): node is PlaceholderNode {
   return "placeholder" in node;
 }
 
-/** @alpha */
-export function useFlattenNodes(rootNodes: PresentationTreeNode[]) {
+/**
+ * Used to get `flatTreeNodes` for the tree using root nodes.
```
grigasp, line 57:
```suggestion
 * Used to get a list of `FlatTreeNode` objects for the given list of `PresentationTreeNode` objects that represent
 * a hierarchical structure. The resulting nodes can be used to render the hierarchy in a flat manner, e.g. using a
 * virtualized list.
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -61,7 +83,11 @@ function getFlatNodes(nodes: PresentationTreeNode[], level: number) {
   return flatNodes;
 }
 
-/** @alpha */
+/**
+ * Used to get error list from all loaded nodes.
```
grigasp, line 87:
```suggestion
 * Finds and returns all error nodes in a given hierarchy in the form of `PresentationTreeNode[]`.
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
grigasp, line 1: Probably not for this PR, but I'd like to review the APIs exported in this source file. I find several things confusing:

- Some types called "node" (`PlaceholderNode` & `FlatNode`) have attributes like "level". IMO, a "node" should not care about "level". Maybe we should have a type like `FlattenedHierarchicalItem<TNode> { level: number; node: TNode }`?..

- `ErrorNode` has an attribute `error` which is also a "node". It's not clear which one I should use in which situations.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
```diff
@@ -6,37 +6,59 @@
...
 export interface ErrorNode {
   error: PresentationInfoNode;
   parent?: PresentationHierarchyNode;
   expandTo: (expandNode: (nodeId: string) => void) => void;
 }
 
-/** @alpha */
+/**
+ * A node describing single tree item position and its content inside tree.
+ * Returned by `useFlatNodesList` hook.
```
grigasp, line 45:
```suggestion
 * Returned by `useFlatTreeNodeList` hook.
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -26,12 +26,14 @@ interface TreeRendererOwnProps {
   rootNodes: PresentationTreeNode[];
   /** Active selection mode used by the tree. Defaults to `"single"`. */
   selectionMode?: SelectionMode;
+  /** Callback for providing a custom implementation of error display component */
```
grigasp, line 29:
```suggestion
  /** A render function for errors' display component. Defaults to `<TreeErrorRenderer />`. */
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -3,32 +3,42 @@
...
   /** Action to perform when the filter button is clicked for this node. */
   onFilterClick?: (hierarchyLevelDetails: HierarchyLevelDetails) => void;
 }
-
+/** @alpha */
 interface TreeErrorRendererOwnProps {
+  /** List of errors to be displayed */
   errorList: ErrorNode[];
+  // Callback to render custom error messages. Component should be wrapped in `ErrorRegion.Item` from `@itwin/itwinui-react` package.
+  renderError?: ({ errorNode, scrollToElement }: { errorNode: ErrorNode } & Pick<LinkedNodeProps, "scrollToElement">) => ReactElement;
```
grigasp, line 25: couldn't `{ errorNode: ErrorNode } & Pick<LinkedNodeProps, "scrollToElement">` be replaced with just `LinkedNodeProps`?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -120,6 +130,7 @@ function ResultSetTooLarge({ errorNode, onFilterClick, limit, onOverrideLimit, s
   );
 }
 
+/** @alpha */
 interface LinkedNodeProps {
   errorNode: ErrorNode;
   scrollToElement: (node: ErrorNode) => void;
```
grigasp, line 136: - Why does this type has to take a node + a function that expects the same node as an argument? It seems either it shouldn't contain a node prop, or the function shouldn't take an argument.

- The type's name doesn't say anything about errors, but the attributes say it takes an "error node". Is this type intended only for errors, or can it be used for any kind of node? In case of the former, I think the type should be renamed. In case of the latter, I think the type should be changed to this:
   ```ts
   interface LinkedNodeProps<TNode> {
     node: TNode;
     scrollToNode: () => void;
   }
   ```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/FlatTreeNode.ts
MartynasStrazdas: We could call all `nodes`  that have interactions with the tree layout items. `FlatTreeNode` => `FlatTreeItem`, `ErrorNode` => `ErrorItem`. This would make more sense as they are used in kiwi tree as items, these items could then be composed of single or multiple nodes like in the case with `ErrorItem/ErrorNode`. 

I guess item interface could be generic, but only `PresentationHierarchyNode` and `PlaceholderNode` would use it and in the case of `PlaceholderNode` it only needs level as other parts are constants. This would leave `ErrorNode/ErrorItem` in a wierd place.

But I agree we should look into the naming more in a separate issue.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
MartynasStrazdas: Changed the structure of error renderer, this part was fixed.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -3,61 +3,105 @@
...
       );
     }
-    return <DefaultErrorContainer key={errorNode.error.id} scrollToElement={scrollToElement} reloadTree={reloadTree} errorNode={errorNode} />;
+    return (
+      <ErrorItemContainer
+        key={errorNode.error.id}
+        errorNode={errorNode}
+        actions={[
+          {
+            action: () => reloadTree?.({ parentNodeId: errorNode.parent?.id, state: "reset" }),
```
MartynasStrazdas, line 96: @grigasp @saskliutas In the new error workflow failing to create a hierarchy displays retry button in errorContainer and retry action near the node, but the reload action itself might not be passed which could cause confusion on how to reload the tree. Should we make it required? Ps. you can still reload the tree with closing the expander and reopening it.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -112,23 +112,22 @@ const HierarchyNode = memo(
...
+            action={() => reloadTree({ parentNodeId: node.id, state: "reset" })}
             show={true}
             icon={refreshSvg}
           />,
         );
       }
 
+      if (!actions || actions.length === 0) {
+        return actionButtons;
+      }
```
grigasp, line 129: Do you think this is necessary? The below will be a no-op anyway if `actions` is empty.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -79,73 +113,44 @@ export function TreeErrorRenderer({ errorList, reloadTree, scrollToElement, getH
...
 }
 
-function MessageWithNode({ errorNode, scrollToElement, message }: LinkedNodeProps & { message: string }) {
+function MessageWithLink({ linkLabel, scrollToElement, message }: MessageWithLinkProps) {
   const splitMessage = message.split("{{node}}", 2);
   return (
     <div style={{ display: "flex", whiteSpace: "pre", flexWrap: "wrap" }}>
       <Text variant={"body-sm"}>{splitMessage[0]}</Text>
-      <LinkedNode errorNode={errorNode} scrollToElement={scrollToElement} />
+      <Anchor onClick={scrollToElement}>{linkLabel}</Anchor>
```
grigasp, line 153: should we render this if `linkLabel` is `undefined`?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
grigasp: I wouldn't mind it being required, especially since `useTree` always returns this function.

packages/hierarchies-react/api/presentation-hierarchies-react.api.md
```diff
@@ -197,8 +194,25 @@ interface TreeErrorItemProps {
         parentNodeId: string | undefined;
         state: "reset";
     }) => void;
+    scrollToElement: (errorNode: ErrorNode) => void;
+}
+
+// @alpha (undocumented)
```
grigasp, line 200: mind adding docs to these new apis?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
MartynasStrazdas: We don't really expect it to be undefined, but with current setup Root Node's parent is undefined, but I don't think it's possible to have rootNode be an errorItem.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: I think we still need to check that actions are not undefined, because you can't call map on undefined even with `?`.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
grigasp: > I don't think it's possible to have rootNode be an errorItem.

wouldn't that be the case if we fail to load the root level (e.g. too many root nodes)?

MartynasStrazdas: Started working on a root error renderer, seems like it would be more suited for a separate issue/PR, we'd need to figure out how we want to handle root node filtering as we currently don't have an ability to do so.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
MartynasStrazdas: Made it required.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
grigasp: Please file an issue and link it here.

MartynasStrazdas: https://github.com/iTwin/viewer-components-react/issues/1290

grigasp (approved):
LGTM, but I'd like @saskliutas review too.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -114,15 +118,17 @@ export function TreeRenderer({
...
   return (
     <LocalizationContextProvider localizedStrings={localizedStrings}>
-      <TreeErrorRenderer
-        errorList={errorList}
-        scrollToElement={scrollToElement}
-        getHierarchyLevelDetails={getHierarchyLevelDetails}
-        onFilterClick={onFilterClick}
-        reloadTree={reloadTree}
-      />
+      {errorRenderer ? errorRenderer({ ...errorRendererProps }) : <TreeErrorRenderer {...errorRendererProps} />}
```
saskliutas, line 131:
```suggestion
      {errorRenderer ? errorRenderer(errorRendererProps) : <TreeErrorRenderer {...errorRendererProps} />}
```

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeErrorRenderer.tsx
```diff
@@ -69,72 +126,44 @@ export function TreeErrorRenderer({ errorList, reloadTree, scrollToElement, getH
...
-              {localizedStrings.increaseHierarchyLimit.replace("{{limit}}", MAX_LIMIT_OVERRIDE.toString())}
-            </Anchor>
+          {actions?.map(
+            (actionItem) =>
+              actionItem.condition() && (
+                <Anchor key={actionItem.label} onClick={actionItem.action}>
+                  {actionItem.label}
+                </Anchor>
+              ),
           )}
```
saskliutas, line 148:
```suggestion
          {actions
            ?.filter(({ condition }) => condition())
            .map(({ label, action }) => (
              <Anchor key={label} onClick={action}>
                {label}
              </Anchor>
            ))}
```

saskliutas (approved):

PR #936, opened 2025-04-16
[hierarchies-react]: Update itwinui-react version

saskliutas:
Updated `itwinui-react` version to the latest that improves Tree components memoization.
Restructured Tree component delivered by `@itwin/presentation-hierarchies-react` package to remove unnecessary wrappers and to better separate virtualization related logic from rendering logic.
Unified actions and decorations handling to be more intuitive.


.changeset/ripe-jars-pull.md
```diff
@@ -0,0 +1,35 @@
+---
+"@itwin/presentation-hierarchies-react": minor
```
grigasp, line 2:
```suggestion
"@itwin/presentation-hierarchies-react": major
```

.changeset/ripe-jars-pull.md
```diff
@@ -0,0 +1,35 @@
...
+After:
+```tsx
+import { Icon, Tree } from "@itwin/itwinui-react";
+import { FilterAction, TreeNodeRenderer } from "@itwin/presentation-hierarchies-react";
+
+return <TreeNodeRenderer
+  actions={(node) => [
+    <FilterAction key="filter" node={node} onFilter={setFilteringOptions} getHierarchyLevelDetails={treeProps.getHierarchyLevelDetails} />,
+    <Tree.ItemAction key="customAction" label="Custom action" onClick={() => log(node.label)} icon={<Icon href={customIconHref} />} />,
+  ]}
```
grigasp, line 33: - `actions` -> `getActions`
- should this be in a `useCallback`?

grigasp (commented):
LGTM. @MartynasStrazdas, please review

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -34,7 +33,7 @@ type TreeNodeProps = ComponentPropsWithoutRef<typeof Tree.Item>;
 /** @alpha */
 export interface TreeNodeRendererOwnProps {
   /** Node that is rendered. */
-  node: FlatTreeNode;
+  node: PresentationHierarchyNode;
```
MartynasStrazdas, line 36: Is there a reason to move item position within the tree outside of the node? Do you think it would make the API more flexible?

packages/hierarchies-react/api/presentation-hierarchies-react.api.md
```diff
@@ -233,25 +222,24 @@ export const TreeNodeRenderer: FC<PropsWithRef<TreeNodeRendererProps & RefAttrib
 
 // @alpha (undocumented)
 interface TreeNodeRendererOwnProps {
-    actions?: Array<(node: PresentationHierarchyNode) => TreeItemAction>;
     error?: ErrorNode;
+    getActions?: (node: PresentationHierarchyNode) => ReactNode[];
```
MartynasStrazdas, line 226: We might as well make this a `ReactElement[]` as we expect a `Tree.ItemAction`.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
```diff
@@ -75,82 +72,44 @@ type TreeNodeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode"> &
...
-        aria-level={node.level}
-        aria-posinset={node.posInLevel}
-        aria-setsize={node.levelSize}
-        label={useMemo(() => (getLabel ? getLabel(node) : node.label), [getLabel, node])}
-        description={useMemo(() => (getSublabel ? getSublabel(node) : undefined), [getSublabel, node])}
+        label={label}
+        description={description}
         selected={selected}
-        expanded={node.isExpanded || node.children === true || node.children.length > 0 ? node.isExpanded : undefined}
+        expanded={node.isExpanded || node.children ? node.isExpanded : undefined}
```
MartynasStrazdas, line 112: This seems to add an expander to last nodes with no children.

grigasp: If the node doesn't have children (`node.children` is an empty array), then I guess we don't care about whether the expander is set to expanded or collapsed mode, since it shouldn't be displayed in the first place. However, I'm not sure what controls its display 🤔

saskliutas: @MartynasStrazdas Yes you are right. I think this was a leftover from testing different memoization strategies. I will revert it.

packages/hierarchies-react/api/presentation-hierarchies-react.api.md
saskliutas: I wanted to match type of `actions` property on `Tree.Item`.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: If we don't want an expander we need to set it to undefined, which in this case does not happen.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
saskliutas: I wanted to keep virtualization related code in one place. I don't think it changes API flexibility.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
grigasp: > `expanded` have to be set to `undefined` in order for expander to be hidden.

wasn't expecting that 🤷‍♂️

packages/hierarchies-react/api/presentation-hierarchies-react.api.md
MartynasStrazdas: This just feels a bit weird cause `ReactNode` already includes ReactElement[], and `ReactNode[]` is a bit ambiguous on what it actually means. But I guess its fine.

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeNodeRenderer.tsx
MartynasStrazdas: We should probably discuss this later when we discuss the whole api as we seem to keep jumping from this to node, back to this.

MartynasStrazdas (approved):

packages/hierarchies-react/api/presentation-hierarchies-react.api.md
saskliutas: Or both should use `ReactNode` as it's union definition also includes `Iterable<ReactNode>`.

Issue #932, opened 2025-04-10
`hierarchies-react`: Review API


grigasp:
We could call all `nodes`  that have interactions with the tree layout `items`. `FlatTreeNode` => `FlatTreeItem`, `ErrorNode` => `ErrorItem`. This would make more sense as they are used in kiwi tree as items, these items could then be composed of single or multiple nodes like in the case with `ErrorItem/ErrorNode`. 

I guess item interface could be generic, but only `PresentationHierarchyNode` and `PlaceholderNode` would use it and in the case of `PlaceholderNode` it only needs level as other parts are constants. This would leave `ErrorNode/ErrorItem` in a wierd place.

But I agree we should look into the naming more in a separate issue.

_Originally posted by @MartynasStrazdas in https://github.com/iTwin/presentation/pull/930#discussion_r2036819429_

grigasp:
Adding another case that needs a review: https://github.com/iTwin/viewer-components-react/pull/1289#discussion_r2041849731.

We need to consistently name our filtering APIs. So far we discussed about the following naming:
- Hierarchy-wide filtering should be called "search". E.g. `filter` -> `searchText`, `getFilteredPaths` -> `getSearchPaths`, etc.
- Hierarchy-level filtering should stay as is, but we should make sure it's called that way in all the public APIs and docs.

grigasp:
Another one: https://github.com/iTwin/presentation/pull/936#discussion_r2048456732

grigasp:
Assigning this to myself as I was one of the people on the team who worked on this package the least. I'll overview the public API and raise problems, then we can discuss solutions.

grigasp:
One more: https://github.com/iTwin/presentation/pull/937#discussion_r2055868100

grigasp:
Here's what I come up with:

- [ ] `PresentationResultSetTooLargeInfoNode` is exported through barrel, but `PresentationNoFilterMatchesInfoNode` is not. Probably becomes irrelevant if we resolve the below issue.

- [ ] Now that we render errors elsewhere, as opposed to rendering them in place of child nodes that we did in 3.x, it probably doesn't make much sense to create a "node" for an error.
  - We should consider removing `PresentationInfoNode` with all its sub-types and instead assign error information to the parent node.
  - If error information is assigned to the parent node, `ErrorNode` should become unnecessary. `useErrorList` could return `PresentationTreeNode` objects with non-optional `error` property, containing the error details.
[...]

PR #937, opened 2025-04-17
[Hierarchies-react]: root error renderer

MartynasStrazdas:
Preview:
Root error
image
image

Custom root error:
image


packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
```diff
@@ -47,31 +50,46 @@ type TreeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode"> &
  * @see https://itwinui.bentley.com/docs/tree
  * @alpha
  */
-export function TreeRenderer({
-  rootNodes,
+export function TreeRenderer({ rootNodes, selectNodes, selectionMode, rootErrorRenderer, ...treeProps }: TreeRendererProps) {
```
MartynasStrazdas, line 53: Currently we allow `onNodeClick` and `onNodeKeyDown` to be passed into TreeRenderer which currently does nothing. How do you think we should handle this omit it or allow it to be passed and included in our `onNodeClick/onNodeKeyDown`? @saskliutas @grigasp 

We also allow various random properties like `onClick` which make no sense we should open an issue to clean up the API.

grigasp: > we should open an issue to clean up the API

we already have one - added a link to this comment to make sure it's covered: https://github.com/iTwin/presentation/issues/932#issuecomment-2826516743

saskliutas (approved):

.changeset/stupid-otters-listen.md
```diff
@@ -0,0 +1,33 @@
...
+
+Custom error message example:
+
+```ts
+<TreeRenderer
+    {...treeProps}
+    rootErrorRenderer={(rootErrorRendererProps) => (
+        <RootErrorRenderer
+            {...rootErrorRendererProps}
+            renderRootError={(rootErrorProps) => {
```
grigasp, line 27: Just a thought... seems a bit of a repetition to have `renderRootError` prop on `RootErrorRenderer` component. Maybe rename the prop to `renderError` or `renderErrorItem`?

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/RootErrorRenderer.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * A component that renders root node error state.
+ *
+ * @alpha
+ */
+export function RootErrorRenderer({ errorNode, renderRootError, getHierarchyLevelDetails, reloadTree }: RootErrorRendererProps) {
+  const { localizedStrings } = useLocalizationContext();
+
+  if (renderRootError) {
+    return renderRootError({ errorNode });
+  }
```
grigasp, line 34: What's the point of having `renderRootError`? It seems whoever would pass that prop could simply completely skip `RootErrorRenderer` and just render their own component. Or am I missing something?

MartynasStrazdas: I wanted to keep the rootErrorRenderer as close as errorRenderer so in a way you could provide the same custom component to both.

saskliutas: Wouldn't styling be broken if you pass component used in `renderError` to the `rootErrorRenderer.renderError`? Because in case of `ErrorRenderer` it used that component for each error item, while in `RootErrorRenderer` only that custom component is rendered.

grigasp: I just don't see any value in it... From your example:

```ts
<TreeRenderer
    {...treeProps}
    rootErrorRenderer={(rootErrorRendererProps) => (
        <RootErrorRenderer
            {...rootErrorRendererProps}
            renderRootError={(rootErrorProps) => {
                return <>My error</>;
            }}
        />
    )}
/>
```

You could instead just do this to get the same result:

```ts
<TreeRenderer
    {...treeProps}
    rootErrorRenderer={(rootErrorRendererProps) => {
        return <>My error</>;
    }}
/>
```

MartynasStrazdas: Removed `renderError` from `rootErrorRenderer`.

.changeset/stupid-otters-listen.md
```diff
@@ -0,0 +1,17 @@
+---
+"@itwin/presentation-hierarchies-react": minor
+---
+
+Exposed `RootErrorRenderer`, which takes `renderRootError` property to render custom root error messages.
```
grigasp, line 5: Looks like this message needs to be updated

MartynasStrazdas: changed

grigasp (approved):

packages/hierarchies-react/src/presentation-hierarchies-react/itwinui/TreeRenderer.tsx
grigasp: > Currently we allow `onNodeClick` and `onNodeKeyDown` to be passed into TreeRenderer which currently does nothing.

After a deeper look, I wouldn't say do nothing. As far as I can tell, they overwrite the callbacks that we create using `useSelectionHandler`. Which, I think, matches the pattern we have with other iTwinUI based components - if that's the case, I don't think we need to do anything about it. Adding @saskliutas for opinion.

PR #941, opened 2025-04-24
Bump `eslint` and `@itwin/eslint-plugin`

saskliutas:
Bumped `eslint` and `@itwin/eslint-plugin` versions.
The new plugin version has `@typescript-eslint/no-non-null-assertion` rule enabled to ban `!` usage in the code. Added ignores where it looked like `nullable` value is handled and fixed some other cases. Disabled this rule for tests.
Also there were quite a few `@typescript-eslint/unbound-method` rule errors. Fixed errors where it was reasonably easy to fix and added ignores in other cases.

### DIFF
--- a/packages/components/src/presentation-components/tree/controlled/UseHierarchyLevelFiltering.ts
+++ b/packages/components/src/presentation-components/tree/controlled/UseHierarchyLevelFiltering.ts
@@ -36,7 +36,7 @@ export function useHierarchyLevelFiltering(props: UseHierarchyLevelFilteringProp
 
   const handleFilterAction = (nodeId: string, info?: PresentationInstanceFilterInfo) => {
     if (ongoingSubscriptions.current.has(nodeId)) {
-      ongoingSubscriptions.current.get(nodeId)?.unsubscribe();
+      ongoingSubscriptions.current.get(nodeId)!.unsubscribe();
       ongoingSubscriptions.current.delete(nodeId);
     }
     const subscription = applyHierarchyLevelFilter(nodeLoader, modelSource, nodeId, () => ongoingSubscriptions.current.delete(nodeId), info);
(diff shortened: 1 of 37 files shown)


packages/components/src/presentation-components/tree/controlled/UseHierarchyLevelFiltering.ts
```diff
@@ -36,7 +36,7 @@ export function useHierarchyLevelFiltering(props: UseHierarchyLevelFilteringProp
 
   const handleFilterAction = (nodeId: string, info?: PresentationInstanceFilterInfo) => {
     if (ongoingSubscriptions.current.has(nodeId)) {
-      ongoingSubscriptions.current.get(nodeId)!.unsubscribe();
+      ongoingSubscriptions.current.get(nodeId)?.unsubscribe();
```
grigasp, line 39: why not `!`? we know there is a subscription for that node

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -93,7 +93,7 @@ class HiliteSetProviderImpl implements HiliteSetProvider {
     let lastEmitTime = performance.now();
     const subject = new Subject<HiliteSet>();
     const subscriptions = (["models", "subCategories", "elements"] as const).map((key) =>
-      observables[key]!.subscribe({
+      observables[key]?.subscribe({
```
grigasp, line 96: I think `!` was fine - we know that `observables` have each of the keys.

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -115,7 +115,7 @@ class HiliteSetProviderImpl implements HiliteSetProvider {
           subject.complete();
         },
         error(err) {
-          subscriptions.forEach((x) => x.unsubscribe());
+          subscriptions.forEach((x) => x?.unsubscribe());
```
grigasp, line 118: this should become unnecessary if we switch the above `?` to `!`

grigasp (commented):
I'm not sure I see value in `@typescript-eslint/no-non-null-assertion` rule. TS already complains if you try to use a nullable value, and requires adding a `!` if you're sure it won't be `undefined`. I don't see how another check on top of that helps in any way,

grigasp (approved):

PR #961, opened 2025-05-12
[Hierarchies-react]: move errors to HierarchyNode as atribute

MartynasStrazdas:


packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -335,48 +336,22 @@ function useTreeInternal({
...
-      };
+    .filter((node): node is TreeModelHierarchyNode => !!node)
+    .map<PresentationHierarchyNode>((node) => {
+      return createPresentationHierarchyNode(node, model);
     });
 }
 
 function createPresentationHierarchyNode(modelNode: TreeModelHierarchyNode, model: TreeModel): PresentationHierarchyNode {
-  let children: Array<PresentationTreeNode> | undefined;
+  let children: Array<PresentationHierarchyNode> | undefined;
```
grigasp, line 354: this is probably going to be a stupid question... but do we need to have `TreeModelHierarchyNode` and `PresentationHierarchyNode`? the two look very similar...

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -78,14 +78,14 @@ export class TreeActions {
...
-                collectTreePartsUntil(this._reset, props.initialRootNode),
+                collectTreePartsUntil(this._reset, props.loadOptions.parent, props.initialRootNode),
                 tap({
                   next: (newModel: TreeModel) => {
                     const childNodes = newModel.parentChildMap.get(props.parentId);
                     const firstChildNode = childNodes?.length ? newModel.idToNode.get(childNodes[0]) : undefined;
                     this.handleLoadedHierarchy(props.parentId, newModel);
                     // only report load duration if no error occurs
-                    if (!(firstChildNode && isTreeModelInfoNode(firstChildNode))) {
+                    if (!firstChildNode) {
```
grigasp, line 88: doesn't look like the comment above makes sense anymore

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -111,6 +111,9 @@ export class TreeActions {
       return;
     }
 
+    if (newModel.rootNode.error) {
+    }
```
grigasp, line 115: needs cleanup?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeModel.ts
```diff
@@ -26,59 +27,52 @@ export interface TreeModelHierarchyNode {
...
   hierarchyLimit?: number | "unbounded";
   instanceFilter?: GenericInstanceFilter;
+  error?: TreeModelError;
 }
 
 /** @internal */
-export interface TreeModelGenericInfoNode {
+export interface TreeModelGenericError {
   id: string;
   parentId: string | undefined;
```
grigasp, line 36: 1. do these error types need to have `parentId`, now that they're assigned to the parent node?
2. is there a reason to have very similar error types, e.g. `TreeModelGenericError` vs `PresentationGenericError`?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -5,13 +5,13 @@
...
 import { GenericInstanceFilter, HierarchyNode, HierarchyProvider, RowsLimitExceededError } from "@itwin/presentation-hierarchies";
-import { isTreeModelHierarchyNode, TreeModelHierarchyNode, TreeModelInfoNode, TreeModelNode, TreeModelRootNode } from "./TreeModel.js";
+import { isTreeModelHierarchyNode, TreeModelError, TreeModelHierarchyNode, TreeModelRootNode } from "./TreeModel.js";
 import { createNodeId } from "./Utils.js";
 
 /** @internal */
 export interface LoadedTreePart {
   parentId: string | undefined;
-  loadedNodes: TreeModelNode[];
+  loadedNodes: (TreeModelHierarchyNode | TreeModelError)[];
```
grigasp, line 14: do we need this array to possibly contain `TreeModelError`? maybe it could return `TreeModelHierarchyNode[]`, and in case of error - just throw it?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -284,12 +286,16 @@ export class TreeActions {
...
-function collectTreePartsUntil(untilNotifier: Observable<void>, rootNode?: TreeModelRootNode) {
+function collectTreePartsUntil(untilNotifier: Observable<void>, parent: TreeModelHierarchyNode | TreeModelRootNode, rootNode?: TreeModelRootNode) {
   return (source: Observable<LoadedTreePart>) =>
     source.pipe(
       reduce<LoadedTreePart, TreeModel>(
         (treeModel, loadedPart) => {
-          addNodesToModel(treeModel, loadedPart);
+          if (!isTreeModelHierarchyNode(loadedPart.loadedNodes[0])) {
+            addErrorToModel(treeModel, loadedPart.loadedNodes[0], parent);
+          } else {
```
grigasp, line 296: as mentioned in the `TreeLoader.ts` comment, I don't like this part - I think we should throw on error instead of returning them in the nodes' array

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -302,23 +308,32 @@ function collectTreePartsUntil(untilNotifier: Observable<void>, rootNode?: TreeM
...
 }
 
-function addNodesToModel(model: TreeModel, hierarchyPart: LoadedTreePart) {
+function addErrorToModel(model: TreeModel, error: TreeModelError, parent: TreeModelHierarchyNode | TreeModelRootNode) {
+  if (parent.id === undefined) {
+    model.rootNode.error = error;
+    return;
+  }
+  model.idToNode.set(parent.id, { ...parent, error });
+  return;
```
grigasp, line 317:
```suggestion
  if (parent.id === undefined) {
    model.rootNode.error = error;
  } else {
    model.idToNode.set(parent.id, { ...parent, error });
  }
```

packages/hierarchies-react/src/presentation-hierarchies-react/TreeNode.ts
```diff
@@ -6,79 +6,64 @@
...
-export type PresentationTreeNode = PresentationHierarchyNode | PresentationInfoNode;
-
-/**
- * An utility function to check if a node is a `PresentationHierarchyNode`.
- * @public
- */
-export function isPresentationHierarchyNode(node: PresentationTreeNode): node is PresentationHierarchyNode {
-  return "children" in node;
-}
+export type PresentationError = PresentationGenericError | PresentationResultSetTooLargeError | PresentationNoFilterMatchesError;
```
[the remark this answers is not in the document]
MartynasStrazdas: My thoughts on naming was to remove `Presentation` prefix from everything `PresentationHierarchyNode` -> `HierarchyNode`. We could then have `PresentationError` -> `HierarchyErrorInfo` or `ErrorInfo`. Don't think we need to add `TreeNode` to the interface as we can document that these errors are atributes of `PresentationHierarchyNode`

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -117,10 +117,10 @@ interface ReloadTreeOptions {
...
-   * Array containing root tree nodes. It is `undefined` on initial render until any nodes are loaded.
-   */
-  rootNodes: PresentationTreeNode[] | undefined;
+  /*   * Array containing root tree nodes. It is `undefined` on initial render until any nodes are loaded. */
+  rootNodes: PresentationHierarchyNode[] | undefined;
+  /** An error object which is defined when root nodes fail to load. Defined as undefined when rootNodes where sucessfully loaded or are in loading process */
+  rootError: PresentationError | undefined;
   /**
    * Specifies whether tree is loading or not. It is set to `true` when initial tree load is in progress
    * or tree is reloading.
```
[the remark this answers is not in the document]
MartynasStrazdas: I tried having them `{ rootNodes: PresentationHierarchyNode[], rootError: undefined, isLoading: false } | { rootNodes: undefined, rootError: PresentationError, isLoading: false } ...`, which ts did not seem to fully be able to handle. But it seems your asking to seperate them further what would be the purpose of that?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeModel.ts
MartynasStrazdas: had some issues merging with the newest `next` version seems the removal of `parentId` has been reverted will remove it. For the differences of `Model` and `Hierarchy` errors i think if we were to remove the model part entirely we could merge them. But having nodes as `Model` nodes and `errors` as hierarchy errors could be weird as we then expose hierarchy parts to the loader.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
MartynasStrazdas: We would need to rewrite some portion of loader to be able to determine when it needs to make an array out of output. More importantly we want to figure out how we want to check if `loadedNodes` are an array every time we use them before the errors are merged inside the ModelNode.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
MartynasStrazdas: When we load nodes we still load modelNodes and errorNodes seperatly. `TheeModelHierarchyNode` also contain some loading information such as `hierarchyLimit` and `instanceFilter`, which would make `PresentationHierarchyNode` overloaded with information it does not need. Having them merged into one I think would add unnecessary complexity.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: Cleaned

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: This part overall seems to be wrong, changed to reflect previous implementation.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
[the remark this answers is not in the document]
grigasp: > what would be the purpose of that?

To make it clear what combinations of them are possible. Currently the API allows all three of them to be set to something and consumers need to read docs to know how to handle them. Ideally, the API itself would explain the possibilities.

In addition, that should also help with TS type checks, e.g.:
```ts
if (x.isLoading) {
  return "loading";
}

// now TS knows that we have either `rootNodes` or `rootError` defined
if (x.rootError) {
  return "error";
}

// now TS knows that we have `rootNodes` - no need to check for `undefined`
return rootNodes.map((n) => n.id).join("\n");
```

packages/hierarchies-react/src/presentation-hierarchies-react/TreeNode.ts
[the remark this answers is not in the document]
grigasp: > My thoughts on naming was to remove `Presentation` prefix from everything `PresentationHierarchyNode` -> `HierarchyNode`.

I don't think renaming it to `HierarchyNode` is a good idea - we already have a concept like this in `@itwin/presentation-hierarchies`, so it would be pretty confusing to have similar-named type in sister package. 

> We could then have `PresentationError` -> `HierarchyErrorInfo` or `ErrorInfo`. Don't think we need to add `TreeNode` to the interface as we can document that these errors are atributes of `PresentationHierarchyNode`.

I'd prefer `ErrorInfo` from these two names.

When choosing `Tree` vs `Hierarchy` prefix, I the distinction of whether we're in the context of data or UI. In case of data - use "hierarchy", in case of a UI component - use "tree". The `@itwin/presentation-hierarchies` package is all about data, so we use "hierarchy" term there. But `@itwin/presentation-hierarchies-react` is about UI components, so we should prefer "tree" and "tree node" here.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
grigasp: To me it looks like the loader could always return `TreeModelHierarchyNode[]`. If an error occurs, then just throw it. Then whoever calls `loadNodes` on the loader (e.g. `TreeActions`) would, instead of checking the type of returned nodes, catch the error and update the parent node.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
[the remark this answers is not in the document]
MartynasStrazdas: But the user would not know if he has error, loading or rootNodes and would need to still check them.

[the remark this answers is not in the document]
grigasp: yes, but it would be clear - both to consumers and to TS type checker - that you can only have one of them defined at a time.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeModel.ts
saskliutas: I don't think we need `TreeModel*Error` types anymore. It made sense when they were part of tree model itself but as they now are part of node you can use `Presentation*Error` as a type for error property on both `TreeModelNode` and `PresetnationNode`

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
saskliutas: What about changing `LoadedTreePart` to 
```
type LoadedTreePart = {
  parentId: string | undefined;
  loadedNodes: TreeModelHierarchyNode[];
} | {
  parentId: string | undefined;
  error: TreeModelError;
}
```

grigasp: that would be okay with me

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -117,10 +117,10 @@ interface ReloadTreeOptions {
...
 /** @public */
 export interface UseTreeResult {
-  /**
-   * Array containing root tree nodes. It is `undefined` on initial render until any nodes are loaded.
-   */
-  rootNodes: PresentationTreeNode[] | undefined;
+  /*   * Array containing root tree nodes. It is `undefined` on initial render until any nodes are loaded. */
+  rootNodes: PresentationHierarchyNode[] | undefined;
+  /** An error object which is defined when root nodes fail to load. Defined as undefined when rootNodes where sucessfully loaded or are in loading process */
+  rootError: PresentationError | undefined;
```
JonasDov, line 123: It sounds a bit strange: `Defined as undefined`, don't you think?
```suggestion
  /*   * Array containing root tree nodes. It is `undefined` on initial render until any nodes are loaded. */
  rootNodes: PresentationHierarchyNode[] | undefined;
  /** An error object which is defined when root nodes fail to load. It is `undefined` when rootNodes where sucessfully loaded or are in loading process */
  rootError: PresentationError | undefined;
```

packages/hierarchies-react/src/presentation-hierarchies-react/TreeNode.ts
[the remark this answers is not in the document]
MartynasStrazdas: Renamed to `ErrorInfo`, would also like to rename `PresentationHierarchyNode` to `TreeNode`, but I thinks it's would fit better in another pr.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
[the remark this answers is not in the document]
MartynasStrazdas: Changed.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeModel.ts
MartynasStrazdas: Removed the type, model and `PresentationHierarchyNode` both now use  `ErrorInfo`

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
MartynasStrazdas: Changed to Saulius suggestion.

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
MartynasStrazdas: Changed

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: Small adjustments were made for this part to meet new loader output format.

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeRenderer.tsx
```diff
@@ -23,18 +23,15 @@ export type TreeNodeRendererProps = ComponentPropsWithoutRef<typeof TreeNodeRend
...
   /** A render function for errors' display component. Defaults to `<TreeErrorRenderer />`. */
   errorRenderer?: (props: TreeErrorRendererProps) => ReactElement;
 }
 
 /** @alpha */
-type TreeRendererProps = Pick<ReturnType<typeof useTree>, "expandNode"> &
-  Partial<Pick<ReturnType<typeof useTree>, "selectNodes" | "isNodeSelected">> &
-  Pick<ReturnType<typeof useTree>, "reloadTree" | "getHierarchyLevelDetails"> &
+type TreeRendererProps = Pick<UseTreeResult, "reloadTree" | "getHierarchyLevelDetails"> &
+  Pick<TreeRenderProps, "expandNode" | "selectNodes" | "isNodeSelected" | "rootNodes"> &
```
grigasp, line 34: Having `TreeRendererProps` & `TreeRenderProps` is not great... Could `useTree` use `TreeRendererProps`, that is defined here, instead?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -78,14 +79,14 @@ export class TreeActions {
...
-                collectTreePartsUntil(this._reset, props.initialRootNode),
+                collectTreePartsUntil(this._reset, props.loadOptions.parent, props.initialRootNode),
                 tap({
                   next: (newModel: TreeModel) => {
                     const childNodes = newModel.parentChildMap.get(props.parentId);
                     const firstChildNode = childNodes?.length ? newModel.idToNode.get(childNodes[0]) : undefined;
                     this.handleLoadedHierarchy(props.parentId, newModel);
                     // only report load duration if no error occurs
-                    if (!(firstChildNode && isTreeModelInfoNode(firstChildNode))) {
+                    if (!(firstChildNode && firstChildNode.error)) {
```
grigasp, line 89: the error's are now on the parent rather than the first child, right?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
```diff
@@ -284,12 +284,16 @@ export class TreeActions {
...
   return (source: Observable<LoadedTreePart>) =>
     source.pipe(
       reduce<LoadedTreePart, TreeModel>(
         (treeModel, loadedPart) => {
-          addNodesToModel(treeModel, loadedPart);
+          if ("loadedNodes" in loadedPart) {
+            addNodesToModel(treeModel, loadedPart.loadedNodes, loadedPart.parentId);
+          } else {
+            addErrorToModel(treeModel, loadedPart.error, parent);
+          }
```
grigasp, line 296: I don't like that we take `parent` as arg, and get `parentId` an attribute of `loadedPart`. `addNodesToModel` takes id from `loadedPart`, but `addErrorToModel` takes `parent` arg.

The two should match, but who knows... what could we do to get rid of one of them?

grigasp: Also, maybe we should consolidate `addNodesToModel` + `addErrorToModel` to `addTreePartToModel`?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -61,56 +67,54 @@ export class TreeLoader implements ITreeLoader {
...
-                  {
-                    id: `${infoNodeIdBase}-no-filter-matches`,
-                    parentId: parent.id,
-                    type: "NoFilterMatches" as const,
-                  },
-                ]
-              : childNodes.map(treeModelNodesFactory),
-        }),
-      ),
+          error: childNodes,
```
grigasp, line 105: Here's what I see:
1. `toArray` gives us either an array of nodes or throws an error.
2. `catchError` gets an error and maps it to `ErrorInfo` object. After the operator we may get either an array of nodes or an `ErrorInfo`.
3. `map` has to do some interesting checks to determine what it got from `catchError`, then returns a `LoadedTreePart`, possibly with error state even if an error wasn't thrown.

Here's what I suggest to simplify this:
1. `toArray` gives us either an array of nodes or throws an error.
2. `map` gets an array, does the filter + nodes count and throws an error if needed, or returns a `LoadedTreePart` with `loadedNodes`.
3. `catchError` gets an error, maps it to `ErrorInfo` object, and returns a `LoadedTreePart` with `error`.

What do you think?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
```diff
@@ -61,56 +67,54 @@ export class TreeLoader implements ITreeLoader {
...
     return this.loadChildren(options).pipe(
-      expand((loadedPart) =>
-        from(loadedPart.loadedNodes.filter((node): node is TreeModelHierarchyNode => isTreeModelHierarchyNode(node) && shouldLoadChildren(node))).pipe(
+      expand((loadedPart) => {
+        return from("loadedNodes" in loadedPart ? loadedPart.loadedNodes.filter((node) => shouldLoadChildren(node)) : []).pipe(
           mergeMap((node) => this.loadChildren({ ...options, parent: node })),
-        ),
-      ),
+        );
+      }),
```
grigasp, line 117: Could we also move the `filter((node) => shouldLoadChildren(node))` part into the `pipe` (use rxjs filter)?


```suggestion
      expand((loadedPart) => 
        "loadedNodes" in loadedPart
          ? from(loadedPart.loadedNodes.filter((node) => shouldLoadChildren(node))).pipe(
              mergeMap((node) => this.loadChildren({ ...options, parent: node })),
            )
          : EMPTY
      ),
```

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
```diff
@@ -116,20 +116,52 @@ interface ReloadTreeOptions {
...
+export type RenderProps =
+  | {
+      /**
+       * An object containing information used to render root error handling UI.
+       * Use in combination with `RootErrorRenderer` or your custom implementation.
+       * Defined when root nodes fail to load.
+       */
+      rootErrorRenderProps: {
+        /** Object containing root error information */
+        rootError: ErrorInfo;
```
grigasp, line 146: maybe just call this `error`? currently there's a bit of a repetition

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeRenderer.tsx
MartynasStrazdas: I don't think that would be beneficial, we would need to define `"expandNode" | "selectNodes" | "isNodeSelected" | "rootNodes"` inside tree renderer but the same problem would occur of what its name should be. 

 we could rename `TreeRenderProps` from `useTree` to `UseTreeRenderProps`?

packages/hierarchies-react/src/presentation-hierarchies-react/UseTree.ts
MartynasStrazdas: Changed

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeLoader.ts
MartynasStrazdas: Added suggestion, also changed filter to rxjs.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: The problem is that the `treeModel` may not contain parent in its `idToNode`, in this case the error would be lost as we need to update that part, therefor we need to get parent from outside. For the `AddNodesToModel` it does not care what the parent is it just needs `parentId` so we can just pass LoadedPart spread out (we need it spread out because the loadedPart definition includes error). The only change here I think could make this cleaner is to have `addNodesToModel` take its own definition of `loadedPart` which does not include error.

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeRenderer.tsx
grigasp: Couldn't we do this?
- Merge all props from `TreeRenderProps (UseTree.ts)` into `TreeRendererOwnProps (TreeRenderer.tsx)`
- Use `TreeRendererProps (TreeRenderer.tsx)` for `UseTreeResult.treeRenderProps (UseTree.ts)`. It will have a bunch of additional optional props, but I don't see why that's a problem.

Looking at the existing `TreeRendererProps` definition:
```ts
type TreeRendererProps = Pick<UseTreeResult, "reloadTree" | "getHierarchyLevelDetails"> &
  TreeRenderProps &
  Pick<TreeErrorRendererProps, "onFilterClick"> &
  Omit<TreeNodeRendererProps, "node" | "aria-level" | "aria-posinset" | "aria-setsize" | "reloadTree" | "selected" | "error"> &
  TreeRendererOwnProps &
  ComponentPropsWithoutRef<typeof LocalizationContextProvider>;
```

Isn't it a bit funny, that we request props that are defined at `UseTreeResult.treeRenderProps`, but then also `UseTreeResult.reloadTree` and `UseTreeResult.getHierarchyLevelDetails`. If they're required by `TreeRenderer`, when why are they not under `UseTreeResult.treeRenderProps`?

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: > The problem is that the `treeModel` may not contain parent in its `idToNode`, in this case the error would be lost as we need to update that part, therefor we need to get parent from outside.

Could `LoadedTreePart` contain `parent` instead of `parentId`?

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeRenderer.tsx
MartynasStrazdas: Moved renderering props to outside, now both `useTree` and `TreeRenderer` (`StrataKitTreeRenderer`) use the same type.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: changed to load parent instead of parentId

packages/hierarchies-react/src/presentation-hierarchies-react/Renderers.ts
```diff
@@ -0,0 +1,98 @@
...
+export type RootErrorRendererProps = {
+  /** Object containing root error information */
+  error: ErrorInfo;
+} & TreeErrorHandlingProps;
+
+/**
+ * Actions used to handle errors.
+ * @alpha
+ */
+interface TreeErrorHandlingProps {
```
grigasp, line 47: Are they really error handling props? I think `getHierarchyLevelDetails` is used for other cases as well

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
grigasp: but now we're going to report load durations even on errors?

grigasp (commented):
Looking at changes in test app, I must say I love the changes to `useTree` return type. I think they make the API easier to use, and even more important - harder to misuse.

packages/hierarchies-react/src/presentation-hierarchies-react/internal/TreeActions.ts
MartynasStrazdas: In this case there would not be any children as the parent was changed, so the if inside would not be hit.

packages/hierarchies-react/src/presentation-hierarchies-react/Renderers.ts
MartynasStrazdas: From the tree's perspective I think it is only being used to provide solutions to errors, as it is not even passed to `NodeRenderer`. Ps. filtering action requires `getHierarchyLevelDetails`, but it is defined by consumer outside the tree.

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -714,6 +743,10 @@ describe("TreeActions", () => {
           children: [],
         },
       ]);
+      // simulate loader return
+      provider.getNodes.callsFake(() => {
```
grigasp, line 747: is this necessary? we're resetting this just below

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -732,7 +765,7 @@ describe("TreeActions", () => {
       newModel = onModelChangedStub.secondCall.args[0];
       expect(getHierarchyNode(newModel, "root-1")).to.not.be.undefined;
       expect(getHierarchyNode(newModel, "child-1")).to.be.undefined;
-      expect(newModel.idToNode.get("root-1-no-filter-matches")).to.not.be.undefined;
+      expect(newModel.idToNode.get("root-1")?.error).to.not.be.undefined;
```
grigasp, line 768: I think we should also check the error type and possibly adjust the test name

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
```diff
@@ -390,6 +398,11 @@ describe("TreeActions", () => {
         },
       ]);
 
+      // simulate loader return
+      provider.getNodes.callsFake(() => {
+        return createAsyncIterator([]);
+      });
```
grigasp, line 404: maybe this could be done in test case setup to avoid repeating it in many tests?

packages/hierarchies-react/src/test/internal/TreeLoader.test.ts
```diff
@@ -327,16 +328,20 @@ describe("TreeLoader", () => {
 });
 
 async function collectNodes(loadObs: Observable<LoadedTreePart>) {
-  const nodes = new Map<string | undefined, TreeModelNode[]>();
-  return new Promise<Map<string | undefined, TreeModelNode[]>>((resolve) => {
+  const nodes = new Map<string | undefined, (TreeModelHierarchyNode | ErrorInfo)[]>();
```
grigasp, line 331: This isn't great. We should test the library and not our test code. Now we have tests like "loads info node when RowsLimitExceededError is thrown", which basically test this `collectNodes` function. I think this function should return either nodes' array, or an error. And each test should be adjusted to have appropriate name.

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
```diff
@@ -290,7 +274,7 @@ describe("useSelectionHandler", () => {
       });
 
       it("skips info nodes when selecting range", async () => {
-        const nodes = [createHierarchyNode("node-1"), createInfoNode("node-2"), createHierarchyNode("node-3")];
+        const nodes = [createHierarchyNode("node-1"), createHierarchyNode("node-3")]; // TODO ADD error: createInfoNode("node-2")
```
grigasp, line 277: needs cleanup?

packages/hierarchies-react/src/test/UseTree.test.tsx
```diff
@@ -63,7 +59,8 @@ describe("useTree", () => {
     hierarchyProvider.getNodes.callsFake(() => createAsyncIterator([]));
     const { result, unmount } = renderHook(useTree, { initialProps });
     await waitFor(() => {
-      expect(result.current.isLoading).to.be.false;
+      const treeRenderProps = getTreeRendererProps(result.current);
+      expect(treeRenderProps).to.not.be.undefined; // TODO
```
grigasp, line 63: needs cleanup

packages/hierarchies-react/src/test/UseTree.test.tsx
```diff
@@ -321,17 +324,23 @@ describe("useTree", () => {
...
+      expect(treeRenderProps).to.not.be.undefined;
+      expect(treeRenderProps?.rootNodes).to.have.lengthOf(1);
+      expect(treeRenderProps?.rootNodes[0].children).to.be.true;
     });
 
     act(() => {
-      result.current.expandNode(createNodeId(rootNodes[0]), true);
+      const treeRenderProps = getTreeRendererProps(result.current);
+      expect(treeRenderProps).to.not.be.undefined;
+      treeRenderProps?.expandNode(createNodeId(rootNodes[0]), true);
```
grigasp, line 336: I think for actions, or maybe even everywhere in tests, we should prefer using `!` instead of `?`. That would also make `expect(treeRenderProps).to.not.be.undefined` checks unnecessary.

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
MartynasStrazdas: Moved to `beforeEach`

packages/hierarchies-react/src/test/UseSelectionHandler.test.tsx
MartynasStrazdas: Removed this test.

packages/hierarchies-react/src/test/UseTree.test.tsx
MartynasStrazdas: removed comment

packages/hierarchies-react/src/test/UseTree.test.tsx
MartynasStrazdas: I think we still don't need to use `!`, in most cases we check for rootNodes checks in this case I removed the check for undefined as the rootNodes check would fail. In other cases the undefined check is there to see if the hook finished loading data.

grigasp: > In other cases the undefined check is there to see if the hook finished loading data.

It makes sense to have the `undefined` check when it's the only check being done. But there's no point to do it when you're testing nested values, e.g.

```ts
const value = getValue();
expect(value).to.not.be.undefined; // this is not necessary
expect(value?.x).to.eq(123);
```

I would also argue that using `!` is better than `?` in tests.
- With `?`, the test will fail either when `value` or when `value.x` are `undefined`. The message will be something like `undefined != 123`, so you won't know which of them was `undefined`.
- With `!`, the test will also fail when either of them are `undefined`. However, the messages will be different: with `value = undefined`, you'll get "attempting to use undefined as object` or something like that, with `value.x = undefined` you'll get `undefined != 123`.

packages/hierarchies-react/src/test/internal/TreeActions.test.ts
grigasp: please also clean up the commented-out code

packages/hierarchies-react/src/test/UseTree.test.tsx
MartynasStrazdas: Changed most `?` to `!` for easier debugging.

.changeset/sixty-spies-jump.md
```diff
@@ -0,0 +1,32 @@
...
+      return (
+        <MyLoadingComponent />
+      );
+    }
+
+    return <StrataKitTreeRenderer {...useTreeResult.treeProps} />;
+}
+```
+
+- Errors are no longer defined as children of `PresentationHierarchyNode` instead are now included as `error` atribute to better reflect UI.
```
grigasp, line 32:
```suggestion
**Breaking changes** to tree state hooks `useTree`, `useUnifiedSelectionTree`, `useIModelTree`, `useIModelUnifiedSelectionTree`:
- All tree rendering props have been moved under `treeRendererProps`. The value can be passed to `<StrataKitTreeRenderer />` component.
- In case an error occurs while loading the root hierarchy level, instead of `treeRendererProps`, the `rootErrorRendererProps` are set, which can be passed to `<StrataKitRootErrorRenderer />` component to render the error state.
- The `isLoading` attribute has been renamed to `isReloading`.
- Errors are no longer defined as children of `PresentationHierarchyNode` and instead are now included as `error` attribute for more fluent API.

When rendering tree state, the recommended order of checks is:
1. If `rootErrorRendererProps` is defined, there was an error - render error state.
2. If `treeRendererProps` is not defined, the component is doing the initial load - render loading state.
3. If `treeRendererProps` is defined, the hierarchy is loaded - render the tree component. 
   
   The `isReloading` attribute may also be set at the same time, indicating that the hierarchy is being reloaded in the background. Consumers may want to render an overlay in this case, or not render the tree at all.

Example:

```ts
function MyComponent () {
    const useTreeResult = useIModelUnifiedSelectionTree({...});

    // handle possible root errors with our provided component or your own custom implementation
    if (treeProps.rootErrorRendererProps) {
      return <StrataKitRootErrorRenderer {...useTreeResult.rootErrorRendererProps} />;
    }

    // handle loading state
    if (!treeProps.treeRendererProps) {
      return <MyLoadingComponent />;
    }

    // handle loaded hierarchy
    return <StrataKitTreeRenderer {...useTreeResult.treeProps} />;
}
```

```

grigasp (approved):

PR #970, opened 2025-05-21
[Hierarchy-react]: expose `TreeRendererProps`

MartynasStrazdas:
In some cases, consumers might want to have some of the required props for `StrataKitTreeRenderer` defined in the same component where `StrataKitTreeRenderer` is returned. With the current implementation that would require some kind of work around to get the ` TreeRendererProps`. This is the case because Pick<ReturnType<useTree>, "TreeRendererProps"> will not return the `TreeRendererProps` as it not always exists. To avoid confusion and to make the consumption easier exposing `TreeRendererProps` would be benefitial.

```ts
function Tree (props: propsType) {
  const useTree = useTree({...props})

  ...

  if (useTree.rootErrorRenderer !== undefined) {
     return <StrataKitRootErrorRenderer {...useTree.rootErrorRenderer} />
  } 

  return <MyTreeRenderer treeRendererProps={useTree.treeRendererProps} /> 
}

function MyTreeRenderer ({treeRendererProps}: {treeRendererProps: TreeRendererProps}) {

   // these are required properties of `StrataKitTreeRenderer` that are not provided from `useTree`
   const onFilterClick = {...}
   const selectionMode = {...}
   const getDecorations = {...}


   return <StrataKitTreeRenderer 
        {...treeRendererProps} 
        onFilterClick={setFilteringOptions}
        selectionMode={"extended"}
        getDecorations={getDecorations}
    />
}

```

### DIFF
--- a/.changeset/fluffy-oranges-raise.md
+++ b/.changeset/fluffy-oranges-raise.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-hierarchies-react": minor
 ---
 
-Exposed type `TreeRendererProps`
+Exposed the type of tree state hooks' result `treeRendererProps` attribute - `TreeRendererProps`. The type can be useful when creating custom tree renderers.
--- a/.changeset/fluffy-oranges-raise.md
+++ b/.changeset/fluffy-oranges-raise.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Exposed type `TreeRendererProps`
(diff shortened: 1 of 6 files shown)


.changeset/fluffy-oranges-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
```
grigasp, line 2:
```suggestion
"@itwin/presentation-hierarchies-react": minor
```

.changeset/fluffy-oranges-raise.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Exposed type `TreeRendererProps`
```
grigasp, line 5:
```suggestion
Exposed the type of tree state hooks' result `treeRendererProps` attribute - `TreeRendererProps`. The type can be useful when creating custom tree renderers.
```

grigasp (approved):

PR #979, opened 2025-06-05
Adjust how `isReloading` is determined

JonasDov:
We can determine if skeleton tree should be loaded if `treeRendererProps?.rootNodes === undefined` or `isReloading === true` (https://github.com/iTwin/contextual-tree-control-react/blob/b080b65804ca66cb18c197c200749c1171e93557/packages/resources-browser-react/src/resources-browser-react/components/ResourcesBrowserTree.tsx#L46). 
There is a problem when using filtering - tree flickers. Right now `isReloading` is set to true only while filtering paths are being retrieved and set.  Before [this](https://github.com/iTwin/presentation/pull/961/files?diff=split&w=0#diff-0f1fbbfa5df647ea9653681a8e33c21da1ede66f6df3c570bf1a65f09aad4aff:~:text=isReloading%3A%20isFiltering%2C) we had additional condition that checked if root nodes were loading. This condition helped with flickering problem, adding it back.

### DIFF
--- a/.changeset/proud-jeans-bow.md
+++ b/.changeset/proud-jeans-bow.md
@@ -2,4 +2,4 @@
 "@itwin/presentation-hierarchies-react": patch
 ---
 
-Adjust how `useUnifiedSelectionTree` determines `isReloading` value. `IsReloading` is now set to `true` if root nodes are being loaded.
+Adjust how `useTree` and other variants of it determines `isReloading` value. `isReloading` is now set to `true` if root nodes are being loaded.
--- a/.changeset/proud-jeans-bow.md
+++ b/.changeset/proud-jeans-bow.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Adjust how `useUnifiedSelectionTree` determines `isReloading` value. `IsReloading` is now set to `true` if root nodes are being loaded.
--- a/packages/hierarchies-react/src/test/UseTree.test.tsx
+++ b/packages/hierarchies-react/src/test/UseTree.test.tsx
@@ -157,10 +157,8 @@ describe("useTree", () => {
     };
 
     const promise = new ResolvablePromise<hierarchiesModule.HierarchyNodeIdentifiersPath[]>();
-    const getFilteredPaths1 = sinon.stub().callsFake(async () => promise);
-
     const { result, rerender } = renderHook(useTree, {
-      initialProps: { getHierarchyProvider: () => customHierarchyProvider, getFilteredPaths: getFilteredPaths1 },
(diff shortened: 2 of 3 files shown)


.changeset/proud-jeans-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Adjust how `useUnifiedSelectionTree` determines `isReloading` value. `IsReloading` is now set to `true` if root nodes are being loaded.
```
saskliutas, line 5: This change updated behavior for all `useTree` hook variants so maybe something like this would be better?

```suggestion
Adjust how `useTree` and other variants of it determines `isReloading` value. `isReloading` is now set to `true` if root nodes are being loaded.
```

saskliutas (approved):

packages/hierarchies-react/src/test/UseTree.test.tsx
```diff
@@ -132,6 +132,66 @@ describe("useTree", () => {
...
+    const rootNode2 = createTestHierarchyNode({ id: "root-2" });
+    const hierarchyChanged = new BeEvent();
+    let getNodesCallCount = 0;
+    const customHierarchyProvider: hierarchiesModule.HierarchyProvider = {
+      async *getNodes() {
+        if (getNodesCallCount < 1) {
+          ++getNodesCallCount;
+          for (const node of [rootNode1, rootNode2]) {
+            yield node;
+          }
```
grigasp, line 146:
```suggestion
          yield rootNode1;
          yield rootNode2;
```

packages/hierarchies-react/src/test/UseTree.test.tsx
```diff
@@ -132,6 +132,66 @@ describe("useTree", () => {
...
+      setHierarchyFilter() {
+        hierarchyChanged.raiseEvent();
+      },
+      async *getNodeInstanceKeys() {},
+      setFormatter() {},
+      hierarchyChanged,
+    };
+
+    const promise = new ResolvablePromise<hierarchiesModule.HierarchyNodeIdentifiersPath[]>();
+    const getFilteredPaths1 = sinon.stub().callsFake(async () => promise);
```
grigasp, line 163: Doesn't look like we need `sinon` for either `getFilteredPaths*` function - we could just do this:
```
const getFilteredPaths1 = (() => promise);
```
And probably we don't even need this variable, so that function could be inlined in the props object for the `render` call.

grigasp (approved):

packages/hierarchies-react/src/test/UseTree.test.tsx
JonasDov: Removed `getFilteredPaths1` and `getFilteredPaths2`

PR #982, opened 2025-06-18
Bump itwinjs-core dependencies to v5.0.0 

JonasDov:


packages/components/src/presentation-components/instance-filter-builder/Utils.ts
```diff
@@ -184,7 +184,8 @@ function numericPropertyValidator({ property, value, operator }: ValidatorContex
   }
 
   function getErrorMessage() {
-    return property.quantityType === undefined
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
+    return property.kindOfQuantityName === undefined || property.quantityType === undefined
```
saskliutas, line 188: Looking at this change second time it looks like initial approach was correct. My bad. 😞

```suggestion
    return property.kindOfQuantityName === undefined && property.quantityType === undefined
```

PR #1004, opened 2025-07-11
Change autoExpand option in HierarchyFilteringPathOptions

JonasDov:
Added option to set `depth` for `HierarchyFilteringPathOptions.autoExpand`
Part of https://github.com/iTwin/viewer-components-react/issues/1319

### DIFF
--- a/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
+++ b/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
@@ -9,12 +9,22 @@ import { GenericNodeKey, GroupingNodeKey, HierarchyNodeKey, InstancesNodeKey } f
 
 /** @public */
 export interface FilterTargetGroupingNodeInfo {
-  /** Key of the grouping node. */
+  /**
+   * Key of the grouping node.
+   * @deprecated in 1.6. This option is no longer needed.
+   */
   key: GroupingNodeKey;
 
   /**
-   * Depth of the grouping node in the hierarchy.
-   * Generally, it can be retrieved from `parentKeys.length`.
+   * Depth up to which nodes in the hierarchy should be expanded.
+   */
+  depth: number;
+}
+
+/** @public */
+export interface FilteringPathAutoExpandOption {
+  /**
+   * Depth up to which nodes in the hierarchy should be expanded.
    */
   depth: number;
 }
@@ -28,7 +38,7 @@ export interface HierarchyFilteringPathOptions {
    * - If it's an instance of `FilterTargetGroupingNodeInfo`, then all nodes up to the grouping node that matches this property,
    * will have `autoExpand` flag.
    */
-  autoExpand?: boolean | FilterTargetGroupingNodeInfo;
+  autoExpand?: boolean | FilterTargetGroupingNodeInfo | FilteringPathAutoExpandOption;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -186,7 +197,7 @@ export function createHierarchyFilteringHelper(
         | {
             pathMatcher: (identifier: HierarchyNodeIdentifier) => boolean;
           },
-    ): Pick<HierarchyNode, "filtering" | "autoExpand"> | undefined => {
+    ): (Pick<HierarchyNode, "autoExpand" | "filtering"> & { filtering?: { autoExpandDepth?: number } }) | undefined => {
```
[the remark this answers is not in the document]
JonasDov: I don't think we can change it to your suggestion, it would be breaking change. HierarchyNode` has this autoExpand type: `autoExpand?: boolean;`. If we change `autoExpand` type be possibly an object, then consumers which did this:
```ts
  const props = filteringHelper.createChildNodeProps({ nodeKey: myNode.key });
  if(props?.autoExpand === true) {
    ...doSomething
  } 
  ```
  would get unexpected errors.

File: packages/hierarchies/src/test/imodel/FilteringHierarchyDefinition.test.ts [hunk 2]
```diff
@@ -339,6 +339,114 @@ describe("FilteringHierarchyDefinition", () => {
       expect(node.autoExpand).to.be.undefined;
     });
 
+    it("doesn't set auto-expand when all filtered children paths autoExpand depth is equal to parent keys length", async () => {
```
saskliutas, line 342: Can we have a couple more tests: 
- check if first node in path has `autoExpand: true` when depth is 2 or more.
- check if correct `autoExpand` value is set when we have 2 paths like:
  - `path: [{ id: 1, class: "Class" }, { id: 2, class: "Class" }]` 
  - `path: [{ id: 1, class: "Class" }, { id: 2, class: "Class" }], options: { autoExpand: { depth: 1 } }`

saskliutas (changes requested):

packages/hierarchies/src/test/imodel/FilteringHierarchyDefinition.test.ts
JonasDov: Added two more tests

packages/hierarchies/src/test/imodel/FilteringHierarchyDefinition.test.ts
```diff
@@ -339,6 +339,178 @@ describe("FilteringHierarchyDefinition", () => {
...
+        [ECSQL_COLUMN_NAME_FilterECInstanceId]: "0x1",
+        [ECSQL_COLUMN_NAME_FilterClassName]: "TestSchema.TestName",
+      };
+      const node = await firstValueFrom(filteringFactory.parseNode(row));
+      const row2 = {
+        [NodeSelectClauseColumnNames.FullClassName]: "",
+        [ECSQL_COLUMN_NAME_FilterECInstanceId]: "0x2",
+        [ECSQL_COLUMN_NAME_FilterClassName]: "TestSchema.TestName",
+      };
+      expect(node.autoExpand).to.be.true;
```
saskliutas, line 494:
```suggestion
      expect(node.autoExpand).to.be.true;

      const row2 = {
        [NodeSelectClauseColumnNames.FullClassName]: "",
        [ECSQL_COLUMN_NAME_FilterECInstanceId]: "0x2",
        [ECSQL_COLUMN_NAME_FilterClassName]: "TestSchema.TestName",
      };
```

saskliutas (approved):

PR #1010, opened 2025-07-16
Fix autoExpand depth not working correctly with grouping nodes

JonasDov:
With `HierarchyFilteringPathOptions.autoExpand` set to `{ depth: number }` some nodes were not getting autoExpand propery set to `true`. 
If each node in hierarchy would be grouped and we had these filtered paths:
```ts
const filterPaths: HierarchyFilteringPath[] = [
  { path: [element, element2, element3], options: { autoExpand: { depth: 2 } } }
]
```
We would expect `element2` to have `autoExpand` set to `true`. But it didn't.

Added `includeGroupingNodes` to `FilteringPathAutoExpandOption` which allows expanding only desired grouping nodes.

### DIFF
--- a/packages/hierarchies/learning/HierarchyFiltering.md
+++ b/packages/hierarchies/learning/HierarchyFiltering.md
@@ -91,7 +91,42 @@ const filteringPath: HierarchyFilteringPath = {
 Additionally, you might not want to add `autoExpand` flag to every node in `HierarchyFilteringPath`. For such cases hierarchies may be expanded up to desired depth, which can be achieved by setting the `autoExpand` property to `{ depth: number }`, where depth represents node's depth in the hierarchy excluding grouping nodes:
 
 <!-- [[include: [Presentation.Hierarchies.HierarchyFiltering.HierarchyFilteringPathImport, Presentation.Hierarchies.HierarchyFiltering.AutoExpandUntilDepthWithoutGrouping.FilteringPath], ts]] -->
+<!-- BEGIN EXTRACTION -->
+
+```ts
+import { HierarchyFilteringPath } from "@itwin/presentation-hierarchies";
+
+const filteringPath: HierarchyFilteringPath = {
+  // Path to the element "C"
+  path: [elementKeys.a, elementKeys.b, elementKeys.c],
+  options: {
+    // Auto-expand the hierarchy up to the specified depth. In this case up to and including element "B"
+    autoExpand: { depth: 2 },
+  },
+};
+```
+
+<!-- END EXTRACTION -->
 
 Also, hierarchies may contain grouping nodes, which don't represent anything by themselves, which means they can't be a filter target. In some cases it may be necessary to auto-expand the hierarchy up to a desired grouping node (and not auto-expand grouping nodes below them), which can be achieved by setting the `autoExpand` property to `{ depth: number, includeGroupingNodes: true }`, where depth represents grouping node depth in the hierarchy:
 
 <!-- [[include: [Presentation.Hierarchies.HierarchyFiltering.HierarchyFilteringPathImport, Presentation.Hierarchies.HierarchyFiltering.AutoExpandUntilDepthWithGrouping.FilteringPath], ts]] -->
+<!-- BEGIN EXTRACTION -->
+
+```ts
+import { HierarchyFilteringPath } from "@itwin/presentation-hierarchies";
+
+// Hierarchy has two grouping nodes under C element: one class grouping and one label grouping node.
+// Get grouping node that groups the "C" element and is the nearest grouping node to it
+const groupingNode = await getSelectedGroupingNode();
+const filteringPath: HierarchyFilteringPath = {
+  // Path to the element "C"
+  path: [elementKeys.a, elementKeys.b, elementKeys.c],
+  options: {
+    // Auto-expand the hierarchy up to the last grouping node. The `depth` attribute equals to the number of parents.
+    autoExpand: { includeGroupingNodes: true, depth: groupingNode.parentKeys.length },
+  },
+};
+```
+
+<!-- END EXTRACTION -->
--- a/packages/hierarchies/learning/HierarchyFiltering.md
+++ b/packages/hierarchies/learning/HierarchyFiltering.md
@@ -88,25 +88,10 @@ const filteringPath: HierarchyFilteringPath = {
 
 <!-- END EXTRACTION -->
 
-Additionally, hierarchies may contain grouping nodes, which don't represent anything by themselves, which means they can't be a filter target. However, in some cases it may be necessary to auto-expand the hierarchy up to a grouping node, which can be achieved by setting the `autoExpand` property to a grouping node's identifier - its key and depth in the hierarchy:
+Additionally, you might not want to add `autoExpand` flag to every node in `HierarchyFilteringPath`. For such cases hierarchies may be expanded up to desired depth, which can be achieved by setting the `autoExpand` property to `{ depth: number }`, where depth represents node's depth in the hierarchy excluding grouping nodes:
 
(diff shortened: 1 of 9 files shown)


packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -42,6 +60,47 @@ export interface HierarchyFilteringPathOptions {
   autoExpand?: boolean | FilterTargetGroupingNodeInfo | FilteringPathAutoExpandOption;
 }
 
+/** @public */
+export namespace HierarchyFilteringPathOptions {
```
saskliutas, line 64: Do we need to expose this to consumers? Mayber `HierarchyFilteringPath.mergeOptions` is enough?

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -42,6 +60,47 @@ export interface HierarchyFilteringPathOptions {
...
+    }
+    if (!rhs || !lhs) {
+      return !!rhs ? rhs : lhs;
+    }
+
+    if (!("key" in lhs) && !lhs.includeGroupingNodes) {
+      if (!("key" in rhs) && !rhs.includeGroupingNodes) {
+        return { depth: lhs.depth > rhs.depth ? lhs.depth : rhs.depth };
+      }
+      return { depth: lhs.depth };
```
saskliutas, line 92: Shouldn't this return `rhs`? As at this point it means `rhs` contains `key` or `includeGroupingNodes` and we lose it.

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -42,6 +60,47 @@ export interface HierarchyFilteringPathOptions {
...
+    }
+    if (!("key" in rhs) && !rhs.includeGroupingNodes) {
+      return { depth: rhs.depth };
```
saskliutas, line 95: Same in here, we reach this point only when `lhs` has `key` or `includeGroupingNodes` and we lose it.

saskliutas (changes requested):

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
JonasDov: There was a comment under `HierarchyFilteringPath.mergeOptions`. It specified how autoExpand property is merged. I wanted to move that explanation here, since more options might be added in the future. I have added a comment to `HierarchyFilteringPath.mergeOptions`: `To know how specific attributes get merged, look at 'HierarchyFilteringPathOptions'.`, Do you think I should remove the comment explaining how `autoExpand` is merged under  `HierarchyFilteringPath.mergeOptions` ?

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
JonasDov: This is the thought process behind this implementation:
In cases where there are two autoExpand options:
- one with `key` / `includeGroupingNodes` and `depth`
- one with only `depth`.
I think we should take `depth` value of the one without `key` / `includeGroupingNodes` set, because this way we are more consistent. Example:
```ts
path = [a, b, c, d]
lhs = { depth: 5, includeGroupingNodes: true };
rhs = { depth: 2 };
```
There are two options for the hierarchy:
1.  `groupingNode1` > `groupingNode2` > `a` > `groupingNode3` > `groupingNode4` > `b` > `c` > `d`;
2. `a` > `groupingNode1` > `b` > `c` > `d`;

If we take the one with the greater depth (lhs): then in scenario 1, these nodes have autoExpand set: `groupingNode1`, `groupingNode2`, `a`, `groupingNode3`, `groupingNode4`; in scenario 2, nodes:  `a`, `groupingNode1`, `b`, `c`, `d` have autoExpand flag set. This means that the options did not change, but result we got is completely different - different non grouping nodes got autoExpanded. 

If we prioritize the one without  `key` / `includeGroupingNodes` set (rhs): then in scenario 1, these nodes will have autoExpand set: `groupingNode1`, `groupingNode2`, `a`, `groupingNode3`, `groupingNode4`, `b`; in scenario 2, nodes: `a`, `groupingNode1`, `b` will have autoExpand flag set. This means that options did not change, but we always see node `b` autoExpanded and non the same non grouping nodes are autoExpanded every time.

Thats why I went with the second approach, in my eyes it is a little bit more consistent.

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
JonasDov: Replied in the comment above.

saskliutas (approved):

packages/hierarchies/learning/HierarchyFiltering.md
```diff
@@ -88,23 +88,43 @@ const filteringPath: HierarchyFilteringPath = {
...
+Also, hierarchies may contain grouping nodes, which don't represent anything by themselves, which means they can't be a filter target. In some cases it may be necessary to auto-expand the hierarchy up to a desired grouping node (and not auto-expand grouping nodes below them), which can be achieved by setting the `autoExpand` property to `{ depth: number, includeGroupingNodes: true }`, where depth represents grouping node depth in the hierarchy:
+
+<!-- [[include: [Presentation.Hierarchies.HierarchyFiltering.HierarchyFilteringPathImport, Presentation.Hierarchies.HierarchyFiltering.AutoExpandUntilDepthWithGrouping.FilteringPath], ts]] -->
+<!-- BEGIN EXTRACTION -->
+
+```ts
+import { HierarchyFilteringPath } from "@itwin/presentation-hierarchies";
+
+// Hierarchy has two grouping nodes under C element: one class grouping and one label grouping node.
+// Get grouping node that groups the "C" element and is the nearest grouping node to it
```
grigasp, line 120: I don't understand this sentence - first, we say that grouping nodes are under "C", but then we say we're getting a grouping node that groups "C". @JonasDov, can you explain?

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -25,8 +25,26 @@ export interface FilterTargetGroupingNodeInfo {
...
+   *
+   * You want to `autoExpand` only `Node1` and `GroupingNode1` in the following hierarchy:
+   * - Node1
+   *   - GroupingNode1
+   *     - GroupingNode2
+   *       - Element1
+   *       - Element2
+   * Then you provide `autoExpand: { depth: 2, includeGroupingNodes: true }`
+   */
+  includeGroupingNodes?: boolean;
```
grigasp, line 47: I feel like this new prop might be very confusing to use, because consumers don't always control or even know how/when grouping nodes are created - how do they know what `depth` to pass when `includeGroupingNodes` is set to `true`?

Taking a step back, I see two separate use cases for auto-expand:
1. Auto-expanding on instance filter. In this case we always target instance-based nodes and we don't care about grouping ones - we just want to expand up to specific instance node. Here, we don't know the path in hierarchy, but we know the instances' path.
2. Auto-expanding to a specific grouping node, when such node is already know before we start the filtering. Here, we know the full path in hierarchy (`parentKeys` on the grouping node).

What do you think about covering those two cases with two separate options: `depthInPath` and `depthInHierarchy`?

JonasDov: > how do they know what `depth` to pass when `includeGroupingNodes` is set to `true`

They should use `parentKeys.length` of the node they want hierarchy to be expanded to. This explanation should probably be added to the comment.

> What do you think about covering those two cases with two separate options: `depthInPath` and `depthInHierarchy`

I agree that `depthInPath` and `depthInHierarchy` should be clearer to the consumer than with `includeGroupingNodes`. Will change to that.

JonasDov: Created a PR that applies your suggestion: https://github.com/iTwin/presentation/pull/1015#issuecomment-3101916686

PR #1023, opened 2025-08-04
[Hierarchies-react]: add inline & menu actions

MartynasStrazdas:


packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/FilterAction.tsx
```diff
@@ -38,6 +47,18 @@ export const FilterAction = memo(function FilterAction({ node, onFilter, getHier
     hierarchyLevelDetails && onFilter?.(hierarchyLevelDetails);
   }, [node, getHierarchyLevelDetails, onFilter]);
 
+  if (!onFilter || !node.isFilterable) {
+    return inline ? (
+      <Tree.ItemAction
```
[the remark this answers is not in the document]
MartynasStrazdas: Changed

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeNodeRenderer.tsx
```diff
@@ -82,23 +101,23 @@ export const StrataKitTreeNodeRenderer: FC<PropsWithRef<TreeNodeRendererProps &
...
-              />,
-            ]
-          : []),
-        ...(getActions ? getActions(node) : []),
-      ],
-      [getActions, node, localizedStrings, reloadTree],
-    );
+    const inlineActions = useMemo(() => {
+      if (node.error !== undefined && node.error.type === "Unknown") {
+        return [
```
saskliutas, line 106: Previous implementation was rendering `Retry` & other actions. Now we render only `Retry` action. Is this intended?

MartynasStrazdas: Yes, from `stratakit` example of how to handle errors, we should only render actions that would solve the error as other actions would not provide any value and could have unexpected behavior because of the error.

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/FilterAction.tsx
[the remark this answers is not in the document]
grigasp: maybe the following would be nicer?

```ts
const disabled = (!onFilter || !node.isFilterable);
if (disabled && !reserveSpace) {
  return undefined;
}

return (
  <Tree.ItemAction
    disabled={disabled}
    ...
  />
);
```

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/RenameAction.tsx
```diff
@@ -23,7 +37,7 @@ export const RenameAction = memo(function RenameAction() {
   }, [setIsRenaming]);
 
   if (!context?.onLabelChanged) {
-    return undefined;
+    return reserveSpace ? <Tree.ItemAction label={rename} onClick={handleClick} visible={false} disabled icon={renameSvg} /> : undefined;
```
grigasp, line 40: avoid duplicating `<Tree.ItemAction />` rendering

MartynasStrazdas: I'd rather keep the reserve action and the actual action separate. Even if it has some code duplication, I think it keeps it cleaner in a way that the handling of undefined component is in one return instead of multiple as well as no need to extend current visible and disabled logic for the actual action.

grigasp: thanks, I can live with that

PR #1024, opened 2025-08-04
[Hierarchies] Clarify `depthInPath` docs

JonasDov:
Clarified `depthInPath` documentation. Previously it wasn't clear if `depthInPath` included the node at position or not. Now it does not. Added tests for this as well.

### DIFF
--- a/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
+++ b/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
@@ -61,9 +61,11 @@ export interface FilteringPathAutoExpandOption {
 /** @public */
 export interface FilteringPathAutoExpandDepthInPath {
   /**
-   * Depth up to which nodes in the filtering path should be expanded.
+   * Depth that tells which nodes in the filtering path should be expanded.
    *
    * Use when you want to expand up to specific instance node and don't care about grouping nodes.
+   *
+   * **NOTE**: All nodes that are up to `depthInPath` will be expanded. Node at the `depthInPath` position won't be expanded.
    */
   depthInPath: number;
 }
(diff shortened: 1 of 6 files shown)


packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -86,6 +88,8 @@ export interface FilteringPathAutoExpandDepthInHierarchy {
    * Then you provide `autoExpand: { depthInHierarchy: 2 }`
    *
    * To get the correct depth use `HierarchyNode.parentKeys.length`.
+   *
+   * **NOTE**: All nodes that are up to and including `depthInHierarchy` will be expanded. Node at the `depthInHierarchy` position will be expanded.
```
grigasp, line 92: I think the second sentence is repeating the first one.

```suggestion
   * **NOTE**: All nodes that are up to and including `depthInHierarchy` will be expanded.
```

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -103,7 +110,12 @@ export class FilteringHierarchyDefinition implements RxjsHierarchyDefinition {
 
       for (const path of child.filtering.filteredChildrenIdentifierPaths) {
         if ("path" in path && path.options?.autoExpand) {
-          return true;
+          if (path.options.autoExpand === true) {
```
grigasp, line 113: why not move the checks for `true` and `object` into `shouldBeExpanded` function? it looks like we're repeating these checks at lines 92-99 as well.

JonasDov: Moved

JonasDov: Moved `shouldExpandGroupingNode` outside of `FilteringHierarchyDefinition` class. Moved `shouldBeExpanded` function outside of `shouldExpandGroupingNode`.

grigasp (approved):

PR #1039, opened 2025-08-12
Components: Use `IModelApp.FormatsProvider`

grigasp:
Part of https://github.com/iTwin/viewer-components-react/issues/1340

### DIFF
--- a/apps/full-stack-tests/src/testing/RulesetTesting.test.ts
+++ b/apps/full-stack-tests/src/testing/RulesetTesting.test.ts
@@ -8,7 +8,8 @@ import { expect } from "chai";
 import ChaiJestSnapshot from "chai-jest-snapshot";
 import path from "path";
 import { Id64 } from "@itwin/core-bentley";
-import { IModelConnection } from "@itwin/core-frontend";
+import { IModelApp, IModelConnection } from "@itwin/core-frontend";
+import { SchemaFormatsProvider } from "@itwin/ecschema-metadata";
 import { ChildNodeSpecificationTypes, ContentSpecificationTypes, RelationshipDirection, Ruleset, RuleTypes } from "@itwin/presentation-common";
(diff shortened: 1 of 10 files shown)


apps/full-stack-tests/src/testing/RulesetTesting.test.ts
```diff
@@ -209,6 +210,9 @@ describe("RulesetTesting", () => {
 
     // set up for testing iModel presentation data
     iModel = TestIModelConnection.openFile(iModelPath);
+
+    // set up schema-based formatter
+    IModelApp.formatsProvider = new SchemaFormatsProvider(iModel.schemaContext, "metric");
```
grigasp, line 215: @hl662, the `generates correct content` revealed an interesting issue - we're not formatting properties using schema-based formats without this piece of code.

- Our formatter gets/creates a formats provider like this:
   ```ts
   const formatsProvider = this._formatsProvider ?? new SchemaFormatsProvider(this._schemaContext, unitSystem);
   ```

- Presentation frontend, when creating a formatter for formatting properties, passes it `IModelApp.formatsProvider`, so we always use that rather than `SchemaFormatsProvider` that's created by default.

- If the application doesn't set `SchemaFormatsProvider`, then we're using the default `QuantityFormatsProvider`, which is useless for formatting schema-based properties.

How do you think this should be solved?

hl662: >If the application doesn't set SchemaFormatsProvider, then we're using the default QuantityFormatsProvider, which is useless for formatting schema-based properties.

I'd say have presentation fallback to instantiate it's own SchemaFormatsProvider, and/or, similar to IModelConnection, itwin-core could expose a method that checks that a formatsProvider is an instance of SchemaFormatsProvider. This would lead to a path towards enforcing any application that uses presentation API to register a SchemaFormatsProvider (and SchemaUnitProvider I guess).

Thoughts?

PR #1050, opened 2025-08-20
[components]: Quantity editor should react to format settings change

saskliutas:
Updated quantity properties editor to react and rerender value when format settings changes. Tested with https://github.com/iTwin/viewer-components-react/pull/1404


.changeset/tired-words-stop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": minor
```
grigasp, line 2: why not patch?

PR #1072, opened 2025-09-26
[components] Fix table loading

saskliutas:
Table was not loading when grouping node was selected using `Presentation.selection` because we were using only instance keys and instance nodes from KeySet.


packages/components/src/test/table/UsePresentationTable.test.tsx
```diff
@@ -184,6 +184,29 @@ describe("usePresentationTableWithUnifiedSelection", () => {
       sinon.stub(Presentation, "selection").get(() => selectionManager);
     });
 
+    it("loads data when grouping node is selected", async () => {
+      const grouingKey = createTestECClassGroupingNodeKey();
+      const keys = new KeySet([grouingKey]);
```
grigasp, line 189:
```suggestion
      const groupingKey = createTestECClassGroupingNodeKey();
      const keys = new KeySet([groupingKey]);
```

grigasp (approved):

PR #1083, opened 2025-10-15
Add unique guid to restart tokens

JonasDov:
creating because of this: https://github.com/iTwin/itwinjs-backlog/issues/1677

Also, added restart tokens to queries that previously did not have them.

### DIFF
--- a/packages/unified-selection/src/unified-selection/SelectionScope.ts
+++ b/packages/unified-selection/src/unified-selection/SelectionScope.ts
@@ -50,7 +50,7 @@ export interface ComputeSelectionProps {
   elementIds: Id64Arg;
   /** Selection scope to compute selection with. */
   scope: SelectionScope;
-  /** Optional id of the component that consumes this function. It can be any unique string that is */
+  /** Optional unique id of the component that consumes this function. It can be any string, as long as it is not shared across different components. */
   componentId?: string;
 }
 


packages/unified-selection/src/unified-selection/SelectionScope.ts
```diff
@@ -113,7 +113,11 @@ async function* computeElementSelection(
     FROM AncestorElements
     WHERE ${recurseUntilRoot ? "" : "Depth = 0 OR"} ParentId IS NULL
   `;
-  yield* executeQuery(queryExecutor, { ctes, ecsql, bindings });
+  yield* executeQuery({
+    queryExecutor,
+    query: { ctes, ecsql, bindings },
+    config: { restartToken: `SelectionScope/compute-element-selection-query/${Guid.createValue()}` },
+  });
```
grigasp: - I think `computeSelection` should take an optional prop for the id (create one if not passed), and pass it down to these local functions.
- I'd create the `restartToken` in format: `computeSelection/{componentId}/element`

packages/unified-selection/src/unified-selection/HiliteSetProvider.ts
```diff
@@ -237,7 +245,13 @@ class HiliteSetProviderImpl implements HiliteSetProvider {
             SELECT ECInstanceId FROM Models
           `,
         ].join(" UNION ");
-        return from(executeQuery(this._imodelAccess, { ctes, ecsql, bindings }));
+        return from(
+          executeQuery({
+            queryExecutor: this._imodelAccess,
+            query: { ctes, ecsql, bindings },
+            config: { restartToken: `HiliteSetProvider/hilited-models-query/${Guid.createValue()}` },
```
grigasp: - let's create the id as a member of `HiliteSetProviderImpl` and use it here.
- I'd change the format of `restartToken` to `HiliteSetProvider/{componentId}/models`

packages/components/src/presentation-components/instance-filter-builder/ECMetadataProvider.ts
```diff
@@ -71,7 +71,10 @@ export class ECMetadataProvider {
         WHERE classDef.ECInstanceId = :id
       `;
       classInfo = await this.createECClassInfo(
-        this._queryReaderFactory(classQuery, QueryBinder.from({ id }), { rowFormat: QueryRowFormat.UseJsPropertyNames }),
+        this._queryReaderFactory(classQuery, QueryBinder.from({ id }), {
+          rowFormat: QueryRowFormat.UseJsPropertyNames,
+          restartToken: `ECMetadataProvider/class-by-id-query/${Guid.createValue()}`,
+        }),
```
grigasp: - let's create the id as a member of `ECMetadataProvider` and use it here.
- I'd change the format of `restartToken` to `ECMetadataProvider/{componentId}/class-by-id`

packages/hierarchies/src/hierarchies/imodel/TreeNodesReader.ts
```diff
@@ -24,6 +24,7 @@ export function readNodes(props: ReadNodesProps): Observable<SourceInstanceHiera
   const parser: RxjsNodeParser = props?.parser ?? ((row) => of(defaultNodesParser(row)));
   const config: Parameters<LimitingECSqlQueryExecutor["createQueryReader"]>[1] = {
     rowFormat: "ECSqlPropertyNames",
+    restartToken: `TreeNodesReader/read-nodes-query/${Guid.createValue()}`,
```
grigasp:
```suggestion
    restartToken: `readNodes/${Guid.createValue()}`,
```

.changeset/red-socks-camp.md
```diff
@@ -0,0 +1,8 @@
+---
+"@itwin/unified-selection": patch
```
grigasp: One of my comments suggests adding an extra optional prop to `computeSelection` - that means it should be a minor bump. Also, since it's an addition, it would be great to provide an extra explanation what the prop means and what its value should be set to. So I think we need a separate change file for the `@itwin/unified-selection` package.

packages/components/src/presentation-components/instance-filter-builder/ECMetadataProvider.ts
JonasDov: Changed

packages/unified-selection/src/unified-selection/SelectionScope.ts
```diff
@@ -50,6 +50,8 @@ export interface ComputeSelectionProps {
   elementIds: Id64Arg;
   /** Selection scope to compute selection with. */
   scope: SelectionScope;
+  /** Optional id of the component that consumes this function. It can be any unique string that is */
```
grigasp: comment needs a review

PR #1115, opened 2025-11-21
[hierarchies-react] Add context menu to tree

saskliutas:
Added ability to render tree node actions in context menu.
Refactored tree actions to use unified base components that allows to share them between different contexts: inline, context menu and actions dropdown.
Actions passed to tree renderer should be rendering newly introduced `TreeActionBase` component.

### DIFF
--- a/scripts/validateMarkdowns.js
+++ b/scripts/validateMarkdowns.js
@@ -23,7 +23,7 @@ Promise.all(
   filePaths.map(
     (filePath) =>
       new Promise((resolve) => {
-        // TODO: fix false negatives for `npmjs.com` links.
+        // TODO: fix false negatives for `npmjs.com` links. https://github.com/iTwin/presentation/issues/1116
         const p = spawn("pnpm", ["markdown-link-check", "-a", "200,403", filePath], { shell: true });
 
         let myStdout = "";


packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeAction.tsx
```diff
@@ -0,0 +1,62 @@
...
+   * - "default" - standard action displayed in the dropdown menu accessible through the ellipsis button on the node.
+   * - "inline" - actions displayed directly on the tree node.
+   * - "context-menu" - actions displayed in the context menu opened by right-clicking the tree node.
+   */
+  variant?: "default" | "inline" | "context-menu";
+  /**
+   * Indicates that the actions is not applicable for current tree node and should be hidden.
+   * When set to true, the action will not be rendered at all.
+   */
+  hide?: boolean;
```
grigasp:
```suggestion
   * Indicates that the action is not applicable for current tree node and should be hidden.
   * When set to true, the action will not be rendered at all.
```

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeAction.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ComponentProps, memo } from "react";
+import { DropdownMenu, Tree } from "@stratakit/structures";
+
+/**
+ * Attributes used to decide how to render tree actions. They allow to differentiate action style based on the context
+ * where that action is used.
```
grigasp: Let's call this `StrataKitTreeActionProps` to be consistent with other type names we have.

PR #1120, opened 2025-11-25
[Hierarchies]: Add autoExpand on filter targets

JonasDov:
Changed `createHierarchyFilteringHelper` to set `autoExpand` flag on filter targets when `depthInPath` is greater than path length.


JonasDov:
Closing in favor of https://github.com/iTwin/presentation/pull/1129

Issue #844, opened 2025-01-20
`hierarchies`: Support fully merged iModel hierarchy providers


grigasp:
At the moment we support merged hierarchy providers, but iModel-based providers won't merge with each other, e.g. see the following:

```
// provider 1 returns:
- A
  - A1
- B

// provider 2 returns:
- A
  - A2
- C

// merged hierarchy:
- A
  - A1
- A
  - A2
- B
- C
```

Instead, we'd like to see this:
```
- A
  - A1
  - A2
- B
- C
```

This kind of merged hierarchy could only be built under assumption that given iModels are related and their schema items match if their names match.

Issues:
- Hierarchy level filtering. 

  At the moment, the filter is passed in the form of `GenericInstanceFilter`, which has no notion of an iModel, but it uses schema items. Even if the iModels are related (e.g. different versions of the same iModel), there could be new schema items added to newer versions of the iModel, so we may receive schema items that are supported in one of the merged iModels, but not the other. What do we do with that?

  Several ideas:
    - In the merged hierarchy provider, accept multiple filters for each iModel separately. It would then know which iModel to forward each filter to. This wouldn't be compatibly with the `HierarchyProvider` interface.
    - In `GenericInstanceFilter` API, add a way to tell which iModel(s) filter rules apply to.
    - When converting `GenericInstanceFilter` to ECSQL, check each given filter rule against the target iModel and ❓skip it❓ if the schema item used in that rule doesn't exist.

PR #1128, opened 2025-12-02
Hierarchies: iModel hierarchies' merging (WIP)

grigasp:
This is a step towards https://github.com/iTwin/presentation/issues/844.

- Fix `equals` & `compare` functions to do a case-insensitive comparison and ignore different variants of schema-class separator.
- Introduce `createMergedIModelHierarchyProvider` function that creates a single hierarchy from multiple versions of the same iModel.
  - The API is very similar to `createIModelHierarchyProvider`, but takes multiple iModels.
  - Having this functionality implemented at this level, instead of `mergeProviders`, makes things a lot less complicated:
    - we don't have to worry about merged hierarchies being based on different hierarchy definitions,
    - we can merge nodes before grouping, which solves the problems related to conditional grouping,

TODO (in separate PRs):
- Ensure hierarchy level filtering works on merged hierarchy, especially when newer versions have schema changes.
- Ensure apps can create a hierarchy level filtering dialog for such merged hierarchies. This will likely land in `@itwin/presentation-components`.
- Docs & learning.

### DIFF
--- a/apps/full-stack-tests/src/hierarchies/HierarchiesMerging.test.ts
+++ b/apps/full-stack-tests/src/hierarchies/HierarchiesMerging.test.ts
@@ -543,12 +543,12 @@ describe("Hierarchies", () => {
           provider: createMergedHierarchyProvider({
             imodels: [
               {
-                // has Y node which is not grouped by class
+                // has one Y node with "y" label
                 ecdb: dbs.base.ecdb,
                 key: "base",
               },
               {
-                // has Z node
+                // has two Y nodes with "y" label - one from `base` and one that exists only in `changeset1`
                 ecdb: dbs.changeset1.ecdb,
                 key: "changeset1",
               },


apps/full-stack-tests/src/hierarchies/HierarchiesMerging.test.ts
```diff
@@ -0,0 +1,1865 @@
...
+                      node: (node) => {
+                        expect(node.label).to.eq(`y`);
+                        expect((node.key as InstancesNodeKey).instanceKeys).to.deep.equalInAnyOrder([{ ...dbs.base.y, imodelKey: "base" }]);
+                      },
+                    },
+                  ],
+                },
+                {
+                  node: (node) => {
+                    expect(node.label).to.eq(`z`);
```
saskliutas: Why there is no `z` label grouping node?

apps/full-stack-tests/src/hierarchies/HierarchiesMerging.test.ts
```diff
@@ -0,0 +1,1865 @@
...
+        await validateHierarchy({
+          provider: createMergedHierarchyProvider({
+            imodels: [
+              {
+                // has Y node which is not grouped by class
+                ecdb: dbs.base.ecdb,
+                key: "base",
+              },
+              {
+                // has Z node
```
saskliutas: Looks like comment is outdated

packages/hierarchies/src/test/HierarchyProvider.test.ts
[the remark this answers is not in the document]
grigasp: Kind of... Without this, `c8` thinks that `HierarchyProvider.ts` is not covered (the source file has only types).

apps/full-stack-tests/src/hierarchies/HierarchiesMerging.test.ts
```diff
@@ -0,0 +1,1865 @@
...
+  const changeset2 = setupChangeset2
+    ? await cloneECDb(changeset1.ecdbPath, `${mochaContext.test!.fullTitle()}-changeset2`, async (ecdb) => setupChangeset2(ecdb, changeset1))
+    : undefined;
+  return {
+    base,
+    changeset1,
+    changeset2,
+    [Symbol.dispose]() {
+      base.ecdb[Symbol.dispose]();
+      changeset1.ecdb[Symbol.dispose]();
```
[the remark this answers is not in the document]
grigasp: yes, fixed

apps/full-stack-tests/run-forever.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+while npm run test:dev; do :; done
```
JonasDov: Why is this file needed?

PR #1129, opened 2025-12-04
[Hierarchies]: Add option to expand filter targets

JonasDov:
Added `autoExpandFilterTarget` option to `HiearchyFilteringPathOptions`

### DIFF
--- a/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
+++ b/packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
@@ -133,9 +133,6 @@ export namespace HierarchyFilteringPath {
 
   /**
    * Merges two given `HierarchyFilteringPathOptions` objects.
-   * - if both inputs are `undefined`, `undefined` is returned,
-   * - else if one of the inputs is `undefined`, the other one is returned.
-   * - else, merge each option individually.
    *
    * For the `reveal` attribute, the merge chooses to `reveal` as deep as the deepest input:
    * - if any one of the inputs is `true`, return `true`,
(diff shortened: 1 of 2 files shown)


packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -415,11 +435,12 @@ class MatchingFilteringPathsReducer {
...
         : undefined),
       ...(parentKeys &&
-      shouldRevealNode({
+      (shouldRevealNode({
         reveal: this._revealOption,
         nodePositionInHierarchy: parentKeys.length,
         nodePositionInPath: parentKeys.filter((key) => !HierarchyNodeKey.isGrouping(key)).length,
-      })
+      }) ||
+        !!this._filterTargetOptions?.autoExpandFilterTarget)
```
grigasp: shouldn't we also check if the node is filter target?

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
```diff
@@ -65,17 +65,20 @@ export class FilteringHierarchyDefinition implements RxjsHierarchyDefinition {
     return (node) => {
       return (this._source.postProcessNode ? this._source.postProcessNode(node) : of(node)).pipe(
         map((processedNode) => {
+          const parentKeysWithoutGroupingNodesLength = processedNode.parentKeys.filter((key) => !HierarchyNodeKey.isGrouping(key)).length;
+          const parentKeysLength = processedNode.parentKeys.length;
           const shouldReveal = ProcessedHierarchyNode.isGroupingNode(processedNode)
```
grigasp: `shouldReveal` -> `shouldAutoExpand`?

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
```diff
@@ -95,6 +103,16 @@ namespace HierarchyFilteringPathOptions {
...
   }
+
+  export function mergeAutoExpandOption(
+    lhs: HierarchyFilteringPathOptions["autoExpandFilterTarget"],
+    rhs: HierarchyFilteringPathOptions["autoExpandFilterTarget"],
+  ): HierarchyFilteringPathOptions["autoExpandFilterTarget"] {
+    if (lhs === true || rhs === true) {
+      return true;
+    }
+    return lhs !== undefined ? lhs : rhs;
```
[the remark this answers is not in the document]
JonasDov: Changed to return undefined when both are falsy

packages/hierarchies/src/hierarchies/HierarchyFiltering.ts
JonasDov: If filterTargetOptions exist, it already means that node is a filter target, no need for this check

packages/hierarchies/src/hierarchies/imodel/FilteringHierarchyDefinition.ts
JonasDov: Changed

packages/hierarchies/src/test/imodel/FilteringHierarchyDefinition.test.ts
```diff
@@ -379,6 +379,23 @@ describe("FilteringHierarchyDefinition", () => {
       expect(result).to.eq(sourceFactoryNode);
     });
 
+    it("sets autoExpand when autoExpand is set to true and node is filter target", async () => {
```
grigasp:
```suggestion
    it("sets autoExpand on node when path's autoExpand is set to true and node is filter target", async () => {
```

grigasp (approved):

PR #1183, opened 2026-01-21
`hierarchies-react`: Tree errors cleanup

grigasp:
Part of https://github.com/iTwin/presentation/issues/932.

Based on [this comment](https://github.com/iTwin/presentation/issues/932#issuecomment-2829940754):

> `TreeErrorRenderer` / `TreeErrorItemProps`:
> - Is there are point for `reloadTree` prop function to take `state: "reset"` option?
> - I think `onFilterClick` should be called `filterHierarchyLevel` to be consistent with `reloadTree` (we're not calling it `onResetClick`).
> - I think `scrollToElement` should be renamed to `scrollToNode` and possibly take `PresentationHierarchyNode` instead of `ErrorNode`.
> - It seems `ErrorNode.expandTo` is similar in function to `scrollToElement` and we should try to make them consistent by moving `expandTo` to `TreeErrorItemProps` and renaming it to `expandToNode`. On the other hand, is `expandTo` needed at all? As far as I can tell, it's only used internally.

### DIFF
--- a/packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeErrorRenderer.tsx
+++ b/packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/TreeErrorRenderer.tsx
@@ -20,7 +20,7 @@ interface TreeErrorRendererOwnProps {
    * as it's used for creating a unique accessible label for the error region.
    */
   treeLabel: string;
-  /** List of error nodes to render errors for be displayed */
+  /** List of error nodes to render errors for. */
   errorNodes: ReturnType<typeof useErrorNodes>;
   /** Callback to render custom error messages. Component should be wrapped in `ErrorRegion.Item` from `@itwin/itwinui-react` package. */
   renderError?: (props: ErrorItemRendererProps) => ReactElement;


packages/hierarchies-react/api/presentation-hierarchies-react.api.md
```diff
@@ -54,23 +54,20 @@ export type ErrorInfo = GenericErrorInfo | ResultSetTooLargeErrorInfo | NoFilter
 interface ErrorItem {
     // (undocumented)
     errorNode: TreeNode & Pick<Required<TreeNode>, "error">;
-    // (undocumented)
-    expandTo: (expandNode: (nodeId: string) => void) => void;
```
grigasp: @saskliutas, do you see value in keeping the `ErrorItem` type after this removal? I'm thinking about changing `useErrorList` to return `Array<TreeNode & Pick<Required<TreeNode>, "error">>` instead 🤔

saskliutas: No need to keep it. It was only useful with `expandTo` function.

saskliutas (approved):

PR #1193, opened 2026-01-26
[Hierarchies-react]: add react compiler

MartynasStrazdas:
Performance review:

using React provider Profiler wrapper and wrapping `<StatelessModelsTreePanel>`

Initial load:
before
react-profiler-no-compiler
after
react-profiler-with-compiler
After rerunning multiple times the initial load does not seem to be impacted by react compiler

Expanding node:
before:
expand-node-no-compiler

after
expand-node-compiler
Expanding a node does not seem to be affected either.

Rename action:
before:
rename-no-compiler
after
rename-action-compiler
Rename action seems to have less updates, minor improvement

It's hard to determine if react compiler has provided much of an improvement as there was no significant decrease in rendering time or rerender amounts.

### DIFF
--- a/packages/hierarchies-react/eslint.config.cjs
+++ b/packages/hierarchies-react/eslint.config.cjs
@@ -18,7 +18,7 @@ module.exports = [
     },
     rules: {
       ...iTwinPlugin.configs.uiConfig.rules,
-      ...reactHooks.configs.flat.recommended.rules,
+      ...reactHooksPlugin.configs.flat.recommended.rules,
     },
   },
   {


packages/hierarchies-react/package.json
```diff
@@ -28,24 +28,23 @@
...
     ".": {
-      "import": "./lib/esm/presentation-hierarchies-react.js"
+      "import": "./lib/esm/presentation-hierarchies-react.mjs"
     },
     "./package.json": "./package.json"
   },
   "scripts": {
-    "build": "npm run -s build:esm",
-    "build:esm": "npm run -s copy:esm && tsc -p tsconfig.esm.json",
-    "copy:esm": "cpx \"./src/**/*.{css,json}\" ./lib/esm",
```
saskliutas: Why don't we need to copy css files? Does `tsdown` takes care of that?

packages/hierarchies-react/tsdown.config.ts
```diff
@@ -0,0 +1,32 @@
...
+import { babel } from "@rollup/plugin-babel";
+
+export default defineConfig({
+  entry: "./src/presentation-hierarchies-react.ts",
+  outDir: "./lib/esm",
+  format: ["esm"],
+  unbundle: true,
+  plugins: [
+    babel({
+      babelHelpers: "bundled",
```
saskliutas: What does `bundled` helper do? It looks weird as we also have `unbundle: true`. 😆

.changeset/ripe-tires-peel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-hierarchies-react": patch
+---
+
+Added react compiler.
```
grigasp:
```suggestion
Process package output with [React Compiler](https://react.dev/reference/react-compiler/compiling-libraries).
```

packages/hierarchies-react/package.json
MartynasStrazdas: I believe i tried this before and double checked now it does not seem to work, the import statement is still removed and the copied out css files do not have the original name.

packages/hierarchies-react/.nycrc
```diff
@@ -4,7 +4,7 @@
   "functions": 100,
   "branches": 100,
   "lines": 100,
-  "include": ["src/presentation-hierarchies-react/**/*", "lib/esm/presentation-hierarchies-react/**/*"],
+  "include": ["src/presentation-hierarchies-react/**/*", "lib/test/presentation-hierarchies-react/**/*"],
```
grigasp: is this intentional? why are we including tests into coverage?

packages/hierarchies-react/api/presentation-hierarchies-react.api.md
```diff
@@ -328,7 +346,7 @@ interface TreeRendererOwnProps {
 export type TreeRendererProps = {
     rootNodes: TreeNode[];
     expandNode: (nodeId: string, isExpanded: boolean) => void;
-    selectNodes: (nodeIds: Array<string>, changeType: SelectionChangeType) => void;
+    selectNodes: (nodeIds: Array<string>, changeType: SelectionChangeType) => void; /** Determines whether a given node is selected. */
```
grigasp: why did this comment appear there?

packages/hierarchies-react/.mocharc.json
```diff
@@ -1,8 +1,8 @@
 {
-  "require": ["ignore-styles", "./lib/esm/test/setup.js"],
+  "require": ["ignore-styles", "./lib/test/test/setup.js"],
   "checkLeaks": true,
   "global": ["IS_REACT_ACT_ENVIRONMENT"],
   "timeout": 60000,
-  "spec": ["./lib/esm/test/**/*.test.js"],
+  "spec": ["./lib/test/**/*.test.js"],
   "exclude": ["./lib/test/coverage/**/*"]
 }
```
grigasp: The ["Testing Strategy" docs section](https://react.dev/reference/react-compiler/compiling-libraries#testing-strategy) says:

> Test your library both with and without compilation to ensure compatibility. Run your existing test suite against the compiled code, and also create a separate test configuration that bypasses the compiler. This helps catch any issues that might arise from the compilation process and ensures your library works correctly in all scenarios.

I'm not sure we really need to always test against non-compiled version of the library, but we definitely want to test against the compiled one, since that is what we're delivering to our consumers. Ideally (and in the scope of this PR - optionally), we also want an easy way to switch to testing against the non-compiled version, so we can easily see if any issue / test failure is due to the compilation process or not.

However, what I'm seeing is that we're currently always testing against the non-compiled version. Need to fix that.

MartynasStrazdas: Tests now use react compiler generated code, this causes the cover function to fail, removing coverage until the react compiler issue is resolved or a solution is found, more => https://github.com/facebook/react/issues/32950

packages/hierarchies-react/src/presentation-hierarchies-react/stratakit/LocalizationContext.tsx
```diff
@@ -173,6 +173,7 @@ export function LocalizationContextProvider({ localizedStrings, children }: Prop
   const [state, setState] = useState({ localizedStrings: { ...defaultLocalizedStrings, ...localizedStrings } });
 
   useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect
```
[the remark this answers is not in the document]
MartynasStrazdas: changed

saskliutas (approved):

PR #1213, opened 2026-02-04
[Editors]: adjustments for quantity property placeholder & editor key control fixes

MartynasStrazdas:
closes https://github.com/iTwin/itwinjs-backlog/issues/1591

New placeholder:
image
Imperial (feet look a bit wierd when displaying `'`):
image

Editors enter behavior when editors are always displayed.

https://github.com/user-attachments/assets/c4eb2fb1-df4e-4221-b57e-24b75b3307cf

Editors cancel behaviour when editors are always displayed on escape, needs changes in `VirtualizedPropertyGridWithDataProvider`.

### DIFF
--- a/apps/full-stack-tests/package.json
+++ b/apps/full-stack-tests/package.json
@@ -63,7 +63,7 @@
     "cross-env": "catalog:build-tools",
     "eslint": "catalog:build-tools",
     "eslint-plugin-react": "catalog:build-tools",
-    "fast-xml-parser": "^5.3.7",
+    "fast-xml-parser": "^5.3.8",
     "global-jsdom": "catalog:test-tools",
     "i18next-http-backend": "^3.0.2",
     "ignore-styles": "catalog:test-tools",
--- a/apps/test-app/frontend/package.json
+++ b/apps/test-app/frontend/package.json
@@ -49,7 +49,7 @@
     "classnames": "catalog:react",
     "eslint": "catalog:build-tools",
     "eslint-plugin-react": "catalog:build-tools",
-    "fast-xml-parser": "^5.3.7",
+    "fast-xml-parser": "^5.3.8",
     "immer": "^10.2.0",
     "react": "catalog:react",
     "react-dom": "catalog:react",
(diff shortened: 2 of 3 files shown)


MartynasStrazdas:
> > @fmikai, please have a look at images and recording in issue description and tell us what you think.
> 
> @grigasp sorry I missed this comment. Regarding the recording showing the behaviour. Is it possible to keep the placeholders (units) while the user is typing?
> 
> For the imperial inches symbol ' .Can we use "ft" instead so we keep the abbreviation format consistent with other placeholders?
> 
> 544983510-f770c2dc-e79d-404f-a9cf-4b474be5c06f


https://github.com/user-attachments/assets/331d68eb-9533-4a7f-a3de-82b1246ae042


https://github.com/user-attachments/assets/e1c7d3c6-57a3-4e73-9c41-ce75bb4f20ca

@fmikai tried out how displaying unit could work when there is no value. For selection if the user settings are saying that single unit is used we select the value until unit. For the case of multi unit or when the user inputs a different unit we just select everything. What are your thoughts on this?

packages/components/src/presentation-components/properties/inputs/QuantityPropertyEditorInput.tsx
```diff
@@ -92,8 +102,17 @@ const QuantityPropertyValueInput = forwardRef<PropertyEditorAttributes, Quantity
         }}
         onFocus={() => {
           setEditing(true);
-          inputRef.current?.setSelectionRange(0, 9999);
+          requestAnimationFrame(() => {
+            if (quantityValue.isSingleUnit && quantityValue.highPrecisionFormattedValue.endsWith(inputProps.placeholder)) {
+              const valueBeforeUnit = quantityValue.highPrecisionFormattedValue.slice(0, -inputProps.placeholder.length);
+              // Check if the value before the unit consists of digits, separators, whitespace or minus sign - if so, select only that part, otherwise select the entire input value
+              inputRef.current?.setSelectionRange(0, /^[\d.,\s/-]+$/.test(valueBeforeUnit) ? valueBeforeUnit.length : 9999);
```
saskliutas: Why is this regex needed? What about `1.5e+6 m`?

PR #1293, opened 2026-04-09
Update formatting config to match `next` branch

saskliutas:
Update formatter config to match `next` branch
Update eslint script to include `"` to make sure all files are linted when running script.


grigasp (approved):
Looked over config changes and formatting changes in some random source files. Looks good.

PR #1300, opened 2026-04-13
[shared] Use more unique alias for label select clause

### LINKED ISSUE #1297: ECSQL select clause helpers should use more unique aliases
ECSQL select clause factories should use more unique aliases as with current ones it's quite easy to have a collision that would cause ambiguity error. Example: 
```ts
const hierarchyDefinition = createPredicateBasedHierarchyDefinition({
      classHierarchyInspector: imodelAccess,
      hierarchy: {
        rootNodes: async () => [
          {
            fullClassName: "BisCore.GeometricModel3d",
            query: {
              ecsql: `
                SELECT
                  ${await nodesQueryFactory.createSelectClause({
                    ecClassId: { selector: "e.ECClassId" },
                    ecInstanceId: { selector: "e.ECInstanceId" },
                    nodeLabel: {
                      selector: await labelsQueryFactory.createSelectClause({
                        classAlias: "e",
                        className: "BisCore.GeometricModel3d",
                      }),
                    },
                  })}
                FROM BisCore.GeometricModel3d e
              `,
            },
          },
        ],
        childNodes: [],
      },
    });
```

Fails with `BE_SQLITE_ERROR: Expression 'e.ECInstanceId' in ECSQL statement is ambiguous.`

saskliutas:
Aliases used in label select clauses produces by presentation-shared were very generic `c`, `e` and it was very easy to encounter collision when using Element or Category classes (especially when generating hierarchies with AI). Added prefix to those aliases to avoid these collisions.

Closes https://github.com/iTwin/presentation/issues/1297

### DIFF
--- a/apps/full-stack-tests/src/hierarchies/NodesQueryClauseFactory.test.ts
+++ b/apps/full-stack-tests/src/hierarchies/NodesQueryClauseFactory.test.ts
@@ -76,7 +76,7 @@ describe("NodesQueryClauseFactory", () => {
     });
   });
 
-  it("creates correct labels when outer query uses alias 'c' (same as default label factory's internal alias)", async () => {
+  it("creates correct labels when outer query uses common alias 'c'", async () => {
     const { imodel } = await buildTestIModel(async (builder) => {
       insertPhysicalModelWithPartition({ builder, codeValue: "Test model" });
     });
--- a/apps/full-stack-tests/src/hierarchies/NodesQueryClauseFactory.test.ts
+++ b/apps/full-stack-tests/src/hierarchies/NodesQueryClauseFactory.test.ts
@@ -1,0 +1,127 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { insertPhysicalModelWithPartition } from "presentation-test-utilities";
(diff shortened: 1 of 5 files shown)


apps/full-stack-tests/src/hierarchies/NodesQueryClauseFactory.test.ts
```diff
@@ -0,0 +1,127 @@
...
+import {
+  createBisInstanceLabelSelectClauseFactory,
+  createDefaultInstanceLabelSelectClauseFactory,
+} from "@itwin/presentation-shared";
+import { initialize, terminate } from "../IntegrationTests.js";
+import { buildTestIModel } from "../TestIModelSetup.js";
+import { NodeValidators, validateHierarchy } from "./HierarchyValidation.js";
+import { createIModelAccess, createProvider } from "./Utils.js";
+
+describe("NodesQueryClauseFactory", () => {
```
grigasp, line 21: is it worth going through the hassle of `NodesQueryClauseFactory` when you only need to test labels factory? maybe just use something like this:

```ts
const ecsql = `
  SELECT ${await labelsQueryFactory.createSelectClause({
    classAlias: "e",
    className: "BisCore.GeometricModel3d",
  })} AS label
  FROM BisCore.GeometricModel3d e
`;
const reader = imodel.createQueryReader(ecsql);
// ... take the first row's label and validate it
```

saskliutas: Simplified tests

grigasp (approved):

Issue #1306, opened 2026-04-17
`hierarchies`: Make `createNodesQueryClauseFactory` internal


grigasp:
With https://github.com/iTwin/presentation/pull/1307 we already provide the factory to consumers through `HierarchyDefinition.defineHierarchyLevel` and there's no reason for them to call it themselves, unless they use a custom instance labels select clause factory. The plan:
- add an optional instance labels select clause factory creator prop to `IModelHierarchyProviderProps`,
- use the given creator to create a custom factory instead of the default one in hierarchy provider,
- use the same labels factory when creating the nodes select clause factory.

That should further simplify defining iModel hierarchies and allow us to remove `createNodesQueryClauseFactory` from public API.

PR #1307, opened 2026-04-17
Shared: iModel instance label select clause factory

### LINKED ISSUE #805: `shared`: Provide an instance label select clause factory that handles `InstanceLabelOverride` rules in iModels
iModels may have [supplemental rulesets](https://www.itwinjs.org/presentation/advanced/rulesetsupplementation/) with [InstanceLabelOverride](https://www.itwinjs.org/presentation/customization/instancelabeloverride/) inserted in them. The legacy Presentation library finds those rules and takes them into account when computing instance labels. 

We should provide an implementation of [IInstanceLabelSelectClauseFactory](https://github.com/iTwin/presentation/tree/master/packages/shared#instance-labels) that finds and handles those embedded rules.

grigasp:
Closes https://github.com/iTwin/presentation/issues/805.

Introduced the label select clause factory, but it has to a query to get the embedded presentation rules, meaning we want to cache the factory per iModel. That made it very inconvenient to use, so I made a change in hierarchy provider to create this labels factory for consumers and feed it into hierarchy definition. I think this greatly simplifies hierarchy definition implementations.

Going forward, I'm also considering to make `createNodesQueryClauseFactory` internal - I don't see why anyone would want to use a custom select clause factory: https://github.com/iTwin/presentation/issues/1306.


packages/shared/src/shared/instance-label-factory-impls/IModelInstanceLabelSelectClauseFactory.ts
```diff
@@ -0,0 +1,452 @@
...
+  return rule.ruleType === "InstanceLabelOverride";
+}
+
+function normalizeInstanceLabelOverrideRule(rule: RuleBase): InstanceLabelOverride | undefined {
+  if (!isInstanceLabelOverrideRule(rule)) {
+    return undefined;
+  }
+
+  if (rule.values.length === 0) {
+    return undefined;
```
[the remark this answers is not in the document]
grigasp: We don't expect malformed JSONs there. If by any chance we find them, we want to throw.

apps/full-stack-tests/src/shared/InstanceLabelSelectClauseFactory.test.ts
```diff
@@ -0,0 +1,221 @@
...
+          },
+        ],
+      };
+      await new RulesetEmbedder({ imodel }).insertRuleset(ruleset);
+      return { target: el };
+    });
+    expect(await selectInstanceLabel({ imodel: imodelConnection, instanceKey: keys.target })).toBe("HighWins");
+  });
+
+  it("applies `Property` spec with `propertySource` — label from parent's property", async () => {
```
JonasDov: I think this dash is not the regular one, maybe we should not use such dashes?

```suggestion
  it("applies `Property` spec with `propertySource` - label from parent's property", async () => {
```

JonasDov: Same for other places

packages/shared/src/shared/instance-label-factory-impls/IModelInstanceLabelSelectClauseFactory.ts
```diff
@@ -0,0 +1,452 @@
...
+        return "NULL";
+      }
+
+      const lastStep = steps[steps.length - 1];
+      let targetClassName: EC.FullClassName;
+      if (lastStep.targetClass) {
+        targetClassName = `${lastStep.targetClass.schemaName}.${lastStep.targetClass.className}`;
+      } else {
+        const relName: EC.FullClassName = `${lastStep.relationship.schemaName}.${lastStep.relationship.className}`;
+        const relClass = (await getClass(ctx.schemaProvider, relName)) as EC.RelationshipClass;
```
JonasDov: Is this cast needed?

packages/shared/src/shared/instance-label-factory-impls/IModelInstanceLabelSelectClauseFactory.ts
```diff
@@ -0,0 +1,452 @@
...
+  const compiledParts: Array<{ selector: string; isRequired: boolean }> = await Promise.all(
+    spec.parts.map(async (part) => ({
+      selector: await compileValueSpec(part.spec, ctx),
+      isRequired: !!part.isRequired,
+    })),
+  );
+
+  const selectors: TypedValueSelectClauseProps[] = [];
+  compiledParts.forEach(({ selector }, index) => {
+    if (index > 0 && separator) {
```
JonasDov, line 337: Is the separator check needed? It seems that its always truthy

grigasp: empty string will be falsy

packages/shared/src/shared/instance-label-factory-impls/IModelInstanceLabelSelectClauseFactory.ts
grigasp: replaced with a proper check

apps/full-stack-tests/src/shared/InstanceLabelSelectClauseFactory.test.ts
grigasp: replaced with a hypen

JonasDov (approved):

PR #1312, opened 2026-04-24
[Components]: new merged value placeholder for editors

### LINKED ISSUE #1189: Components: Change the way property editors represent merged values
> [@fmikai](https://github.com/fmikai) there is another issue with displaying placeholders, ux input would be greatly appreciated: Currently, when displaying multiple values for a property in read-only mode, we show **varies** text. The issue arises when we have editors always displayed — in this case, we do not display any value, which can be misleading for users, because there are two different values for the property instead of no value displayed.
> 
> Example: Node 1 has length set to 5m. Node 2 has length set to 3m. When both nodes are selected:
> 
> * Read-only mode: Property grid displays **varies**
> * Always show editors mode: Property grid displays nothing (only a greyed-out placeholder)
> 
> @MartynasStrazdas In the property grid, whenever we have a multiple selection with different values, we display two dashes, like --.
> 
> Could we do something like this:
> - Disabled field in read only and active filed in show-editor. Both using the two dashes: --
> Image
>  

 _Originally posted by @fmikai in [#1591](https://github.com/iTwin/itwinjs-backlog/issues/1591#issuecomment-3784886310)_

MartynasStrazdas:
Closes https://github.com/iTwin/presentation/issues/1189
Preview:
image
image


packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
```diff
@@ -67,7 +67,7 @@ export function useQuantityValueInput({ initialRawValue, schemaContext, koqName
       const defaultValue =
         prev.quantityValue.rawValue !== undefined
           ? newDefaultFormatter.applyFormatting(prev.quantityValue.rawValue)
-          : newHighPrecisionFormatter.unitConversions[0].label;
+          : `-- ${newHighPrecisionFormatter.unitConversions[0].label}`;
```
saskliutas, line 70: Why this didn't cause changes in tests? Do we not check if correct value is rendered or there is no test for this?

packages/components/src/presentation-components/common/ContentDataProvider.ts
```diff
@@ -455,15 +455,6 @@ export class ContentDataProvider implements IContentDataProvider {
...
 
-      // we always get formatted content from presentation manager - ensure
-      // we set `_isContentFormatted = true` when we finish getting content, to
-      // avoid formatting it again unnecessarily
-      using _ = {
-        [Symbol.dispose]: () => {
-          this._isContentFormatted = true;
-        },
-      };
-
```
saskliutas, line 466: Why is was removed?

packages/components/src/presentation-components/common/ContentDataProvider.ts
```diff
@@ -589,9 +580,18 @@ class ContentFormatter {
...
+    mergedFieldNames?: string[],
   ) {
     for (const field of fields) {
       const value = values[field.name];
+      // for merged quantity fields, format display value as "-- unit"
+      if (value === undefined && mergedFieldNames?.includes(field.name) && isFieldWithKoq(field)) {
+        const unitLabel = await this._propertyValueFormatter.getUnitLabel(field, this._unitSystem);
+        if (unitLabel) {
+          displayValues[field.name] = `-- ${unitLabel}`;
+        }
```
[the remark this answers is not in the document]
MartynasStrazdas: unit test added

packages/components/src/presentation-components/properties/inputs/NumericPropertyInput.tsx
```diff
@@ -22,6 +22,11 @@ export const NumericPropertyInput = forwardRef<PropertyEditorAttributes, Numeric
   const property: WithConstraints<PropertyDescription> = propertyRecord.property;
 
   const [inputValue, setInputValue] = useState<string>(() => getInputTargetFromPropertyRecord(propertyRecord) ?? "");
+  const [isEditing, setEditing] = useState(false);
+  const displayValue = useMemo(
+    () => (!isEditing && propertyRecord.isMerged && inputValue === "" ? "--" : inputValue),
+    [isEditing, propertyRecord.isMerged, inputValue],
+  );
```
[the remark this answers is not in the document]
MartynasStrazdas: added a change

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
MartynasStrazdas: There were no tests that covered merged properties for the editors. Added some.

packages/components/src/presentation-components/common/ContentDataProvider.ts
MartynasStrazdas: Reverted, previously when testing with appUI changes this seemed to make formatting not trigger, but it seems it was an issue with the build.

pnpm-lock.yaml
saskliutas, line 1: Why lock file changed without any dep version changes?

MartynasStrazdas: Not sure why that one dependency had '@types/node': 22.19.17, every other dependency has dep on node type 24 (except for electron which also has it on 22). Our types/node in catalog is also on 24.12.2. 
I have manually reverted this just in case back to 22.

saskliutas (approved):

PR #834, opened 2025-01-10
Add property constraints

JonasDov:
This is a successor of:
itwinjs-core: https://github.com/iTwin/itwinjs-core/pull/7534
imodel-native: https://github.com/iTwin/imodel-native/pull/963

Afterwards changes will need to be made in appui for text editors

PR #1323, opened 2026-04-30
[Components]: apply constraints for quantity editor

### LINKED ISSUE #779: Support ECProperty value range
ECProperty supports adding constraints to values. 
https://www.itwinjs.org/bis/ec/ec-property/#additional-optional-attributes

These are not enforced by the ECDb.
They should be enforced by the inputs in the property grid.

To query those, you can use the [meta.ECPropertyDef](https://www.itwinjs.org/learning/ecdbmeta.ecschema/#ecpropertydef) table

JonasDov:
Part of https://github.com/iTwin/presentation/issues/779.

Previous changes made in https://github.com/iTwin/presentation/pull/834 applied only for numeric editors. Adjusted quantity editor to also support this. 

To complete https://github.com/iTwin/presentation/issues/779, changes in https://github.com/iTwin/appui/pull/1544 should be merged and then consumed in https://github.com/iTwin/viewer-components-react/tree/master/packages/itwin/property-grid


packages/components/src/presentation-components/properties/inputs/QuantityPropertyEditorInput.tsx
```diff
@@ -72,16 +84,30 @@ const QuantityPropertyValueInput = forwardRef<PropertyEditorAttributes, Quantity
...
+
+      onCommit({
+        propertyRecord,
+        newValue: {
+          valueFormat: PropertyValueFormat.Primitive,
+          value: valueToCommit.rawValue,
+          displayValue: valueToCommit.defaultFormattedValue,
+          roundingError: valueToCommit.roundingError,
+        },
+      });
```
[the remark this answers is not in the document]
JonasDov: added

packages/components/src/presentation-components/properties/inputs/NumericPropertyInput.tsx
```diff
@@ -165,32 +165,14 @@ export const NumericInput = forwardRef<PropertyEditorAttributes, NumericInputPro
...
-  min: number | undefined;
-  max: number | undefined;
-} {
-  if ("minimumValue" in constraints) {
-    return { min: constraints.minimumValue, max: constraints.maximumValue };
-  }
-
-  return { min: undefined, max: undefined };
+  const constrainedNumber = applyNumericConstraints({ value: inputAsNumber, min, max });
+  return constrainedNumber !== inputAsNumber ? constrainedNumber.toString() : inputAsString;
```
grigasp: why not just `return applyNumericConstraints({ value: inputAsNumber, min, max }).toString();`?

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
```diff
@@ -114,7 +114,26 @@ export function useQuantityValueInput({ initialRawValue, schemaContext, koqName
     );
   };
 
-  return { quantityValue, inputProps: { onChange, placeholder, disabled: !highPrecisionFormatter || !parser } };
+  return {
+    quantityValue,
+    inputProps: { onChange, placeholder, disabled: !highPrecisionFormatter || !parser },
+    setNewValue: (rawValue: number): QuantityValue => {
```
saskliutas: Why do we need this? Isn't it enough to validate/clamp in `QuantityPropertyValueInput` before calling `onCommit`? After that I would expect input to rerender/refresh and reset state. We shouldn't need to do that explicitly

packages/components/src/presentation-components/properties/inputs/NumericPropertyInput.tsx
JonasDov: Changed

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
JonasDov: ~`UseQuantityValueInput` hook has the formatters, so some sort of callback should be present in order to properly format the value that might get clamped. Changed to `applyConstraints`. Please have a look~

Changed the implementation, now `rawValue` will always have constraints applied, while display value won't. Please have a look

packages/components/src/test/properties/inputs/QuantityPropertyEditorInput.test.tsx
```diff
@@ -185,6 +201,19 @@ describe("<QuantityPropertyEditorInput />", () => {
     await waitFor(() => expect(input.selectionEnd).toBe(8));
   });
 
+  it("does not throw when onCommit is not provided and input is blurred", async () => {
```
saskliutas: This test does not look related to changes, why is it needed?

packages/components/src/presentation-components/properties/inputs/UseQuantityValueInput.tsx
```diff
@@ -88,33 +95,33 @@ export function useQuantityValueInput({ initialRawValue, schemaContext, koqName
...
-  const { highPrecisionFormatter, parser, defaultFormatter } = useFormatterAndParser(
-    koqName,
-    schemaContext,
-    onFormatterLoad.current,
-  );
+  const useFormatterAndParserResult = useFormatterAndParser(koqName, schemaContext, onFormatterLoad.current);
 
   const onChange: ChangeEventHandler<HTMLInputElement> = (e) => {
-    assert(parser !== undefined); // input should be disabled if parser is `undefined`
+    assert(useFormatterAndParserResult.parser !== undefined); // input should be disabled if parser is `undefined`
```
saskliutas: Maybe change return type of `useFormatterAndParser` to `{ highPrecisionFormatter: FormatterSpec; parser: ParserSpec; defaultFormatter: FormatterSpec } | undefined` so it would be enought to check `useFormatterAndParserResult !== undefined`

JonasDov: changed

packages/components/src/test/properties/inputs/QuantityPropertyEditorInput.test.tsx
JonasDov: I had adjusted the implemenation of `onBlur`, due to that adjustment coverage was failing. This test was just for that. Reverted `onBlur` changes, and removed this test.

saskliutas (approved):

Issue #817, opened 2024-12-18
Investigate possibility to create content descriptor on the frontend


grigasp:

grigasp:
## Initial results

### TS vs Non-TS Comparison

| Model | Non-TS Avg | TS Avg | **Speedup** | Notes         |
| ----- | ---------- | ------ | ----------- | ------------- |
| L-1   | 174.43     | 13.55  | **12.9x**   | Biggest win   |
| XL-3  | 126.55     | 24.15  | **5.2x**    |               |
| XL-2  | 101.40     | 19.61  | **5.2x**    |               |
| M-10  | 0.84       | 0.20   | **4.2x**    |               |
| L-3   | 49.45      | 11.99  | **4.1x**    |               |
| L-4   | 27.11      | 6.90   | **3.9x**    |               |
| XL-1  | 84.06      | 22.31  | **3.8x**    |               |
| L-7   | 12.50      | 3.28   | **3.8x**    |               |
| M-8   | 1.76       | 0.49   | **3.6x**    |               |
| M-7   | 2.13       | 0.61   | **3.5x**    |               |
| M-4   | 3.79       | 1.09   | **3.5x**    |               |
| XL-5  | 81.36      | 23.76  | **3.4x**    |               |
| M-6   | 2.40       | 0.70   | **3.4x**    |               |
| XL-7  | 4.50       | 1.44   | **3.1x**    |               |
| XL-4  | 48.45      | 16.41  | **3.0x**    |               |
| L-9   | 4.70       | 1.62   | **2.9x**    |               |
| S-1   | 0.87       | 0.31   | **2.8x**    |               |
| M-5   | 2.93       | 1.05   | **2.8x**    |               |
| L-2   | 51.91      | 19.51  | **2.7x**    |               |
| L-6   | 12.37      | 4.76   | **2.6x**    |               |
| M-1   | 6.45       | 2.59   | **2.5x**    |               |
| L-8   | 5.40       | 2.69   | **2.0x**    |               |
| M-12  | 0.47       | 0.24   | **2.0x**    |               |
| L-5   | 20.10      | 10.62  | **1.9x**    |               |
| L-10  | 1.70       | 0.88   | **1.9x**    |               |
| M-11  | 0.52       | 0.28   | **1.9x**    |               |
| M-13  | 0.34       | 0.20   | **1.7x**    |               |
| M-14  | 0.21       | 0.13   | **1.6x**    |               |
[...]

PR #1343, opened 2026-05-12
`content`: Initial proposal version

grigasp:
Follow-up of https://github.com/iTwin/presentation/issues/817. Created as a PR so we can easily discuss on specific points. Especially interested in hearing thoughts on some of the open questions.

### DIFF
--- a/packages/content/ts-content-rewrite-proposal.md
+++ b/packages/content/ts-content-rewrite-proposal.md
@@ -28,13 +28,16 @@ The distinction between "class only" and "class + instance IDs" matters for sour
 
 ### Relationship path
 
-This concept is already defined as `RelationshipPath` in the `@itwin/presentation-shared` package (see `RelationshipPathStep` in `packages/shared/src/shared/Metadata.ts`). Each step specifies source class, target class, relationship class, and whether the direction is reversed.
+This concept is already defined as `RelationshipPath` in the `@itwin/presentation-shared` package (see `RelationshipPathStep` in `packages/shared/src/shared/Metadata.ts`). In the current codebase, each step specifies the source class, target class, relationship class, and whether the direction is reversed.
 
-A sequence of steps describing how to navigate from one class to another through EC relationships. Each step specifies:
+A sequence of steps describing how to navigate from one class to another through EC relationships. In the current shared type, each step specifies:
 
 - The relationship class to traverse.
 - The direction (forward = source→target, backward = target→source).
 - The source and target classes at that step.
+
+For this rewrite, we may want to extend the step shape with optional per-step filters, such as:
+
 - An optional **instance filter** on the relationship class (to restrict which relationship instances are traversed).
 - An optional **instance filter** on the target class (to restrict which target instances are reached).
 
--- a/packages/content/ts-content-rewrite-proposal.md
+++ b/packages/content/ts-content-rewrite-proposal.md
@@ -93,7 +93,8 @@ Describes a single data column in the content result. Key attributes:
 - **Relationship path** — which relationship path this field comes from, and which class along that path the field belongs to. Empty path means the field belongs to the target class directly. The path tells the query builder how to JOIN to the field's class; the class identifies which table the column lives in.
 - **Category** — logical grouping for UI display.
 - **Flags** — read-only, hidden (still queried but not displayed in UI), etc.
-- **Renderer / editor hints** — optional overrides for how to display or edit the value.
+
+Note: Renderer and editor selection is a **view-model concern**, not part of the field data model. The pipeline field describes *what* the data is (type, identity, category); consumers decide *how* to render or edit it based on their own configuration (type-based defaults, class + property mappings, explicit overrides in UI component props, etc.).
 
 Field kinds:
 
@@ -387,8 +388,6 @@ interface PropertyOverrides {
   label?: string;
   categoryId?: string; // references a CategoryDefinition.id
   readOnly?: boolean;
-  renderer?: string;
-  editor?: string;
 }
 ```
 
--- a/packages/content/ts-content-rewrite-proposal.md
+++ b/packages/content/ts-content-rewrite-proposal.md
@@ -760,12 +760,7 @@ The pipeline exposes an **async iterator** — consumers `for await` over conten
 
 7. ~~**1:many related content loading strategy:**~~ Moved to Implementation Notes — this is an internal implementation concern, not an API design question.
 
-8. **Distinct values: raw vs. formatted:**
-   `getDistinctValues` runs `SELECT DISTINCT` on raw values, but consumers (e.g., table filter dropdowns) want distinct _display_ values. Multiple raw values may format to the same display string (e.g., different timestamps → same date, different precisions → same rounded number). Options:
-   - **A) Return raw, consumer deduplicates after formatting:** Simple for us, more work to consumers.
-   - **B) Accept a formatter function as input:** The pipeline fetches raw values, applies the consumer-provided formatter, then deduplicates. Formatting stays a consumer concern, but the pipeline handles deduplication.
-
-   This interacts with the "formatting is a frontend concern" decision — need to reconcile.
+8. ~~**Distinct values: raw vs. formatted:**~~ **Resolved.** `getDistinctValues` returns raw values only (consistent with "formatting is a frontend concern"). A reduce-like consumer utility accepts a formatter function and collapses the raw values into a `Map<formattedValue, rawValue[]>` — grouping raw values that format to the same display string. This keeps deduplication logic out of the pipeline while giving consumers a one-liner for the common "filter dropdown" use case.
 
 9. ~~**SQLite/ECSQL query limits:**~~ Moved to Implementation Notes — this is an internal implementation concern, not an API design question.
 
--- a/packages/content/ts-content-rewrite-proposal.md
+++ b/packages/content/ts-content-rewrite-proposal.md
@@ -105,7 +105,7 @@ Note: The current system's "related content field" (a field that contains child 
 
 **Content values (`ContentValues`)** — the raw data for one row of the content result:
 
-- **Primary keys** — which instance(s) this row represents.
+- **Primary key** — which instance this row represents.
 - **Values** — raw values keyed by field identity. All fields (property, SQL calculated, and external) are populated by the pipeline.
 
 This is a plain data bag — serializable, no behavior, no reference to the descriptor.
@@ -114,7 +114,7 @@ This is a plain data bag — serializable, no behavior, no reference to the desc
 
 - `getValue(field: Field): unknown` — retrieve a value by field reference.
 - `getFieldsByGroup(path): Array<{ field, value }>` — iterate related field groups with field+value pairs.
-- Access to `primaryKeys` and the underlying `descriptor`.
+- Access to `primaryKey` and the underlying `descriptor`.
 
 The pipeline's async iterator yields `ContentItem` instances. Consumers work with `ContentItem`; `ContentValues` is an internal/serialization-level concept (used by external fields providers' `resolve` function, export utilities, etc.).
 
@@ -311,20 +311,21 @@ Then pipeline calls external fields provider's resolve(items):
   - Fetches sensor data from IoT service
   - Attaches "iot.currentFlow" and "iot.lastMaintenance" values to items
 
-Output: ContentItem {
-  primaryKeys: [{ class: "ProcessPhysical:Pump", id: "0x3a" }],
-  values: {
-    "Pump.Name":              "Main Circulation Pump",
-    "Pump.FlowRate":          12.5,
-    "flowRate_gpm":           198131.25,
-    "PumpType.Manufacturer":  "Siemens",
-    "PumpType.Model":         "CR 95-3",
-    "OperatingParametersAspect.MaxPressure": 25.0,
-    "OperatingParametersAspect.MaxTemp":     180.0,
-    "iot.currentFlow":        187.3,
-    "iot.lastMaintenance":    "2026-04-22T08:00:00Z",
-  }
+Output: ContentValues {
+  primaryKey:               { className: "ProcessPhysical:Pump", id: "0x3a" },
+  "Pump.Name":              "Main Circulation Pump",
+  "Pump.FlowRate":          12.5,
+  "flowRate_gpm":           198131.25,
+  "PumpType.Manufacturer":  "Siemens",
+  "PumpType.Model":         "CR 95-3",
+  "OperatingParametersAspect.MaxPressure": 25.0,
+  "OperatingParametersAspect.MaxTemp":     180.0,
+  "iot.currentFlow":        187.3,
+  "iot.lastMaintenance":    "2026-04-22T08:00:00Z",
 }
+
+The consumer iterates ContentItem accessors (descriptor + ContentValues),
+which expose typed value access per field.
 ```
 
 ---
@@ -626,6 +627,54 @@ When the consumer provides instances of multiple different classes, the system c
 - Each field carries metadata about which content source(s) it belongs to.
 
 Merging (collapsing multiple instances into one record, marking conflicting values as "varies") is not part of this API. It is a consumer-level concern — components like the property grid perform merging on their own from the per-instance content items.
+
+**Example:** Consumer selects a `Pump` (0x3a) and a `Valve` (0x4b):
+
+```
+Targets:
+  ContentTarget { class: "ProcessPhysical:Pump", instanceIds: ["0x3a"] }
+  ContentTarget { class: "ProcessPhysical:Valve", instanceIds: ["0x4b"] }
+
+Stage 1 resolves each independently:
+  ContentSource for Pump → paths: [Pump → PumpType, Pump → OperatingParametersAspect]
+  ContentSource for Valve → paths: [Valve → ValveType]
+
+Stage 2 builds a unified descriptor (union of all fields):
+  - "Element.Name"              (string)  ← shared (both Pump and Valve inherit from Element)
+  - "Pump.FlowRate"             (double)  ← Pump only
+  - "PumpType.Manufacturer"     (string)  ← Pump only
+  - "OperatingParametersAspect.MaxPressure" (double) ← Pump only
+  - "Valve.NominalDiameter"     (double)  ← Valve only
+  - "ValveType.ConnectionType"  (string)  ← Valve only
+
+Stage 3 builds two queries (one per content source) selecting columns for all
+fields in the descriptor. Non-applicable columns SELECT NULL.
+
+Stage 4 executes the queries and produces two ContentValues:
+  ContentValues (Pump 0x3a):
+    {
+      primaryKey: { className: "ProcessPhysical:Pump", id: "0x3a" },
+      "Element.Name":                           "Main Pump",
+      "Pump.FlowRate":                          12.5,
+      "PumpType.Manufacturer":                  "Siemens",
+      "OperatingParametersAspect.MaxPressure":  25.0,
+      "Valve.NominalDiameter":                  undefined,  ← not applicable
+      "ValveType.ConnectionType":               undefined,  ← not applicable
+    }
+
+  ContentValues (Valve 0x4b):
+    {
+      primaryKey: { className: "ProcessPhysical:Valve", id: "0x4b" },
+      "Element.Name":                           "Inlet Valve",
+      "Pump.FlowRate":                          undefined,  ← not applicable
+      "PumpType.Manufacturer":                  undefined,  ← not applicable
+      "OperatingParametersAspect.MaxPressure":  undefined,  ← not applicable
+      "Valve.NominalDiameter":                  150.0,
+      "ValveType.ConnectionType":               "Flanged",
+    }
+
+The consumer iterates ContentItem accessors (descriptor + ContentValues), which expose typed value access per field.
+```
 
 ### Field identity must be stable
 
--- a/packages/content/ts-content-rewrite-proposal.md
+++ b/packages/content/ts-content-rewrite-proposal.md
@@ -752,8 +752,7 @@ The pipeline exposes an **async iterator** — consumers `for await` over conten
 
 3. ~~**Streaming vs. paging for large result sets:**~~ **Resolved.** The pipeline exposes an async iterator to consumers. Internally, it pages using efficient cursor-based strategies (e.g., keyset pagination rather than `OFFSET`) hidden behind the iterator. Consumers simply `for await` over content items without worrying about page boundaries or paging strategy.
 
-4. **Cross-iModel content:**
-   Is there a use case for loading content across multiple iModels in a single request? If so, how does the target/source model extend?
+4. ~~**Cross-iModel content:**~~ **Resolved.** No need identified. The pipeline targets a single iModel. Consumers that need data from multiple iModels can run separate pipelines and combine results themselves.
 
 5. ~~**Undo/redo interaction:**~~ **Resolved.** Consumer's responsibility. The pipeline treats the descriptor as an opaque input — it does not track modification history or support undo.
 
--- a/packages/content/ts-content-rewrite-proposal.md
+++ b/packages/content/ts-content-rewrite-proposal.md
@@ -14,7 +14,7 @@ Replace the native-backed presentation content pipeline with a pure TypeScript i
 
 ### Content target
 
-The starting point for a content request, scoped to a **single EC class**. Answers the question: "what am I getting properties *for*?"
+The starting point for a content request, scoped to a **single EC class**. Answers the question: "what am I getting properties _for_?"
 
 A content target specifies:
 
@@ -22,9 +22,9 @@ A content target specifies:
 - Optionally, a set of **instance IDs** of that class (to scope to specific instances rather than all instances of the class).
 - Optionally, an **instance filter** (a predicate to further restrict which instances are in scope, beyond explicit IDs).
 
-When a consumer selects multiple instances of *different* classes, this becomes multiple content targets — one per distinct class. The higher-level consumer API accepts a mixed collection and groups by class internally.
+When a consumer selects multiple instances of _different_ classes, this becomes multiple content targets — one per distinct class. The higher-level consumer API accepts a mixed collection and groups by class internally.
 
-The distinction between "class only" and "class + instance IDs" matters for source resolution: when instance IDs are provided, the system queries only those specific instances to determine which relationship paths are relevant (e.g., only include a path to an aspect class if at least one of the specified instances actually *has* an aspect of that class). Without instance IDs, the system must query all instances of the target class to discover which related classes exist in the data.
+The distinction between "class only" and "class + instance IDs" matters for source resolution: when instance IDs are provided, the system queries only those specific instances to determine which relationship paths are relevant (e.g., only include a path to an aspect class if at least one of the specified instances actually _has_ an aspect of that class). Without instance IDs, the system must query all instances of the target class to discover which related classes exist in the data.
 
 ### Relationship path
 
@@ -58,7 +58,7 @@ Fields declared against the target class itself require no relationship path (ze
 
 A content source defines everything needed to construct the FROM + JOINs of an ECSQL query for one target class. The SELECT clause (which columns) is determined later by Stage 2, which reads EC schema metadata and applies the property specs to generate fields.
 
-Which relationship paths are included depends on both the target class *and* the target instance IDs:
+Which relationship paths are included depends on both the target class _and_ the target instance IDs:
 
 - Without instance IDs: paths are determined by querying all instances of the target class to find which related classes actually exist in the data.
 - With instance IDs: paths are determined by querying only the specified instances (e.g., only include a path to `ElementUniqueAspect` subclass X if at least one of the given instances actually has an aspect of that class).
@@ -69,7 +69,7 @@ Multiple content sources (from multiple targets) are fed together into descripto
 
 ### Content descriptor
 
-The schema of the content result. Computed *before* loading any values. Describes *what fields exist* — it is purely structural and does not carry request-level concerns like sorting, filtering, or paging. Contains:
+The schema of the content result. Computed _before_ loading any values. Describes _what fields exist_ — it is purely structural and does not carry request-level concerns like sorting, filtering, or paging. Contains:
 
 - The list of **content sources** that were used to compute it (one per target class).
 - The full list of fields, organized as a two-level structure:
@@ -176,7 +176,7 @@ Executes the ECSQL and materializes rows into content items. After materializing
 - `getSize` only needs Stage 1 (for the join shape) + a simplified query build (COUNT over the join shape — no descriptor or field enumeration needed).
 - `getInstanceKeys` only needs Stage 1 + a simplified query build (SELECT keys from the join shape — no descriptor needed).
 - `getDescriptor` only needs stages 1–2.
-- `getDistinctValues` needs a simplified query build from a single field's metadata (the field already carries its path and class — combined with the content target, that's enough to build the DISTINCT query directly). However, consumers expect distinct *display* values, not raw values — see Open Questions.
+- `getDistinctValues` needs a simplified query build from a single field's metadata (the field already carries its path and class — combined with the content target, that's enough to build the DISTINCT query directly). However, consumers expect distinct _display_ values, not raw values — see Open Questions.
 - `getItems` runs stages 1–4.
 
 The pipeline ends at raw content items. Any post-processing (formatting, merging, label resolution) is the consumer’s responsibility, aided by composable utility functions (see Consumer Utilities below).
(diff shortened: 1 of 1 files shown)


packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,491 @@
...
+   `getDistinctValues` runs `SELECT DISTINCT` on raw values, but consumers (e.g., table filter dropdowns) want distinct _display_ values. Multiple raw values may format to the same display string (e.g., different timestamps → same date, different precisions → same rounded number). Options:
+   - **A) Return raw, consumer deduplicates after formatting:** Simple for us, more work to consumers.
+   - **B) Accept a formatter function as input:** The pipeline fetches raw values, applies the consumer-provided formatter, then deduplicates. Formatting stays a consumer concern, but the pipeline handles deduplication.
+
+   This interacts with the "formatting is a frontend concern" decision — need to reconcile.
+
+9. **SQLite/ECSQL query limits**
+
+   SQLite imposes hard limits that directly constrain the queries this pipeline can generate. The most relevant ones:
+   - **64 tables in a JOIN** (hard limit, cannot be raised) — each relationship path step adds at least one JOIN. A content source with many related classes can exhaust this quickly.
```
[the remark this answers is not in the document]
grigasp: from the proposed library's point of view it's a hard limit

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,494 @@
...
+- **Category** — logical grouping for UI display.
+- **Flags** — read-only, hidden, merged, etc.
+- **Renderer / editor hints** — optional overrides for how to display or edit the value.
+
+Field kinds:
+
+1. **Property field** — backed by a real EC property. Carries the property's class and name.
+2. **SQL calculated field** — carries an ECSQL expression evaluated in the query. The query builder includes it in the SELECT clause; the value comes back from the database like any other column. Can participate in sorting, filtering, and distinct values.
+3. **Computed field** — carries a TS function evaluated during value loading (Stage 4). After SQL results are materialized, the value loader calls the function with the partially-populated content item (all SQL-sourced values are present). The value does not exist in the query and cannot participate in SQL-level sorting or filtering.
```
saskliutas: Can you give more context on this kind of field? If I understand correctly it allows to define "custom" field that might be contructed from values of other fields or some hard coded values? If it's based on TS/JS function how will serialization work if we want to cache/persist descriptor for reusing?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,494 @@
...
+
+Describes a single data column in the content result. Key attributes:
+
+- **Identity** — a stable key that uniquely identifies this field across descriptor rebuilds. Must survive label changes and category moves. Derived from: source class + property name + path (for related), or a stable name for calculated fields.
+- **Label** — display name shown to the user.
+- **Value type** — primitive type, struct, array, or navigation.
+- **Relationship path** — which relationship path this field comes from, and which class along that path the field belongs to. Empty path means the field belongs to the target class directly. The path tells the query builder how to JOIN to the field's class; the class identifies which table the column lives in.
+- **Category** — logical grouping for UI display.
+- **Flags** — read-only, hidden, merged, etc.
+- **Renderer / editor hints** — optional overrides for how to display or edit the value.
```
saskliutas: I guess this is needed to support existing presentation rules that might be embedded in iModel? It would be nice to drop it from field as it feels more like UI/Frontend information for presetnating data.

Found later in this document that editor/renderer can be specified by property provider as well. It makes sense to have them in that place but it still feels weird that while building content we indicate what editors/renderers should be used while they ight not be available in UI.

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,494 @@
...
+
+**Provider property specs vs. descriptor transformers:** Both can customize field metadata (labels, categories, display flags), but they serve different purposes:
+
+- **Provider property specs** are **local** — they express knowledge intrinsic to the path being declared. The provider knows the domain semantics of its path and sets initial metadata accordingly (e.g., "call `UserLabel` → `Name`", or define a `source_information` category and assign properties to it). This metadata is always appropriate when this path is included. Categories are defined as lightweight objects (`{ id, label, description? }`) alongside the path declaration — this is domain knowledge, not UI layout.
+- **Descriptor transformers** are **global** — they see the full field list across all providers and make cross-cutting decisions that no single provider could make (e.g., "hide all read-only fields", "reorder by priority", "move all `BisCore` fields to a System category").
+
+Provider specs run first (during field generation), transformers run second (after all fields are collected). If a transformer conflicts with a provider spec, the transformer wins — it has full context.
```
saskliutas: This sentence is a little bit confusing for me. From what I understand `Property specs` are comming from stage 1 and are not something that "executes" like transformer but just an input while running field generation. Maybe it should be - `Provider specs are applied during first step (field generation)`? Or am I missing something?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,494 @@
...
+2. ~~**Descriptor caching and invalidation:**~~ **Resolved.** Consumer's responsibility. The pipeline does not track descriptor staleness or versioning — consumers decide when to rebuild (e.g., after schema changes or user actions).
+
+3. ~~**Streaming vs. paging for large result sets:**~~ **Resolved.** The pipeline exposes an async iterator to consumers. Internally, it pages using efficient cursor-based strategies (e.g., keyset pagination rather than `OFFSET`) hidden behind the iterator. Consumers simply `for await` over content items without worrying about page boundaries or paging strategy.
+
+4. **Cross-iModel content:**
+   Is there a use case for loading content across multiple iModels in a single request? If so, how does the target/source model extend?
+
+5. ~~**Undo/redo interaction:**~~ **Resolved.** Consumer's responsibility. The pipeline treats the descriptor as an opaque input — it does not track modification history or support undo.
+
+6. **External value sources (non-iModel data):**
```
saskliutas: This might be not an existing use case but I still want to share it: external values that depends not on primary key but on some property value. E.g. instead of using instance id to load external values consumer might want to use some property from element aspect. In this case is there a way to load that property with content but hide in the UI? I guess this could be done by consumer after stage 4 when external values are loaded and content is formatted/merged/reduced?

saskliutas (commented):
Looks good to me. I might be biased to the existing approach in native but at least in this concept it feels that each part of content building is better separated/defined and have only single responsibility. A lot of pain points that exist in native implementation are removed/moved to more appropriate place/step.

packages/content/ts-content-rewrite-proposal.md
grigasp: I definitely want to avoid having UI data on the fields. Ideally, we should be carrying enough metadata for the renderer/editor system to know what renderer/editor to use.

AFAIK, in most cases we want specific renderer/editor assigned to specific ECProperty. Maybe the renderer/editor system should be capable of identifying that from PropertyRecord / Field? Can the new editor system do that?

packages/content/ts-content-rewrite-proposal.md
grigasp: > In this case is there a way to load that property with content but hide in the UI?

You mean a case like using "External source ID" to load some data, but not showing the "External source ID" property itself in the UI?

saskliutas: Yes, something like that.

packages/content/ts-content-rewrite-proposal.md
grigasp: rephrased

packages/content/ts-content-rewrite-proposal.md
saskliutas: I think in majority of use cases we don't need `editor/renderer` fields to determine what to use already. There are only some niche cases:
- QuantityEditors. But in the new system default one from AppUI should be enough for us.
- NavigationPropertyRenderer that allows to select target by clicking value. This ones feels UI specific (not all componetns might want to render like this) and probably could be set during stage 4 by consumer as it knows if this is necessary/will work.

And then we have custom editors that I think should be handled by consumer. E.g. property-grid package registers some custom editors and have some logic that maps content into PropertyRecord and knows which properties should use custom editors.

At the moment it feels like this is UI specific like formatting and should be handled as the last step before rendering. Unless I am missing something.

packages/content/ts-content-rewrite-proposal.md
grigasp: reconsidered after your question, I think we can get away without them - updated the doc

packages/content/ts-content-rewrite-proposal.md
grigasp: Yeah I agree, I'm just worried that with the current renderer/editor capabilities we won't be able to assign specific editor to a property like `MySchema.MyClass.MyProperty`. We have to make sure the renderer/editor system is capable of finding this metadata on the properties.

saskliutas: I am not entirely sure that this is responsibility of editor system itself. I think this should be a step between raw content -> UI like formatting. Editor/Renderer systems already support preferred editor/renderer so that could be set when mapping raw content -> component specific data.

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,494 @@
...
+
+Not part of the content item:
+
+- **Labels** — handled by a separate labels API. No need to duplicate here.
+- **Display values** — formatting is a frontend concern, done from raw values at render time.
+- **Merged field names** — property merging (collapsing multiple instances into one record with "varies" markers) is a UI-level concern handled by components that care about it (e.g., property grid), not the content pipeline.
+
+---
+
+## Pipeline
```
JonasDov, line 129: Maybe you could add some sort of concrete example (simple one) which would showcase what is the input and output of each stage?

packages/content/ts-content-rewrite-proposal.md
grigasp: Here're the cases I see:

1. Assign a renderer/editor for a property based on its type. This applies to any properties, not necessarily backed by EC. The generic renderers/editors are assigned by default, but overriding at app/component level should be possible. And I think the new editor system fully supprorts this.

2. Assign a renderer/editor for an EC-backed property. Examples:
   - As an app component developer, I want `SchemaX.ClassY.PropertyZ` to use `Switch` editor instead of `Checkbox` editor.
   - As a domain developer, I add `SchemaX.ClassY.PropertyZ` property to `SchemaX` schema and want it to use my own editor component that I deliver with my domain package. In this case, I don't know/care what component the editor will be used in.

   AFAICT, the new editors' system at least partially supports both cases through `EditorsRegistryProvider`. What's unclear to me is how in the second case the domain developer can make his editor to be used for the property he added to schema.

saskliutas: > As an app component developer, I want SchemaX.ClassY.PropertyZ to use Switch editor instead of Checkbox editor.

I might be biased to my idea but I think this should be handled like I stated in previous reply. Component developer adds editor info to PropertyRecord when mapping raw content -> UI data.

> As a domain developer, I add SchemaX.ClassY.PropertyZ property to SchemaX schema and want it to use my own editor component that I deliver with my domain package. In this case, I don't know/care what component the editor will be used in.

This one I guess could go both ways. What I don't like about having editor/renderer in property spec is that like in this example domain developer expects his editor to be used but that won't be the case without consumer actually registering his editor. I guess it would be more convenient if editor would be specified in property spec and it would be enough for consumer just to add spec to pipeline and editor to registry.

Other approach for domain developer could be to export not only editor but also an utility that could be used by consumer during raw content -> UI data mapping. That way consumer would explicitly register custom editor and use utility to determine which fields/properties should use it.

However, this approach seems more verbose and probably more complicated on consumer side. Although it still feels wrong to store UI specific data on content source. Do we have actual use cases for domain developers wanting to have custom editor used for property? Maybe we could do first pass without storing this on content source and add latter? It shouldn't be a breaking change just to add optional props in a couple places. It does not looks like these props would be used in any internal logic while creating content.

grigasp:
Changes:

1. **Removed computed fields** — only two field kinds remain (property + SQL calculated). External data uses the new External fields provider.
2. **New "External fields provider"** — declares fields + populates them post-query via `resolve` callback. Inputs selected from finalized descriptor (late-binding).
3. **Renamed "Properties provider" → "iModel fields provider"** — returns `FieldsProviderContribution` with `relatedProperties`, `calculatedFields`, and `categories`.
4. **Renamed `PathDeclaration` → `RelatedPropertiesDeclaration`** — categories lifted to provider contribution level.
5. **Narrowed descriptor transformer** — no adding fields, no reordering. Only hide/remove/modify metadata.
6. **Clarified hidden vs. removed** — hidden fields are still queried; removed fields are not.
7. **Display ordering is post-pipeline** — consumer's responsibility, not the descriptor's.

packages/content/ts-content-rewrite-proposal.md
grigasp: please see recent changes, especially the "External fields provider"

packages/content/ts-content-rewrite-proposal.md
grigasp: added

packages/content/ts-content-rewrite-proposal.md
saskliutas: Looks good. This kind of explains TS-based fields from initial plan. 😆 One concern that I had with TS-based fields is related to descriptor caching. It looks like descriptor is almost a plain object that can be easily serialized and reused between run, except for external fields. Do we case about the ability to serialize descriptor easily?

packages/content/ts-content-rewrite-proposal.md
JonasDov: Thanks much clearer now.

packages/content/ts-content-rewrite-proposal.md
grigasp: > Do we case about the ability to serialize descriptor easily?

I think we should - POC shows that it can still take tens of seconds to create it. Added a section with my thoughts - please have a look.

packages/content/ts-content-rewrite-proposal.md
grigasp: > What I don't like about having editor/renderer in property spec

Let's be clear - I don't like that either 👍 And I think we can avoid them if the renderer/editor system is capable of identifying properties by:
- ECProperty identifier (`schema.class.property`),
- For external fields - probably stable name?

If that's possible, then we don't need the renderer/editor attributes on field spec. Domain developers can:
1. Add a property to their schema.
2. Deliver an `EditorSpec` that targets that specific property.

Then on the app side, it's app developer's responsibility to register the domains' editor specs. If that doesn't happen, we fall back to default editors based on property type.

packages/content/ts-content-rewrite-proposal.md
saskliutas: Looks good. I guess it makes sense to cache content source instead of descriptor.

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,697 @@
...
+Query returns 1 row. Value loader maps columns to iModel fields.
+Then pipeline calls external fields provider's resolve(items):
+  - Provider reads "Pump.Name" = "Main Circulation Pump" from each item
+  - Fetches sensor data from IoT service
+  - Attaches "iot.currentFlow" and "iot.lastMaintenance" values to items
+
+Output: ContentItem {
+  primaryKeys: [{ class: "ProcessPhysical:Pump", id: "0x3a" }],
+  values: {
+    "Pump.Name":              "Main Circulation Pump",
```
JonasDov, line 317: Lets say i am the caller of `getItems`, how do i know the types of those values? Do i need to do a separate request to `getDescriptor`, and then determine the types from there?
If content item had value & type, then I wouldn't need to do a call to `getDescriptor`? e.g.:
```ts
type ContentItem = {
  ...
  values: {
    [key: string]: { value: string; type: "string" } | { value: number; type "number" } | { value: string; type: "dateTime" } ...
  }
}

const contentItem: ContentItem = {
  primaryKeys: [{ class: "ProcessPhysical:Pump", id: "0x3a" }],
  values: {
    "Pump.Name": { value: "Main Circulation Pump", type: "string" },
    "Pump.FlowRate": { value: 12.5, type: "number" )
  ...
    "iot.lastMaintenance":    { value: "2026-04-22T08:00:00Z", type: "dateTime" }
  }
}
```

grigasp: Good point, this touches one of pain points that we have today, where we have a similar architecture - fields (metadata) separate from items (data).

Maybe the "pipeline" should have one more stage, or we should enhance the existing stage 4 to do the association? @saskliutas, interested in your thoughts.

saskliutas: Wouldn't enhancing stage 4 would require to have descriptor already? I guess this defeats the purpose of initial comment that we should be able to know types of values only running state 4 without having descriptor?

However, I think that descriptor should be available when running queries already so we could have stage 5 that would map content + descriptor into some more convenient format. Maybe it could be incorporated into stage 4 as well as you suggest. But the end result in my view is that you will need to have descriptor when running stage 4 regardless.

packages/content/ts-content-rewrite-proposal.md
grigasp: Since we agreed on the first part, I removed renderer and editor attributes from the data model.

However we decide to do it, binding domain properties to domain renderers/editors is a UI concern and is out of scope for this document.

Added a note to the document, please have a look.

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,731 @@
...
+  // Descriptor contribution (applied during Stage 2)
+  fields: FieldDeclaration[]; // fields this provider will populate
+  categories?: CategoryDefinition[]; // shared pool referenced by fields via categoryId
+
+  // Input discovery (called after descriptor is finalized)
+  inputs?: (descriptor: Descriptor) => InputFieldRef[];
+
+  // Value population (called during Stage 4)
+  resolve: (items: ContentItem[]) => Promise<void>;
+}
```
JonasDov, line 494: As i understand, `ExternalFieldsProvider` has two main purposes:
1. In stage 2. Provide the descriptors external field types.
2. In stage 4 Provide the values of those external fields.

What i don't fully understand:
1. Why does `externaFieldsProvider` responsible for hiding descriptor fields? Isn't this the purpose of Descriptor transformer?
2. Does resolve depend on `ContentItem[]` to know what are the external values are?
3. This api allows to modify the `ContentItem[]` in any way they want (add ContentItems, remove them, modify them). Why not make the resolve provide the extended content items that should be added to the array?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+```
+
+The `Descriptor` provides utility methods for field lookup — find by class + property name, by label, by path, by category, etc. External fields provider authors use these utilities inside `inputs` rather than depending on internal identity encoding.
+
+External fields providers run during Stage 4, after query execution but before items are emitted to the consumer. They enable external data to be first-class in the descriptor — their declared fields appear in the field list, carry categories and metadata, and can be hidden/shown like any other field.
+
+**Constraints:**
+
+- External fields cannot participate in SQL-level sorting, filtering, or distinct values (their data doesn't exist in the query).
+- External fields providers must not modify fields they don't own.
```
JonasDov, line 513: We can mark the fields as `readonly`?

packages/content/ts-content-rewrite-proposal.md
saskliutas: Looks ok. Although the note in `Field (property descriptor)` about editor/renderer looks out of place for me. Maybe it should be moved under `Key Design Decisions`?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+
+**What to cache:** Only the **content sources** — the output of Stage 1 (resolved concrete paths + calculated field declarations per target). The descriptor itself is not cached.
+
+**Why cache only Stage 1:** Stage 2 is cheap (in-memory metadata reads), so re-running it from cached sources adds negligible cost. This avoids all complexity around "restoring" a descriptor:
+
+- No partial re-application of external fields.
+- No "delta" transformer pass.
+- Transformers, external fields providers, and iModel field metadata all run fresh and uniformly on every descriptor build.
+
+**Cache persistence:** The cache is persistent across sessions (disk-backed).
```
JonasDov, line 561: How will we achieve this?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+**Restore flow:**
+
+1. Deserialize cached content sources (concrete paths + calculated field declarations).
+2. Run Stage 2 fresh — enumerate fields from EC schema metadata, append external fields, call `inputs`, run transformers.
+3. Descriptor is ready for Stage 3/4.
+
+**Cache key:** `(content target, iModel data version, sorted iModel fields provider ID set)`.
+
+- **Content target** — the full content target as provided by the consumer (class, optional instance IDs, optional instance filter). Different targets produce different cache entries.
+- **iModel data version** — changeset ID or a monotonic counter incremented on write transactions. Path resolution depends on actual relationship instances (e.g., associating a new aspect to an element adds fields), not just schema structure.
```
JonasDov, line 572: > monotonic counter incremented on write transactions

maybe changeset ID is enough?

packages/content/ts-content-rewrite-proposal.md
saskliutas: I think achieving this is out of scope of this document? In my view that should be an implementation detail and this document should focus on decisions that would allow us to achieve this.

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+
+For **1:many relationships** (e.g., element → multi aspects of the same class), naive JOINing causes row explosion — if an element has 1000 aspects, the JOIN produces 1000 rows, each duplicating all element properties. The loading strategy for 1:many needs careful consideration (see Open Questions).
+
+**Cardinality cannot be trusted from the schema alone.** EC relationship definitions frequently declare `many-to-many` even when the actual data invariant is `1:1` (e.g., `ElementOwnsUniqueAspect` is declared `(1:N)` but each element has at most one unique aspect per class). Therefore:
+
+- **Providers can supply a cardinality hint** on their related properties declarations (e.g., "this step is always 1:1 in practice"). The system trusts the hint and uses it to choose the loading strategy.
+- **For paths without hints**, the system can determine effective cardinality during path resolution (Stage 1) — the same query that resolves concrete classes can also check whether any primary instance has more than one related instance along the path. This is a data-driven check, not a schema-driven one.
+
+### Multi-target behavior
```
JonasDov, line 622: Could you provide some examples for this, not sure I fully follow...

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+## Open Questions
+
+1. ~~**Calculated fields at query vs. post-processing level:**~~ **Resolved.** Only **SQL calculated fields** exist as a pipeline-level concept (ECSQL expression, evaluated in query, participates in sorting/filtering). Per-item value derivation from already-loaded data is handled by consumer utilities (`map`). See "No pipeline-level computed fields" in Key Design Decisions.
+
+2. ~~**Descriptor caching and invalidation:**~~ **Resolved.** Consumer's responsibility. The pipeline does not track descriptor staleness or versioning — consumers decide when to rebuild (e.g., after schema changes or user actions).
+
+3. ~~**Streaming vs. paging for large result sets:**~~ **Resolved.** The pipeline exposes an async iterator to consumers. Internally, it pages using efficient cursor-based strategies (e.g., keyset pagination rather than `OFFSET`) hidden behind the iterator. Consumers simply `for await` over content items without worrying about page boundaries or paging strategy.
+
+4. **Cross-iModel content:**
+   Is there a use case for loading content across multiple iModels in a single request? If so, how does the target/source model extend?
```
JonasDov: I could think of one use case: comparing changes across different imodel versions?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+
+7. **1:many related content loading strategy:**
+   For 1:1 relationships, JOINing into the main query is fine. For 1:many, naive JOINs cause row explosion. Options:
+   - **A) Separate queries + stitch:** Run a separate query per 1:many path, keyed by primary instance ID. Stitch results into content items during value loading. Clean separation, but multiple round-trips and complexity in Stage 4.
+   - **B) `GROUP_CONCAT` / aggregation in SQL:** Use SQLite’s `GROUP_CONCAT` (or `json_group_array`) to aggregate 1:many values into a single column per row. Single query, no row explosion, but parsing aggregated values adds complexity and there may be limits on column size.
+   - **C) Hybrid:** Use aggregation for small cardinalities and separate queries for large ones, with a threshold.
```
JonasDov: I see option A as the least complex and maybe it's good enough for initial implementation? I see this as an internal implemetation detail, which in my eyes should not affect the API. I would rather go with the simplest approach on the initial implemenation (since there is no clear best choice) and then we can improve later?

packages/content/ts-content-rewrite-proposal.md
saskliutas: Wouldn't it be better for comparison to make 2 requests: load content for iModel v1 and content for iModel v2 then compare? How would it help to merge content from both iModels?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+
+   Need to evaluate trade-offs around query count, parsing complexity, paging interaction, and typical cardinalities in real iModels.
+
+8. **Distinct values: raw vs. formatted:**
+   `getDistinctValues` runs `SELECT DISTINCT` on raw values, but consumers (e.g., table filter dropdowns) want distinct _display_ values. Multiple raw values may format to the same display string (e.g., different timestamps → same date, different precisions → same rounded number). Options:
+   - **A) Return raw, consumer deduplicates after formatting:** Simple for us, more work to consumers.
+   - **B) Accept a formatter function as input:** The pipeline fetches raw values, applies the consumer-provided formatter, then deduplicates. Formatting stays a consumer concern, but the pipeline handles deduplication.
```
JonasDov: I would go with option A, since the pipeline never cares about display values or formatting, it should not care in this place. Can they not use the `map` or `reduce` utility function?

packages/content/ts-content-rewrite-proposal.md
```diff
@@ -0,0 +1,730 @@
...
+
+   This interacts with the "formatting is a frontend concern" decision — need to reconcile.
+
+9. **SQLite/ECSQL query limits**
```
JonasDov: I would like to keep the initial implementation as simple as possible. Maybe fill out separate issues for things listed here, could come back later to these when initial implementation is complete?

packages/content/ts-content-rewrite-proposal.md
JonasDov: @saskliutas  I'm talking about this comment:
> `getItems` runs stages 1–4.

As I understand the input to `getItems` would be something like `ContentTarget { class: "ProcessPhysical:Pump", instanceIds: ["0x3a"] }`. In the example, returned `contentItem` would be:
```ts
const contentItem = {
  primaryKeys: [{ class: "ProcessPhysical:Pump", id: "0x3a" }],
  values: {
    "Pump.Name": "Main Circulation Pump",
    ...
  }
}
```

In such case, i don't have information about the types of those values. I would need to call `getDescriptor` and then join them together?

packages/content/ts-content-rewrite-proposal.md
JonasDov: Maybe `caching across sessions` should not be part of initial implementation then, and for initial implementation make a new cache for each session?

packages/content/ts-content-rewrite-proposal.md
saskliutas: I think in this case there is no reason not to return descriptor as well as items? We will need to build descriptor in the process so we can return it as well. Or as you suggested include additional metadata into content item. However, there might be cases where you call getItems and need more metadata than included in content items and you will need descriptor or we will need to introduce more data into content item.

I guess ideal approach would be to have frequently used metadata on content item like `type` and return descriptor as well?

packages/content/ts-content-rewrite-proposal.md
saskliutas: I am not sure if it should be implemented by presentation-content at all. We can provide utilities for caching but I think consumer should decide where to store descriptor between sessions.

packages/content/ts-content-rewrite-proposal.md
JonasDov: I guess you are right. If need be we could always add a wrapper which does something like this?

packages/content/ts-content-rewrite-proposal.md
grigasp: > 1. Why does `externaFieldsProvider` responsible for hiding descriptor fields? Isn't this the purpose of Descriptor transformer?

That requirement is based on [this comment](https://github.com/iTwin/presentation/pull/1343#discussion_r3233444178). But I see your point - external fields provider should not request fields to be hidden - it should requested them to be available to it. So if a field is requested to be shown by imodel properties provider - show it, otherwise only make it available, but keep it hidden. Updated the doc.

> 2\. Does resolve depend on `ContentItem[]` to know what are the external values are?

It can. The system ensures that items have values for the requested `input` fields (I hope we can express that intent through the type system).

> 3\. This api allows to modify the `ContentItem[]` in any way they want (add ContentItems, remove them, modify them). Why not make the resolve provide the extended content items that should be added to the array?

This document is not defining API, the TS code in there to act as an example. The actual API will be decided in later steps.

packages/content/ts-content-rewrite-proposal.md
grigasp: It think it makes no sense to request (or allow requesting) values without having the descriptor. Whether the descriptor is an input to `getItems` call or is retrieved internally when calling `getItems` is not in the context of this document and will be decided later.

So assuming we have both - the descriptor and the values - would you agree that we should return them in a structure where metadata sits next to the value?

packages/content/ts-content-rewrite-proposal.md
grigasp: Yes. But that's an API decision, which we'll need to make in later steps.

packages/content/ts-content-rewrite-proposal.md
saskliutas: Yes, I don't see anything against merging values and metadata into single data structure.

packages/content/ts-content-rewrite-proposal.md
grigasp: Yes, caching it is not our responsibility. The idea behind this section is to ensure it can be cached in a serialized way (on disk or some service).

packages/content/ts-content-rewrite-proposal.md
grigasp: Probably not for editing apps. This is an internal detail which we'll figure out at implementation stage.

packages/content/ts-content-rewrite-proposal.md
grigasp: moved

packages/content/ts-content-rewrite-proposal.md
grigasp: Made some changes - we now have `ContentValues` for raw values and `ContentItem` which holds the descriptor and values, and provides convenient access to fields/values. API specifics can be decided later.

packages/content/ts-content-rewrite-proposal.md
grigasp: Added an example

packages/content/ts-content-rewrite-proposal.md
grigasp: Set this as resolved - no need.

packages/content/ts-content-rewrite-proposal.md
grigasp: >  I see this as an internal implemetation detail, which in my eyes should not affect the API.

Agree, moved this out of "open questions" to a new "implementation notes" section. We can decide on this later, but I do agree that option A sounds like the way to go.

packages/content/ts-content-rewrite-proposal.md
grigasp: Moved this section to "implementation notes".

packages/content/ts-content-rewrite-proposal.md
grigasp: `reduce` would be what they need to use, but all consumers would be doing the same boilerplate, which I'd like to avoid. Maybe provide a function built on top of `reduce` like:

```ts
function groupFormatted(items: Array<{ rawValue, field }>, formatter): Map<string, RawValue[]> {
  return rawValues.reduce((formattedGroups, rawValue) => {
    const formatted = formatter(rawValue, field);
    let group = formattedGroups.get(formatted);
    if (!group) {
      group = [];
      formattedGroups.set(formatted, group);
    }
    group.push(rawValue);
  }, new Map());
}
```

JonasDov (approved):
LGTM

saskliutas (approved):

PR #1347, opened 2026-05-14
Replace dependabot with custom dependencies update action

saskliutas:
~~Avoid major bumps when bumping deps~~

Replaced dependabot with custom actions that runs `pnpm -r update`. Dependabot does not support bumps following specified range.


JonasDov (approved):

.github/dependabot.yml
```diff
@@ -6,6 +6,10 @@ updates:
       interval: "weekly"
       day: "sunday"
     versioning-strategy: "increase"
+    allow:
+      - update-types:
+          - "version-update:semver-minor"
+          - "version-update:semver-patch"
```
grigasp: can't we ask to update according to given semver range?

grigasp: `minor` is not safe for `0.x`. also, not every package uses semver, so even `minor` can be unsafe

saskliutas: Dependabot does not support updates based on specified range. Switched to custom action. 😞

grigasp (approved):

PR #1357, opened 2026-05-21
`content`: API

grigasp:
- Updated the plan to match the proposed API (file renamed to `ts-content-rewrite.md`).
- I'm planning to extract the `InstanceFilter` / `RelationshipPathStep.instanceFilter` change in `@itwin/presentation-shared` into a separate PR (https://github.com/iTwin/presentation/pull/1360).
- Looking for feedback, especially for the open questions in `ts-content-api.md`

### DIFF
--- a/packages/content/src/content/Content.ts
+++ b/packages/content/src/content/Content.ts
@@ -1,0 +1,238 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * Content pipeline overview
+ * =========================
+ *
+ * Stage 1 — Source resolution (`resolveContentSources`)
+ *   - Calls `IModelFieldsProvider.getContribution()` for each target.
+ *   - Performs schema introspection and resolves relationship paths
+ *     (including provider-declared related properties).
+ *   - Output: `ContentSource[]` — serializable, cacheable.
+ *
+ * Stage 2 — Descriptor building (`createContentProvider` → `getContentDescriptor`)
+ *   - Enumerates iModel fields from the resolved sources.
+ *   - Appends `ExternalFieldsProvider.fields` declarations.
+ *   - Runs `DescriptorTransformer.transform()` in ascending priority order.
+ *   - Output: frozen `ContentDescriptor`.
+ *
+ * Stage 3 — Query building (`getItems` / `getSize` / `getInstanceKeys`)
+ *   - Calls `QueryFilterer.getFilterClauses()` to inject WHERE/JOIN clauses.
+ *   - Applies `ContentValueFilter` entries as additional WHERE clauses.
+ *   - Builds ECSQL.
+ *     - `getSize` and `getInstanceKeys` depend only on Stage 1 (resolved sources).
+ *     - `getItems` additionally requires Stage 2 (the descriptor) to know which
+ *       columns to SELECT.
+ *
+ * Stage 4 — Query execution and value population (`getItems`)
+ *   - Executes the built ECSQL query.
+ *   - Calls `ExternalFieldsProvider.getValues()` per batch of items.
+ *   - Merges external values into the final `ContentItem` accessors.
+ */
+
+import type { ECSqlQueryExecutor, ECSchemaProvider, InstanceKey, Value } from "@itwin/presentation-shared";
+import type { ContentSource, ContentTarget } from "./ContentTarget.js";
+import type { DescriptorTransformer } from "./extensions/DescriptorTransformer.js";
+import type { ExternalFieldsProvider } from "./extensions/ExternalFieldsProvider.js";
+import type { IModelFieldsProvider } from "./extensions/IModelFieldsProvider.js";
+import type { QueryFilterer } from "./extensions/QueryFilterer.js";
+import type { ContentDescriptor } from "./model/ContentDescriptor.js";
+import type { ContentItem } from "./model/ContentItem.js";
+import type { CalculatedField, PropertyField } from "./model/Field.js";
+
+/**
+ * Sorting specification for content value requests.
+ *
+ * @public
+ */
+interface ContentSortSpec {
+  /** The field to sort by. */
+  field: PropertyField | CalculatedField;
+  /** Sort direction. */
+  direction: "asc" | "desc";
+}
+
+/**
+ * A value filter applied during query building (Stage 3).
+ * Adds a WHERE clause to the final query — does not affect which fields
+ * exist in the descriptor (only which rows are returned).
+ *
+ * @public
+ */
+export interface ContentValueFilter {
+  /** The field to filter on. */
+  field: PropertyField | CalculatedField;
+  /**
+   * For composite fields (structs, points), the member to compare.
+   * Example: `"x"` for a Point3d field, `"Street"` for an Address struct.
+   * Omit for scalar fields.
+   */
+  member?: string;
+  /** The filter operator. */
+  operator: ValueFilterOperator;
+  /** The value(s) to compare against. */
+  value: Value;
+}
+
+/** @public */
+type ValueFilterOperator =
+  | "is-equal"
+  | "is-not-equal"
+  | "is-null"
+  | "is-not-null"
+  | "less-than"
+  | "less-than-or-equal"
+  | "greater-than"
+  | "greater-than-or-equal"
+  | "like"
+  | "is-in";
+
+/**
+ * Request options passed alongside the descriptor when loading values.
+ * Controls _how_ to query, not _what fields exist_.
+ *
+ * The same descriptor can be reused with different request options
+ * (different pages, sort orders, filters).
+ *
+ * @public
+ */
+interface ContentRequestOptions {
+  /** Sorting specification. Applied as ORDER BY in the generated query. */
+  sorting?: ContentSortSpec[];
+
+  /**
+   * Value filters. Applied as additional WHERE clauses during query building.
+   * Multiple filters are ANDed together.
+   * Does not affect the descriptor — only which rows are returned.
+   */
+  filters?: ContentValueFilter[];
+}
+
+/**
+ * App-level extension point registration shared by both `resolveContentSources`
+ * and `createContentProvider`.
+ *
+ * @public
+ */
+export interface ContentConfiguration {
+  /** iModel fields providers (contribute related properties and calculated fields). */
+  fieldsProviders?: IModelFieldsProvider[];
+
+  /** External fields providers (declare + populate fields from outside the iModel). */
+  externalFieldsProviders?: ExternalFieldsProvider[];
+
+  /** Descriptor transformers (modify descriptor after field enumeration). */
+  descriptorTransformers?: DescriptorTransformer[];
+
+  /** Query filterers (inject WHERE clauses into built queries). */
+  queryFilterers?: QueryFilterer[];
+}
+
+/**
+ * Props for resolving content sources.
+ *
+ * @public
+ */
+interface ResolveContentSourcesProps {
+  /** Access to the iModel for schema introspection and path resolution. */
+  imodelAccess: ECSqlQueryExecutor | ECSchemaProvider;
+  /** The content targets to resolve. */
+  targets: ContentTarget[];
+  /** Extension point configuration (only `fieldsProviders` is used for resolution). */
+  config?: Pick<ContentConfiguration, "fieldsProviders">;
+}
+
+/**
+ * Resolve content sources for the given targets and provider declarations.
+ *
+ * This is the expensive step — performs schema introspection and data-driven
+ * path resolution for all targets and provider-declared related properties.
+ *
+ * The returned sources capture the fully-resolved join shapes (including
+ * provider-declared related properties). They are serializable and can be
+ * cached to disk for reuse — skipping the expensive resolution on subsequent loads.
+ *
+ * Cache invalidation: sources become stale when the iModel schema changes
+ * or provider declarations change.
+ *
+ * @public
+ */
+export async function resolveContentSources(_props: ResolveContentSourcesProps): Promise<ContentSource[]> {
+  throw new Error("Not implemented");
+}
+
+/**
+ * Configuration for creating a content provider from pre-resolved sources.
+ *
+ * @public
+ */
+interface ContentProviderProps {
+  /** Access to the iModel for running ECSQL queries. */
+  imodelAccess: ECSqlQueryExecutor | ECSchemaProvider;
+
+  /** Pre-resolved content sources (output of `resolveContentSources`). */
+  sources: ContentSource[];
+
+  /** Extension point configuration. */
+  config?: ContentConfiguration;
+}
+
+/**
+ * A stateful content provider built from pre-resolved sources.
+ *
+ * Created via `createContentProvider`, which builds the descriptor from the
+ * resolved sources. The descriptor is exposed for inspection/modification
+ * before querying items.
+ *
+ * Methods that only need source/filter information (`getSize`, `getInstanceKeys`,
+ * `getDistinctValues`) do not depend on the descriptor. Only `getItems` uses it.
+ *
+ * @public
+ */
+interface ContentProvider {
+  /**
+   * Get the content descriptor for the configured sources.
+   * Built lazily on first call and cached for subsequent calls.
+   *
+   * The descriptor reflects all fields providers and descriptor transformers
+   * from the content configuration.
+   */
+  getContentDescriptor(): Promise<Readonly<ContentDescriptor>>;
+
+  /**
+   * Get the total number of content items matching the configured sources.
+   *
+   * @param options - Optional filters (affects the count).
+   */
+  getSize(options?: Pick<ContentRequestOptions, "filters">): Promise<number>;
+
+  /**
+   * Get instance keys for all items matching the configured sources.
+   *
+   * @param options - Optional filters (affects which keys are returned).
+   */
+  getInstanceKeys(options?: Pick<ContentRequestOptions, "filters">): AsyncIterable<InstanceKey>;
+
+  /**
+   * Load content items using the descriptor.
+   *
+   * Returns an async iterator over ContentItem accessors. Internally pages
+   * using cursor-based pagination; consumers simply `for await` over items.
+   *
+   * @param options - Sorting and filtering options.
+   */
+  getItems(options?: ContentRequestOptions): AsyncIterable<ContentItem>;
+}
+
+/**
+ * Create a content provider from pre-resolved sources.
+ * The descriptor is built lazily on the first call to `getContentDescriptor`.
+ *
+ * @public
+ */
+export function createContentProvider(_props: ContentProviderProps): ContentProvider {
+  throw new Error("Not implemented");
+}
--- a/packages/content/src/content/ContentTarget.ts
+++ b/packages/content/src/content/ContentTarget.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import type { Id64String } from "@itwin/core-bentley";
-import type { EC, InstanceFilter, RelationshipPath } from "@itwin/presentation-shared";
+import type { EC, ECSqlBinding, RelationshipPath } from "@itwin/presentation-shared";
 import type { BaseFieldsProvider } from "./extensions/BaseFieldsProvider.js";
 
 /**
@@ -45,7 +45,35 @@ export interface ContentTarget {
    * Optional filter predicate to further restrict which instances are in scope.
    * Applied during source resolution (Stage 1) — affects which paths are discovered.
    */
-  instanceFilter?: InstanceFilter;
+  instanceFilter?: {
+    /**
+     * ECSQL WHERE clause expression (without the WHERE keyword).
+     *
+     * Use `primaryClassAlias` (defaults to `"this"`) followed by a dot to reference properties
+     * of the primary class. At query generation time, the pipeline performs
+     * a literal replacement of all `{primaryClassAlias}.` occurrences with the actual query alias.
+     *
+     * @example
+     * ```
+     * expression: "this.Area > :minArea"
+     * ```
+     */
+    expression: string;
+
+    /**
+     * The placeholder used in `expression` to reference the primary class (`primaryClass`).
+     * Every occurrence of `{primaryClassAlias}.` in the expression will be replaced with the
+     * actual query alias at query generation time.
+     *
+     * @default "this"
+     */
+    primaryClassAlias?: string;
+
+    /**
+     * Bind values for the expression, keyed by parameter name.
+     */
+    bindings?: Record<string, ECSqlBinding>;
+  };
 }
 
 /**
--- a/packages/content/src/content/ContentTarget.ts
+++ b/packages/content/src/content/ContentTarget.ts
@@ -1,0 +1,95 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { Id64String } from "@itwin/core-bentley";
+import type { EC, InstanceFilter, RelationshipPath } from "@itwin/presentation-shared";
+import type { BaseFieldsProvider } from "./extensions/BaseFieldsProvider.js";
+
+/**
+ * Hint about the effective cardinality of a relationship path.
+ *
+ * - `"one"`: each target instance joins to at most one related instance.
+ * - `"many"`: each target instance may join to multiple related instances.
+ *
+ * @public
+ */
+export type CardinalityHint = "one" | "many";
+
+/**
+ * The starting point for a content request, scoped to a single EC class.
+ *
+ * Answers the question: "what am I getting properties for?"
+ *
+ * When the consumer selects instances of multiple different classes,
+ * this becomes multiple content targets — one per distinct class.
+ *
+ * @public
+ */
+export interface ContentTarget {
+  /**
+   * Full class name of the primary class whose properties we want (e.g., "BisCore.Element").
+   */
+  primaryClass: EC.FullClassName;
+
+  /**
+   * Optional set of instance IDs to scope to specific instances.
+   * When provided, source resolution queries only these instances to determine
+   * which relationship paths are relevant.
+   * When omitted, all instances of the class are queried.
+   */
+  instanceIds?: Id64String[];
+
+  /**
+   * Optional filter predicate to further restrict which instances are in scope.
+   * Applied during source resolution (Stage 1) — affects which paths are discovered.
+   */
+  instanceFilter?: InstanceFilter;
+}
+
+/**
+ * A resolved join shape for a single content target.
+ * Output of source resolution (Stage 1). Contains the target and the concrete
+ * relationship paths discovered by querying iModel's data.
+ *
+ * @public
+ */
+export interface ContentSource {
+  /** The content target this source was resolved from. */
+  target: ContentTarget;
+
+  /**
+   * Resolved declaration groups — one per provider declaration that produced
+   * concrete paths during resolution.
+   *
+   * Each group links back to its originating provider and declaration index,
+   * allowing Stage 2 to re-fetch the declaration's property specs and
+   * cardinality hint without storing them on the cached source.
+   */
+  resolvedDeclarations: ResolvedDeclarationGroup[];
+}
+
+/**
+ * A group of concrete relationship paths resolved from a single provider declaration.
+ *
+ * During source resolution (Stage 1), each provider's `RelatedPropertiesDeclaration`
+ * may resolve to one or more concrete paths. This group preserves that association
+ * so Stage 2 can look up the declaration's property specs and cardinality hint
+ * by re-calling the provider.
+ *
+ * @public
+ */
+interface ResolvedDeclarationGroup {
+  /** ID of the provider that contributed the originating declaration. */
+  providerId: BaseFieldsProvider["id"];
+
+  /** 0-based index into the provider's `relatedProperties` array. */
+  declarationIndex: number;
+
+  /**
+   * Concrete relationship paths resolved from the declaration's generic path.
+   * All classes are concrete — no base classes.
+   */
+  paths: RelationshipPath[];
+}
--- a/packages/content/src/content/ContentTarget.ts
+++ b/packages/content/src/content/ContentTarget.ts
@@ -84,7 +84,10 @@ interface ResolvedDeclarationGroup {
   /** ID of the provider that contributed the originating declaration. */
   providerId: BaseFieldsProvider["id"];
 
-  /** 0-based index into the provider's `relatedProperties` array. */
+  /**
+   * 0-based index into the `relatedProperties` array of the `FieldsProviderContribution`
+   * returned by the {@link IModelFieldsProvider}'s `getContribution()` call.
+   */
   declarationIndex: number;
 
   /**
--- a/packages/content/src/content/DistinctValues.ts
+++ b/packages/content/src/content/DistinctValues.ts
@@ -1,0 +1,40 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { ECSqlQueryExecutor, Value } from "@itwin/presentation-shared";
+import type { ContentValueFilter } from "./Content.js";
+import type { ContentTarget } from "./ContentTarget.js";
+import type { CalculatedField, PropertyField } from "./model/Field.js";
+
+/**
+ * Props for `getDistinctFieldValues`.
+ *
+ * @public
+ */
+interface GetDistinctFieldValuesProps {
+  /** Access to the iModel for running ECSQL queries. */
+  imodelAccess: ECSqlQueryExecutor;
+
+  /** The content targets to query against. */
+  targets: ContentTarget[];
+
+  /** The field to get distinct values for. */
+  field: PropertyField | CalculatedField;
+
+  /** Optional filters (restricts which rows contribute distinct values). */
+  filters?: ContentValueFilter[];
+}
+
+/**
+ * Get distinct raw values for a single field across the given content targets.
+ *
+ * The field itself carries the join path from the content target to the property,
+ * so resolved sources are not needed.
+ *
+ * @public
+ */
+export function getDistinctFieldValues(_props: GetDistinctFieldValuesProps): AsyncIterable<Value> {
+  throw new Error("Not implemented");
+}
--- a/packages/content/src/content/extensions/DescriptorTransformer.ts
+++ b/packages/content/src/content/extensions/DescriptorTransformer.ts
@@ -5,7 +5,7 @@
 
 import type { ContentSource } from "../ContentTarget.js";
 import type { CategoryDefinition } from "../model/Category.js";
-import type { Field, RelatedFieldGroup } from "../model/Field.js";
+import type { Field } from "../model/Field.js";
 
 // cspell:words spliceable
 
@@ -79,29 +79,18 @@ type TransformableField = Omit<Field, "identity"> & { readonly identity: string 
 type SpliceableReadonlyArray<T> = ReadonlyArray<T> & { splice(start: number, deleteCount?: number): T[] };
 
 /**
- * A related field group exposed to transformers — nested fields have frozen identities.
- *
- * @public
- */
-interface TransformableRelatedFieldGroup extends Omit<RelatedFieldGroup, "fields" | "nestedGroups"> {
-  fields: SpliceableReadonlyArray<TransformableField>;
-  nestedGroups?: SpliceableReadonlyArray<TransformableRelatedFieldGroup>;
-}
-
-/**
  * A constrained view of {@link ContentDescriptor} exposed to descriptor transformers.
  *
  * Enforces transformer rules at the type level:
  * - `sources` is readonly — the resolved source structure is immutable at this stage.
  * - Field `identity` is readonly — must not be changed.
  * - Field metadata (`label`, `categoryId`, `hidden`, `readOnly`) remains mutable.
- * - Field arrays are readonly but allow element removal via `splice`.
+ * - Field array is readonly but allows element removal via `splice`.
  *
  * @public
  */
 interface TransformableDescriptor {
   readonly sources: readonly ContentSource[];
-  readonly directFields: SpliceableReadonlyArray<TransformableField>;
-  readonly relatedFieldGroups: SpliceableReadonlyArray<TransformableRelatedFieldGroup>;
+  readonly fields: SpliceableReadonlyArray<TransformableField>;
   readonly categories: Record<CategoryDefinition["id"], CategoryDefinition>;
 }
--- a/packages/content/src/content/extensions/DescriptorTransformer.ts
+++ b/packages/content/src/content/extensions/DescriptorTransformer.ts
@@ -67,7 +67,7 @@ export function defineDescriptorTransformer(transformer: DescriptorTransformer):
  *
  * @public
  */
-type TransformableField = Field & { readonly identity: string };
+type TransformableField = Omit<Field, "identity"> & { readonly identity: string };
 
 /**
  * A related field group exposed to transformers — nested fields have frozen identities.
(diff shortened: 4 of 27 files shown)


packages/shared/src/shared/Metadata.ts
```diff
@@ -338,6 +340,96 @@ export namespace EC {
...
+  kind: "struct";
+  members: StructMember[];
+}
+
+/** @public */
+interface StructMember {
+  name: string;
+  label: string;
+  type: ValueDescriptor;
+}
```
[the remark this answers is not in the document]
grigasp: TS doesn't error on this

packages/shared/src/presentation-shared.ts
```diff
@@ -36,4 +37,5 @@ export {
   trimWhitespace,
 } from "./shared/Utils.js";
 export { InstanceKey, PrimitiveValue, TypedPrimitiveValue } from "./shared/Values.js";
+export type { ArrayValue, StructValue, Value } from "./shared/Values.js";
```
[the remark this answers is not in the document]
grigasp: Keeping this to make sure I don't forget to add the changelog

packages/content/src/content/model/Field.ts
```diff
@@ -0,0 +1,132 @@
...
+  label: string;
+  /** The value shape for this field. */
+  type: ValueDescriptor;
+  /** The category this field belongs to (by ID). */
+  categoryId?: string;
+  /** If true, the field is queried but not displayed in UI. */
+  hidden?: boolean;
+  /** If true, the field is read-only. */
+  readOnly?: boolean;
+}
```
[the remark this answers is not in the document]
grigasp: It doesn't cause any failures

packages/content/src/content/model/PropertySpec.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @public
+ */
+export interface StepPropertySpec {
+  /** 0-based position in the path. */
+  stepIndex: number;
+  /** Properties from the target class at this step. */
+  target?: ClassPropertySpec;
+  /** Properties from the relationship class at this step. */
+  relationship?: ClassPropertySpec;
+}
```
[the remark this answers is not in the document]
grigasp: works fine

packages/content/src/content/ContentTarget.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /** 0-based index into the provider's `relatedProperties` array. */
+  declarationIndex: number;
+
+  /**
+   * Concrete relationship paths resolved from the declaration's generic path.
+   * All classes are concrete — no base classes.
+   */
+  paths: RelationshipPath[];
+}
```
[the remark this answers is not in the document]
grigasp: works fine

packages/content/src/content/extensions/IModelFieldsProvider.ts
```diff
@@ -0,0 +1,157 @@
...
+ * @public
+ */
+interface FieldsProviderContribution {
+  /** Related properties declarations — paths to navigate and properties to load. */
+  relatedProperties?: RelatedPropertiesDeclaration[];
+  /** Calculated field declarations — ECSQL expressions evaluated in the query. */
+  calculatedFields?: CalculatedFieldDeclaration[];
+  /** Category definitions contributed by this provider, keyed by category ID. */
+  categories?: Record<string, CategoryDefinition>;
+}
```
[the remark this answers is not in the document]
grigasp: works fine

packages/content/src/content/extensions/ExternalFieldsProvider.ts
```diff
@@ -0,0 +1,149 @@
...
+   * after SQL-backed fields are populated.
+   *
+   * Each item contains pre-extracted `inputValues` keyed by the names declared in `inputs`.
+   * Must return an array parallel to `items`, where each element contains
+   * values for exactly the declared field IDs.
+   */
+  getValues(props: {
+    items: Array<{ inputValues: { [K in TInputKeys]: Value } }>;
+  }): Promise<Array<ExternalFieldValueRecord<TOutputFieldIds>>>;
+}
```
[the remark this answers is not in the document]
grigasp: works fine

packages/content/src/content/Content.ts
```diff
@@ -0,0 +1,238 @@
...
+ * cached to disk for reuse — skipping the expensive resolution on subsequent loads.
+ *
+ * Cache invalidation: sources become stale when the iModel schema changes
+ * or provider declarations change.
+ *
+ * @public
+ */
+export async function resolveContentSources(_props: ResolveContentSourcesProps): Promise<ContentSource[]> {
+  throw new Error("Not implemented");
+}
```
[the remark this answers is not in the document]
grigasp: works fine

packages/content/src/content/DistinctValues.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  /** The content targets to query against. */
+  targets: ContentTarget[];
+
+  /** The field to get distinct values for. */
+  field: PropertyField | CalculatedField;
+
+  /** Optional filters (restricts which rows contribute distinct values). */
+  filters?: ContentValueFilter[];
+}
```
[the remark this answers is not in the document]
grigasp: works fine

packages/content/src/presentation-content.ts
```diff
@@ -0,0 +1,29 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+// Core data model
+export type { ContentTarget, ContentSource } from "./content/ContentTarget.js";
+export { ContentDescriptor } from "./content/model/ContentDescriptor.js";
+export type { Field, PropertyField, CalculatedField, ExternalField, RelatedFieldGroup } from "./content/model/Field.js";
+export type { CategoryDefinition } from "./content/model/Category.js";
+export type { ContentItem, ContentValues } from "./content/model/ContentItem.js";
+
```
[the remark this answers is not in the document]
grigasp: `ContentDescriptor` is an interface as well as a namespace. This is fine.

packages/content/src/content/model/ContentDescriptor.ts
```diff
@@ -0,0 +1,98 @@
...
+  ): Field | undefined {
+    for (const field of iterateFields(descriptor)) {
+      if (
+        field.kind === "property" &&
+        field.sourceClassName === sourceClassName &&
+        field.propertyName === propertyName
+      ) {
+        return field;
+      }
+    }
```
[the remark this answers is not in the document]
grigasp: fixed

packages/content/src/content/model/Field.ts
```diff
@@ -0,0 +1,132 @@
...
+ */
+interface BaseField {
+  /**
+   * Stable identity key that uniquely identifies this field across descriptor rebuilds.
+   * Derived from: sourceClassName + propertyAccessPath (for property fields),
+   * or a declared stable name (for calculated/external fields).
+   *
+   * @see {@link ../OPEN_QUESTIONS.md | Open Questions} — encoding strategy for long paths.
+   */
+  identity: string;
```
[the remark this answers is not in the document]
grigasp: removed the `@link`

packages/content/ts-content-api.md
```diff
@@ -0,0 +1,80 @@
+# API Open Questions
+
+## RelatedFieldGroup — keep or flatten?
```
saskliutas, line 3: I like the idea of flattening fields. I think it would simplify things as there is no need to traverse fields recursively when looking for some property.  If needed we could provide an utility for building tree structure out of fields array.

packages/content/ts-content-api.md
```diff
@@ -0,0 +1,80 @@
...
+**Arguments for flattening (single `fields: Field[]` on descriptor):**
+
+- `pathFromTarget` already encodes the same path — consumers can group by it
+- Fields wouldn't exist in two places (direct vs nested), simplifying iteration
+- Simpler descriptor serialization/comparison
+- Less complex descriptor schema
+
+**Decision:** TBD
+
+## Field identity encoding for related property fields
```
saskliutas: I vote for `2.` hashing and keeping the leaf readable (probably could hash leaf as well). If I understand correctly all information included in field identity is already stored on field so for debugging purpose all information could be found (of course less conveniently than looking at identifier).

packages/content/ts-content-api.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Considerations:**
+
+- Merging at descriptor level simplifies consumers but complicates identity (which source class does the identity reference?)
+- Separate fields are simpler but push complexity to every consumer
+- The existing native implementation merges them — do we want to preserve that behavior?
+
+**Decision:** TBD
+
+## Blanket property overrides in `ClassPropertySpec`
```
saskliutas, line 68: I am leaning towards `1.` and having explicit `defaultOverrides?: PropertyOverrides`. For me this approach looks cleaner and does not require some specific knowledge about supported wildcard keys.

packages/content/ts-content-api.md
```diff
@@ -0,0 +1,80 @@
...
+
+- Must be unique within a descriptor (same property via different paths → different identities)
+- Must be deterministic for the same descriptor inputs (so rebuilding a descriptor matches already-loaded values within the same session)
+- Used as keys in `ContentValues.values` map (client-side only, not in SQL)
+- Cross-session stability is **not required** — identities are not persisted to storage.
+- This means index-based or order-dependent schemes are viable as long as the ordering algorithm is deterministic for the same inputs.
+
+**Decision:** TBD
+
+## Multi-source property fields
```
saskliutas, line 48: Does `sourceClassName` has to be concrete class and match content target? Could use base class where the property is actually defined? So for example if we have targets: `BisCore.PhysicalElement` and `BisCore.SpatialElement` for property `UserLabel` field could have `BisCore.Element`?

grigasp: In the case you described, no property/field merging is necessary - it's just a single-property field. This section is about a situation where:
- There's a class A with property `FlowRate` (display label = "Flow rate")
- There's a class B with property `FLOW_RATE` (display label = "Flow rate")
- The two properties are of the same type, koq, etc. But they're defined on different classes and have different names.
- User requests content for A + B instances.

saskliutas: For this case I would prefer not to merge and leave it up to UI. I would like for content to match the underlying data. This should simplify implementation on our end and maybe would shows problems with the underlying data for fixing in the future.

packages/content/ts-content-api.md
MartynasStrazdas: Agree flat structure would allow for much simpler transformations in `DescriptorTransformer`.

packages/content/ts-content-api.md
MartynasStrazdas: Agree just let it be handled on UI side if there is a for some consumers need it to be handled here, we can always add an optional 3 option.

packages/content/src/content/Content.ts
```diff
@@ -0,0 +1,238 @@
...
+  | "less-than"
+  | "less-than-or-equal"
+  | "greater-than"
+  | "greater-than-or-equal"
+  | "like"
+  | "is-in";
+
+/**
+ * Request options passed alongside the descriptor when loading values.
+ * Controls _how_ to query, not _what fields exist_.
```
MartynasStrazdas, line 95: Weird comment are `_` supposed to be there?

packages/shared/src/shared/Metadata.ts
```diff
@@ -338,6 +340,96 @@ export namespace EC {
...
+interface PrimitiveValueDescriptor {
+  kind: "primitive";
+  primitiveType: PrimitiveValueType;
+  /**
+   * Full name of the KindOfQuantity associated with this property (e.g., `"Units.LENGTH"`).
+   * Only meaningful for numeric primitive types (`double`, `int`, `long`).
+   * Determines how the value should be formatted and which units to display.
+   */
+  kindOfQuantity?: string;
+}
```
JonasDov: What do you think about changing to something like:

```suggestion
type NumericPrimitiveValueType = Extract<PrimitiveValueType, "Double" | "Int" | "Long">;
type PrimitiveValueDescriptor = {
  kind: "primitive";
} & (
  {
    primitiveType: Exclude<PrimitiveValueType, NumericPrimitiveValueType>;
    kindOfQuantity?: undefined;
  }
  |
  {
    primitiveType: NumericPrimitiveValueType;
    /**
     * Full name of the KindOfQuantity associated with this property (e.g., `"Units.LENGTH"`).
     * Only meaningful for numeric primitive types (`double`, `int`, `long`).
     * Determines how the value should be formatted and which units to display.
     */
    kindOfQuantity?: string;
  }
);
```

With the current approach users can do this:
```ts
const d: PrimitiveValueDescriptor = { kind: "primitive", primitiveType: "Date", kindOfQuantity: "Units.LENGTH" };
```

packages/shared/src/shared/Metadata.ts
```diff
@@ -338,6 +340,96 @@ export namespace EC {
...
+interface StructValueDescriptor {
+  kind: "struct";
+  members: StructMember[];
+}
+
+/** @public */
+interface StructMember {
+  name: string;
+  label: string;
+  type: ValueDescriptor;
```
MartynasStrazdas, line 419: Would suggest renaming `type` to `descriptor`, because `ValueDescriptor` defines more than just `kind/type`

packages/shared/src/shared/Values.ts
JonasDov: Could we not use hard coded strings here and reuse the type deinfed in `Metadata.ts`?
e.g.
```ts
{ value: number; type: "Integer" | "Long" }
```
would be:
```ts
{ value: number; type:  Extract<PrimitiveValueType, "Integer" | "Long"> }
```

packages/shared/src/shared/Metadata.ts
```diff
@@ -338,6 +340,96 @@ export namespace EC {
  */
 export type PrimitiveValueType = "Id" | Exclude<EC.PrimitiveType, "Binary" | "IGeometry">;
```
JonasDov, line 362: A comment why `Binary` and `IGeometry` are excluded would be helpful. Also, explaining why `Id` was added here would be helpful.

packages/shared/src/shared/Metadata.ts
```diff
@@ -338,6 +340,96 @@ export namespace EC {
...
+ * - `ArrayValueDescriptor`: an ordered collection of a single element type.
+ *
+ * @public
+ */
+export type ValueDescriptor = PrimitiveValueDescriptor | StructValueDescriptor | ArrayValueDescriptor;
+
+/** @public */
+interface PrimitiveValueDescriptor {
+  kind: "primitive";
+  primitiveType: PrimitiveValueType;
```
JonasDov: To me it is a little confusing that `PrimitiveProperty` also has `primitiveType` property, which has a different type than `primitiveType` in this interface. Should this have a different name?

packages/shared/src/shared/Metadata.ts
JonasDov, line 419: Could you add comments explaining what each of these describe?:
E.g.
```ts
/** Describes `Value` type of `StructValue` property. */
type: ValueDescriptor;
```

packages/content/src/content/Content.ts
MartynasStrazdas: Did not know these were symbols for cursive resolving.

grigasp: Yes, see https://github.com/adam-p/markdown-here/wiki/markdown-cheatsheet#emphasis

packages/shared/src/shared/Metadata.ts
grigasp: Not sure. `EC.PrimitiveProperty` is a different thing than `PrimitiveValueDescriptor`. The former is an `EC` type (what we get from schema) and the latter is "our" type. Yes, they are similar, but ours is tailored for our use cases (no binary or geometry properties, less "noise"). However, the meaning of `primitiveType` between them is the same, even if the type is slightly different. Interested in what others on @iTwin/itwinjs-core-presentation think.

packages/shared/src/shared/Metadata.ts
grigasp: changed

packages/shared/src/shared/Values.ts
grigasp: https://github.com/iTwin/presentation/pull/1361

packages/shared/src/shared/Metadata.ts
grigasp: https://github.com/iTwin/presentation/pull/1361

packages/shared/src/shared/Metadata.ts
grigasp: I think that would make this attribute less discoverable to consumers, as `type` is very clear. IMO, a "type" doesn't necessary have to be described in one word, it can be described using a structure, too. In this case we call that structure `ValueDescriptor`.

Interested in others' opinion. @iTwin/itwinjs-core-presentation

JonasDov: I think `type` is clearer.

saskliutas: For me `type` fits better this use case. `descriptor` would be confusing at least for me.

packages/shared/src/shared/Metadata.ts
saskliutas: I think `PrimitiveValueDescriptor` could have `type` instead of `primitiveType`? It should be clear from `PrimitiveValueDescriptor` interface name that type is for primitive value?

packages/content/src/content/model/Category.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @public
+ */
+export interface CategoryDefinition {
+  /** Stable identity for deduplication and cross-provider sharing. */
+  id: string;
+  /** Display label shown to the user. */
+  label: string;
+  /** Optional parent category ID, forming a tree. */
+  parentId?: string;
```
JonasDov, line 23: Maybe change to `parentCategoryId`?

packages/content/src/content/ContentTarget.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+export type CardinalityHint = "one" | "many";
+
+/**
+ * The starting point for a content request, scoped to a single EC class.
+ *
+ * Answers the question: "what am I getting properties for?"
+ *
+ * When the consumer selects instances of multiple different classes,
+ * this becomes multiple content targets — one per distinct class.
```
JonasDov, line 26: Should this comment be here? Shouldn't it be moved to the place where `contentTarget` should be provided?

packages/content/src/content/ContentTarget.ts
```diff
@@ -0,0 +1,95 @@
...
+ * so Stage 2 can look up the declaration's property specs and cardinality hint
+ * by re-calling the provider.
+ *
+ * @public
+ */
+interface ResolvedDeclarationGroup {
+  /** ID of the provider that contributed the originating declaration. */
+  providerId: BaseFieldsProvider["id"];
+
+  /** 0-based index into the provider's `relatedProperties` array. */
```
JonasDov: It seems that only iModelFieldsProvider has this `relatedProperties`. Should this be updated?

packages/content/src/content/extensions/IModelFieldsProvider.ts
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @public
+ */
+interface FieldsProviderContribution {
+  /** Related properties declarations — paths to navigate and properties to load. */
+  relatedProperties?: RelatedPropertiesDeclaration[];
+  /** Calculated field declarations — ECSQL expressions evaluated in the query. */
+  calculatedFields?: CalculatedFieldDeclaration[];
+  /** Category definitions contributed by this provider, keyed by category ID. */
+  categories?: Record<string, CategoryDefinition>;
```
JonasDov: It is not clear what the `string` in `Record<string, CategoryDefinition>` means

packages/content/src/content/extensions/IModelFieldsProvider.ts
```diff
@@ -0,0 +1,157 @@
...
+ *       relatedProperties: [{ path: [...] }],
+ *       calculatedFields: [{ id: "calc1", label: "Calc", expression: "...", type: { kind: "primitive", primitiveType: "double" } }],
+ *     };
+ *   },
+ * });
+ * ```
+ *
+ * @public
+ */
+export function defineIModelFieldsProvider(provider: IModelFieldsProvider): IModelFieldsProvider {
```
JonasDov, line 155: This function is strange: it receives provider as a prop and returns it. Why would consumers use it if they can can just do:
```ts
  const myProvider: IModelFieldsProvider = {
    id: "my-domain_v1",
    async getContribution({ target }) {
      if (!target.primaryClass.startsWith("MySchema")) {
        return undefined;
      }
      return {
        relatedProperties: [{ path: [...] }],
        calculatedFields: [{ id: "calc1", label: "Calc", expression: "...", type: { kind: "primitive", primitiveType: "double" } }],
      };
    },
  };
  ```

File: packages/content/src/content/extensions/DescriptorTransformer.ts [hunk 27]
```diff
@@ -0,0 +1,60 @@
...
+ * Use cases:
+ * - Hiding specific fields based on user preferences or component needs.
+ * - Overriding field labels, categories, priorities.
+ * - Cross-provider decisions (e.g., "move all BisCore fields to a System category").
+ *
+ * Rules:
+ * - Transformers may hide, remove, or modify field metadata.
+ * - Transformers must NOT change field identity (the stable key).
+ * - Transformers must NOT add new fields (that's the provider's responsibility).
+ * - Transformers must NOT reorder fields (display order is a UI concern).
```
JonasDov, line 34: These should be enforced by typescript types. Maybe define a separate type like `TransformableContentDescriptor` which makes some properties readonly. Simple example: [typescriptlang](https://www.typescriptlang.org/play/?strictBuiltinIteratorReturn=true#code/JYOwLgpgTgZghgYwgAgCoQM5mQbwLABQyyAtgJ4CCUUAXMgEoRwAmA9iADaXVxkA8AeQBGAKwB8AbkIBfQqEixEKYSNyFiUJm05lkwZnSxRQAcylFSrZsBhk4QjhABqcDobDGQZmYQTssyJBYdOgBALxqFuRUtMgA2jh6BsgARKDAYMCuKQA0lta29o4ubqnpmdnSALo+BIRBYAB00dRxAAxVjfrIESkgEADuKfWYTS1Q7Z0kVjZ2Ds6uPan9QyNYzdxQjQAOAK4YABYAFIn6dCm5+bNFC6UX0gCUhEA) Of what i mean

packages/content/src/content/extensions/ExternalFieldsProvider.ts
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Field declarations — the fields this provider will populate.
+   * Each field ID must appear exactly once.
+   */
+  fields: { [K in keyof TOutputFieldIds]: ExternalFieldDeclaration<TOutputFieldIds[K]> };
+
+  /**
+   * Category definitions used by this provider's fields, keyed by category ID.
+   */
+  categories?: Record<string, CategoryDefinition>;
```
JonasDov: Similar to other `Record` comments, could there be a better type than `string` here which would in itself explain what it is?

packages/content/src/content/extensions/IModelFieldsProvider.ts
JonasDov: Same applies to `defineDescriptorTransformer`, `defineQueryFilterer` and `defineExternalFieldsProvider`

packages/content/ts-content-api.md
JonasDov: Agree

packages/content/ts-content-api.md
JonasDov: Agree

packages/content/src/content/extensions/IModelFieldsProvider.ts
grigasp: It's there to keep consistency with `defineExternalFieldsProvider`, which is there to avoid having to specify type arguments. There we have the following example:
```ts
const iotProvider = defineExternalFieldsProvider({
  id: "iot-sensors_v1",
  fields: [
    { id: "currentFlow", label: "Current Flow", type: { kind: "primitive", primitiveType: "double" } },
    { id: "lastMaintenance", label: "Last Maintenance", type: { kind: "primitive", primitiveType: "dateTime" } },
  ],
  inputs: {
    serialNo: { className: "MySchema:Pump", propertyName: "SerialNumber" },
    deviceId: { className: "MySchema:Device", propertyName: "DeviceId", path: [{ relationship: "MySchema:PumpHasDevice", direction: "forward", target: "MySchema:Device" }] },
  },
  async getValues({ items }) {
    const serials = items.map((item) => item.inputValues.serialNo);
    const data = await fetchFromIoTService(serials);
    return items.map((_, i) => ({
      "currentFlow": data[i].flow,
      "lastMaintenance": data[i].lastMaintenance,
    }));
  },
});
```

Without the `defineExternalFieldsProvider` function, we'd need to specify the type like this:
```ts
const iotProvider:ExternalFieldsProvider<"serialNo" | "deviceId",  ["currentFlow", "lastMaintenance"]> = {
  // ...
};
```

packages/content/src/content/extensions/IModelFieldsProvider.ts
grigasp: > <...> keyed by category ID.

🤔

packages/content/src/content/extensions/ExternalFieldsProvider.ts
grigasp: I think the best we could do is `type CategoryId = string`, but I just tried & see that intellisense still shows it as `string`, so it only affects if you look directly at code. And there we have a tsdoc saying "keyed by category ID", which, IMO, is clear enough.

packages/content/src/content/ContentTarget.ts
grigasp: I think it adds some clarity, explaining that we need one `ContentTarget` per input class, so consumers don't try how to include multiple input classes within one `ContentTarget`. At that point, I would expect consumers to look at the API that "takes" `ContentTarget` to see how to pass multiple of them, which is the `resolveContentSources`, which takes `targets: ContentTarget[]`.

packages/shared/src/shared/Metadata.ts
grigasp: changed to `type`

packages/shared/src/shared/Metadata.ts
grigasp: added

packages/content/src/content/ContentTarget.ts
grigasp: updated the doc

packages/shared/src/presentation-shared.ts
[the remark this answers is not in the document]
grigasp: added the changelog

packages/content/src/content/model/Category.ts
grigasp: I don't think the extra "category" in the name is need - if the attribute name doesn't tell the type of parent, then I think it can be assumed that the type is the same as the object owning this attribute. And we also have that explained in doc.

packages/content/ts-content-api.md
JonasDov: Agree

packages/content/src/content/extensions/DescriptorTransformer.ts
grigasp: Good idea, implemented some of this through your suggested `TransformableDescriptor`. Sadly, we can't enforce add/reorder of fields, since we have to allow removals.

packages/content/ts-content-api.md
JonasDov: agree

packages/content/src/content/extensions/IModelFieldsProvider.ts
JonasDov: What do you think about changing `Record<string, CategoryDefinition>` -> `Record<CategoryDefinition["id"], CategoryDefinition>`

packages/content/src/content/extensions/DescriptorTransformer.ts
```diff
@@ -0,0 +1,98 @@
...
+  return transformer;
+}
+
+/**
+ * A field with its identity made readonly — transformers may modify metadata
+ * (label, categoryId, hidden, readOnly) but must not change identity.
+ *
+ * @public
+ */
+type TransformableField = Field & { readonly identity: string };
```
JonasDov:
```suggestion
type TransformableField = Omit<Field, "identity"> & { readonly identity: string };
```

packages/content/src/content/extensions/DescriptorTransformer.ts
JonasDov: Well, it could be enforced: e.g. by making the arrays read-only and adding some sort of callback which allows removals (or maybe modify the type to allow something like `delete descriptor.directFields[idx]`. But not sure if that's worth it: we'd have to internally convert `ContentDescriptor` to `TransformableDescriptor` (now it can be passed in place). Also we'd have to document how to remove the values. What do you think?

JonasDov: Realized that we don't need to define a custom way to delete. Just define the type like so:
```ts
type DeletableReadonlyArray<T> = ReadonlyArray<T> & {
  splice(start: number, deleteCount?: number): T[];
};
interface TransformableDescriptor {
  readonly sources: readonly ContentSource[];
  directFields: DeletableReadonlyArray<TransformableField>;
  relatedFieldGroups: DeletableReadonlyArray<TransformableRelatedFieldGroup>;
  categories: Record<string, CategoryDefinition>;
}
interface TransformableRelatedFieldGroup extends Omit<RelatedFieldGroup, "fields" | "nestedGroups"> {
  fields:  DeletableReadonlyArray<TransformableField>;
  nestedGroups?: DeletableReadonlyArray<TransformableRelatedFieldGroup>;
}
```
When defined like that, I think that `ContentDescriptor` should satisfy `TransformableDescriptor`. And users can delete values by calling `descriptor.splice(indexToDelete, 1)`

grigasp: yep, I think that works, have a look

packages/content/src/content/extensions/IModelFieldsProvider.ts
grigasp: changed

packages/content/src/content/extensions/DescriptorTransformer.ts
JonasDov: Did you mark `deleteCount` as non optional on purpose?

grigasp: Kind of intentional based on my assumption that `undefined` = "don't remove". Realized that `undefined` = "remove from start till the end", so changed to optional.

packages/content/ts-content-api.md
grigasp: I'm worried about hash collisions. While improbable, it's still possible, so we'll need some logic to handle them, which will complicate things... I'm leaning towards indexing.. E.g. a format like `F{field_index}.{property_name}`.

packages/content/ts-content-api.md
grigasp: Updated the doc and added the `defaultOverrides` attribute

packages/content/ts-content-api.md
grigasp: Updated the doc.

packages/content/ts-content-api.md
grigasp: Made the change. 

On situation that worried me was segmented paths. Example:
```
RelatedFieldGroup (path: A -> B -> C)
  RelatedFieldGroup (path: C -> D)
    Field (combined path: A -> B -> C -> D)
```
With the above structure, we want Categories to be created for each `RelatedFieldGroup`, e.g.:
```
Category C
  Category D
    Field
```
Without `RelatedFieldGroup` and with just the `Field.pathFromTarget` attribute, it seems impossible to achieve. However, I realized that it's actually possible from `IModelFieldsProvider`'s perspective - it knows the split points and it can assign (possibly hierarchical) categories to properties. So for the above situation it would create categories C -> D, and assign D to each field. Added some utilities to help with that.

packages/content/ts-content-api.md
saskliutas: I guess to avoid any complications coming with hashing we can start with index. If index won't be sufficient we can thing about introducing hash later.

grigasp: I'll spend some more time on this... It would be very nice if we could generate stable ids purely from path+property. The recent change to have all fields in a flat list + having stable unique ids would allow us to very quickly find them.

grigasp: Ok, so here's my latest take on this topic...

I think my initial worry about long identities was unfounded. E.g. identities of 50k fields with 4 step paths stringified into the identity would result in ~20MB, and this is like the worst case scenario. In more realistic scenario like 1k fields with average 2 step paths, we'd be looking at ~260 kB of memory. So I suggest going with the format `{property class name}.{property name}({full serialized path form content target to property class})`.

packages/content/src/content/model/ContentDescriptor.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface ContentDescriptor {
+  /** The content sources used to compute this descriptor (one per target class). */
+  sources: ContentSource[];
+
+  /**
+   * All fields in this descriptor — property fields, calculated fields, and external fields.
+   * Related fields carry a non-empty `pathFromTarget` indicating the relationship path
+   * from the target class to the field's source class.
+   */
+  fields: Field[];
```
grigasp: I'm thinking about changing this to `Record<Field["identity"], Field>`. This would:
1. enforce identity uniqueness at the type level,
2. allow O(1) lookups by identity.

The downsides:
- Iterating / mapping isn't as clean as on array: `Object.values(descriptor.fields)`.
- Fields' order is not preserved. But I think that's a UI concern anyway.

Anyone have any concerns or objections?

grigasp: Benchmark: https://jsbm.dev/MVL5uukN9ELYJ

image

saskliutas: Looks good

packages/content/ts-content-api.md
saskliutas: Sounds good

grigasp: updated the doc

packages/content/src/content/model/ContentDescriptor.ts
grigasp: Made the change.

packages/content/src/content/model/Category.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/** @public */
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace CategoryDefinition {
+  /**
+   * Computes a deterministic category ID from a relationship path.
+   * Two providers using the same path will produce the same ID, allowing their fields
+   * to merge under one category.
+   */
+  export function computeIdFromRelationshipPath(path: RelationshipPath): string {
```
JonasDov: What do you think about changing the return type to `CategoryDefinition["id"]`?

```suggestion
  export function computeIdFromRelationshipPath(path: RelationshipPath): CategoryDefinition["id"] {
```

Same for `PropertyField.computeId`.

packages/content/src/content/model/Category.ts
```diff
@@ -0,0 +1,66 @@
...
+   * Creates a `CategoryDefinition` for a related field path.
+   * The `id` is deterministically derived from `path`, so multiple providers
+   * producing fields via the same path will share the same category.
+   */
+  export function createFromRelationshipPath(props: {
+    parentCategoryId?: string;
+    path: RelationshipPath;
+    label: string;
+    description?: string;
+  }): CategoryDefinition {
```
JonasDov: What do you think about changing props type to:

```suggestion
  export function createFromRelationshipPath(props: {
    parentCategoryId?: CategoryDefinition["parentId"];
    path: RelationshipPath;
    label: CategoryDefinition["label"];
    description?: CategoryDefinition["description"];
  }): CategoryDefinition {
```

packages/content/src/presentation-content.ts
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+// Core data model
+export type { ContentTarget, ContentSource } from "./content/ContentTarget.js";
+export { ContentDescriptor } from "./content/model/ContentDescriptor.js";
```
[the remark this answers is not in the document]
grigasp: not an issue

packages/content/src/content/model/Category.ts
JonasDov: What do you think about changing function name to `computeId`. Do we really need to indicate the props from which the id is computed? 
I would see a purpose of this name if there were multiple functions which computed the id (based on different props), but since there is only one, Id go with `computeId` rename.

packages/shared/src/shared/Metadata.ts
```diff
@@ -361,6 +361,74 @@ export namespace EC {
  */
 export type PrimitiveValueType = "Id" | Exclude<EC.PrimitiveType, "Binary" | "IGeometry">;
 
+/** @public */
+type NumericPrimitiveValueType = Extract<PrimitiveValueType, "Double" | "Int" | "Long">;
```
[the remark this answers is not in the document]
grigasp: fixed

packages/content/src/content/model/ContentDescriptor.ts
```diff
@@ -0,0 +1,37 @@
...
+  /** The content sources used to compute this descriptor (one per target class). */
+  sources: ContentSource[];
+
+  /**
+   * All fields in this descriptor — property fields, calculated fields, and external fields,
+   * keyed by field ID.
+   * Related fields carry a non-empty `pathFromTarget` indicating the relationship path
+   * from the target class to the field's source class.
+   */
+  fields: Record<Field["id"], Field>;
```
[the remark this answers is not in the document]
grigasp: fixed

packages/content/src/content/extensions/IModelFieldsProvider.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  /** Hint about expected cardinality (affects loading strategy). */
+  cardinalityHint?: CardinalityHint;
+
+  /**
+   * Optional custom resolution callback. When provided, the system delegates
+   * path resolution to this callback instead of using default discovery.
+   *
+   * The callback receives the iModel accessor and target, and returns concrete paths.
+   * The declaration's `properties` are carried forward to each returned path.
```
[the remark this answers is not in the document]
grigasp: update the doc

packages/content/src/content/model/Category.ts
```diff
@@ -0,0 +1,66 @@
...
+  export function computeIdFromRelationshipPath(path: RelationshipPath): string {
+    let identifier = "";
+    for (const step of path) {
+      const separator = step.relationshipReverse ? "<-" : "->";
+      if (identifier.length === 0) {
+        identifier = step.sourceClassName;
+      }
+      identifier += `${separator}${step.relationshipName}${separator}${step.targetClassName}`;
+    }
+    return identifier;
```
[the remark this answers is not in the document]
grigasp: fixed

packages/content/ts-content-api.md
```diff
@@ -0,0 +1,84 @@
...
+- Ready-made tree for hierarchical UI rendering
+
+**Arguments for flattening (single `fields: Field[]` on descriptor):**
+
+- `pathFromTarget` already encodes the same path — consumers can group by it
+- Fields wouldn't exist in two places (direct vs nested), simplifying iteration
+- Simpler descriptor serialization/comparison
+- Less complex descriptor schema
+
+**Decision:** Flatten. The descriptor holds a single `fields: Field[]` array; `RelatedFieldGroup` is removed.
```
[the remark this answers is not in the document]
grigasp: updated the docs based on recent chnages

packages/content/src/content/model/Category.ts
grigasp: changed

packages/content/src/content/model/Category.ts
grigasp: Changed to `props: { path: RelationshipPath } & Pick<CategoryDefinition, "label" | "description" | "parentId">`. Also, renamed the function to `create`

packages/content/src/content/model/Category.ts
grigasp: Renamed & changed the argument to props object to make it easier to create overloads in the future, if needed.

JonasDov: I don't see changes.

grigasp: I guess next time I should push before replying 🤣

JonasDov (approved):

PR #1373, opened 2026-05-27
[components] Cleanup deprecations

saskliutas:
Part of https://github.com/iTwin/presentation/issues/1364

Removed deprecated APIs.
Removed code paths that handles older itwinjs-core versions.


packages/components/src/presentation-components/common/RulesetRegistrationHelper.ts
```diff
@@ -40,6 +40,7 @@ export class RulesetRegistrationHelper implements Disposable {
   }
 
   /** @deprecated in 5.7. Use `[Symbol.dispose]` instead. */
```
JonasDov, line 42: How did you decide which deprecations to keep vs which ones to remove?

JonasDov (approved):

packages/components/src/presentation-components/common/RulesetRegistrationHelper.ts
saskliutas: I removed deprecations from 3.x and 4.x versions.

packages/components/src/presentation-components/propertygrid/DataProvider.ts
JonasDov, line 176: > I removed deprecations from 3.x and 4.x versions.

Should this be removed?

JonasDov, line 195: > I removed deprecations from 3.x and 4.x versions.

Should this be removed?

grigasp (approved):

packages/components/src/presentation-components/propertygrid/DataProvider.ts
saskliutas: It cannot be removed because of this https://github.com/iTwin/presentation/pull/1370

saskliutas: Same as https://github.com/iTwin/presentation/pull/1373#discussion_r3317080284

PR #1377, opened 2026-05-28
`createECSqlQueryExecutor`: queries are now automatically cancelled when consumers break out of iteration

grigasp:
Query cancellation will be needed for https://github.com/iTwin/presentation/issues/1376, where we'll use POC-proved approach to send several variations of the same query, wait for the first one to complete and cancel the others.


packages/core-interop/src/test/QueryExecutor.test.ts
```diff
@@ -201,25 +201,129 @@ describe("createECSqlQueryExecutor", () => {
 
       expect(resultRows).toEqual(rows);
     });
+
+    describe("return()", () => {
```
JonasDov, line 205: > send several variations of the same query, wait for the first one to complete and cancel the others.

Could you add a test that checks this use case specifically?

grigasp: not sure what you mean 🤔 we test cancellation behavior that's added in this PR. the described use case is not there yet.

JonasDov: On a second look, you're right, the cancellation behaviour is already tested here. A test for the racing use case would be better placed in the implementing class, not here. Resolving

JonasDov (approved):

PR #1387, opened 2026-06-02
support react 19 version

### LINKED ISSUE #803: Update packages to React v19

MartynasStrazdas:


scripts/regression/setupRegressionTests.js
```diff
@@ -173,3 +182,28 @@ function applyGitPatch(patchFile) {
...
+
+  // Pin exact versions to prevent dual-instance with resolutionMode: lowest-direct
+  workspace.overrides["react"] = "18.3.1";
+  workspace.overrides["react-dom"] = "18.3.1";
+  workspace.overrides["@types/react"] = "18.3.1";
+  workspace.overrides["@types/react-dom"] = "18.3.1";
+
+  // Remove peerDependencyRules for react 19 since we're downgrading to 18
+  delete workspace.peerDependencyRules;
+
```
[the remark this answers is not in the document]
MartynasStrazdas: If we ever need it to work differently we will change it then

apps/full-stack-tests/src/hierarchies-react/learning-snippets/Localization.test.tsx
```diff
@@ -145,8 +145,8 @@ describe("Hierarchies React", () => {
         }
         // __PUBLISH_EXTRACT_END__
 
-        const { getAllByRole, getByRole, user } = render(<MyTreeComponent imodelAccess={access} />);
-        const modelB = await waitFor(() => getByRole("treeitem", { name: "My Model B" }));
+        const { getAllByRole, getByTitle, user } = render(<MyTreeComponent imodelAccess={access} />);
```
saskliutas: Why this change was needed?

scripts/regression/setupRegressionTests.js
```diff
@@ -173,3 +182,28 @@ function applyGitPatch(patchFile) {
     process.exit(1);
   }
 }
+
+function downgradeReactCatalog() {
```
saskliutas: Can we handle this through `pnpmfile`? Check `dependencies.patch` we already patch `pnpmfile` to override versions maybe this coulde be done in the same patch?

scripts/regression/dependencies.patch
```diff
@@ -51,6 +54,13 @@ index dd4c53ca..d1e351e0 100644
 +      }
 +    }
 +  }
++
++  // downgrade react dependencies to 18.3.1
++  for (const [reactPkg, version] of Object.entries(reactOverrides)) {
```
saskliutas: Could we use overrides instead of updating each package? There is hook for configuring pnpm config I guess it should allow to add `overrides` https://pnpm.io/pnpmfile#hooksupdateconfigconfig-config--promiseconfig

apps/full-stack-tests/src/hierarchies-react/learning-snippets/Localization.test.tsx
MartynasStrazdas: So it seems the name "changed" and now includes apply filter, changed this into regex so we wouldn't need the exact string (string with exact: false only disables case sensitivity and doesn't allow substrings).

saskliutas (approved):

apps/full-stack-tests/src/hierarchies-react/learning-snippets/Localization.test.tsx
saskliutas: Based on docs it looks like `exact: false` should match substrings. 😕 ~~https://testing-library.com/docs/queries/about#textmatch~~ https://vitest.dev/api/browser/locators.html#getbyrole

MartynasStrazdas: ~~I think the docs are a bit confusing, it mentions that `name` is by default case-insensitive and searches for a substring (meaning exact is false), but doesn't provide examples of using string for substring search so i am inclined to think string is used for full string search and regex for substring at least it seems to work that way~~

the docs are for browser mode just noticed

PR #1388, opened 2026-06-02
[components] Support multiple versions at the same time

### LINKED ISSUE #1117: Components: Can't use multiple versions of the package
**Repro:**
1. Add an extra dependency of `@itwin/presentation-components` with an alias `@itwin/presentation-components-2` to `presentation-full-stack-tests`.
2. `pnpm install`
3. Add `import "@itwin/presentation-components-2"` to one of the components' full stack test files.
4. `pnpm build`
5. Run the tests
6. Observe the following error:
   ```ts
   Error: PropertyEditorManager.registerEditor error: type 'double:presentation-quantity-editor' already registered to 'QuantityPropertyEditorBase'
        at PropertyEditorManager.registerEditor ({repo}\presentation\node_modules\.pnpm\@itwin+components-react@5.1_fc7cf220f01e741c3ed8bddd99407e65\node_modules\@itwin\components-react\src\components-react\editors\PropertyEditorManager.tsx:140:13)
        at initializePropertyValueEditors ({repo}\presentation\packages\components\src\presentation-components\Initialization.ts:52:25)
        at Presentation.initialize ({repo}\presentation\node_modules\.pnpm\@itwin+presentation-fronten_9b661b016157e7786ce9f7c6315c8551\node_modules\@itwin\presentation-frontend\src\presentation-frontend\Presentation.ts:108:29)
        at async initialize ({repo}\presentation\packages\testing\src\presentation-testing\Helpers.ts:121:3)
        at async initialize (src\IntegrationTests.ts:66:3)
        at async Context.<anonymous> (src\components\properties\PropertyEditors.test.tsx:24:5)
   ```

saskliutas:
Updated built in editors and renderers names to avoid clashes between different versions of presentation-components. Added guid suffix to the name so that editors from different versions of components packages would have different names. This will result in multiple editors being registered if multiple versions are used but each copy of package should map to its renderer.

For InstanceKeyValue added mapping from `SelectableInstance` name to runtime name with guid suffix. This should ensure that presentation rules that are using `SelectableInstance` as renderer name are handled correctly.

Closes https://github.com/iTwin/presentation/issues/1117


saskliutas:
@grigasp This approach I came up with does not break anything and possibly could be made in 5.x minor or patch. Although to be safe I opted to do it in major bump. What do you think?

.changeset/strong-cycles-punch.md
```diff
@@ -0,0 +1,6 @@
+"@itwin/presentation-components": major
+---
+
+Allowed multiple `@itwin/presentation-components` versions to coexist in the same application by giving built-in property editors and renderers unique registration names.
+
+This avoids duplicate registration errors when different package versions are loaded together.
```
grigasp, line 7: Did you confirm we can now use 2 versions of this package in the same app?

.changeset/strong-cycles-punch.md
```diff
@@ -0,0 +1,6 @@
+"@itwin/presentation-components": major
```
grigasp: Is it major? Sounds like not even a minor.

packages/components/src/test/instance-filter-builder/Utils.test.ts
```diff
@@ -48,7 +49,19 @@ describe("createInstanceFilterPropertyInfos", () => {
     });
 
     const input = createInstanceFilterPropertyInfos(descriptor);
-    expect(input).toMatchSnapshot();
+    // replace editor name with a placeholder in snapshot to avoid having to update snapshots every time source code changes.
```
grigasp: I believe the guid will be different on every run, not just when source code changes

```suggestion
    // replace editor name with a placeholder in snapshot to avoid it changing everytime the test is run
```

packages/components/src/test/instance-filter-builder/Utils.test.ts
```diff
@@ -48,7 +49,19 @@ describe("createInstanceFilterPropertyInfos", () => {
...
+    const expected = input.map((info) =>
+      info.propertyDescription.editor?.name === QuantityEditorName
+        ? {
+            ...info,
+            propertyDescription: {
+              ...info.propertyDescription,
+              editor: { name: "presentation-quantity-editor-placeholder" },
```
grigasp:
```suggestion
              editor: { name: "presentation-quantity-editor-{guid}" },
```

grigasp (commented):
Looks good. I think we should merge this into `master`.

.changeset/strong-cycles-punch.md
saskliutas: Tested using steps in https://github.com/iTwin/presentation/issues/1117. No errors were produced.

.changeset/strong-cycles-punch.md
saskliutas: Changed to minor and target to `master`

grigasp (approved):

packages/components/src/presentation-components/common/ContentBuilder.ts
```diff
@@ -98,7 +101,11 @@ export function createPropertyDescriptionFromFieldInfo(info: FieldInfo) {
   };
 
   if (info.renderer) {
-    description.renderer = { name: info.renderer.name };
+    if (info.renderer.name === InstanceKeyValueRendererNameInRules) {
+      description.renderer = { name: InstanceKeyValueRendererName };
+    } else {
+      description.renderer = { name: info.renderer.name };
+    }
```
MartynasStrazdas: Can avoid nested ifs with
`description.renderer = { name: info.renderer.name === InstanceKeyValueRendererNameInRules ? InstanceKeyValueRendererName : info.renderer.name };`

MartynasStrazdas (approved):

packages/components/src/presentation-components/common/ContentBuilder.ts
saskliutas: Updated

Issue #1084, opened 2025-10-16
Components: Replace `ECMetadataProvider` in favor of `ECSchemaContext`


grigasp:
[ECMetadataProvider](https://github.com/iTwin/presentation/blob/247d6a139a30d45c67bd91942be5e9f0529f6740/packages/components/src/presentation-components/instance-filter-builder/ECMetadataProvider.ts#L48) serves the same purpose of `ECSchemaContext` that we can now access through `IModelConnection`. The latter also provides caching, that's shared across all iModel users, so should be preferred.

grigasp:
Released with `@itwin/presentation-components @ 6.0`

PR #1389, opened 2026-06-03
Replace ECMetadataProvider with SchemaView

saskliutas:
Closes https://github.com/iTwin/presentation/issues/1084

Replaced our custom `ECMetadataProvider` with `SchemaView` for class hierarchy look up.


packages/components/src/test/_helpers/Common.ts
```diff
@@ -118,6 +119,53 @@ export function stubVirtualization() {
...
+
+    const derivedClass = classesByFullName.get(classInfo.name);
+    const baseClass = classesByFullName.get(baseClassFullName);
+    if (derivedClass && baseClass) {
+      derivedClass.baseClass = baseClass;
+    }
+  });
+
+  return {
+    findClass: (identifier: string) => classesById.get(identifier) ?? classesByFullName.get(identifier),
```
grigasp: is `findClass` expected to find class by id? why do we need the lookup by id?

packages/components/src/presentation-components/instance-filter-builder/SchemaViewUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+  );
+}
+
+/** @internal */
+export function getClassId(classView: SchemaView.Class | string): string {
+  if (typeof classView === "string") {
+    return classView;
+  }
+  return Id64.fromUint32Pair(classView.ecInstanceId, 0);
+}
```
grigasp, line 27: should use `Id64String` instead of `string` for ids

packages/components/src/test/_helpers/Common.ts
saskliutas: This was copied from content tests. Removed lookup by id.

packages/components/src/presentation-components/instance-filter-builder/SchemaViewUtils.ts
saskliutas: Changed to Id64String

grigasp (approved):

PR #1393, opened 2026-06-08
[Components]: Fix race condition in tree reload handling

MartynasStrazdas:
~~A Flaky test (`detects a change in rule condition`) appeared when we switched to react 19, because of slight differences how react runs effects.~~

~~From my understanding the test fails because setting a variable trigger a reload, which produces a new modelSource React then re-runs the useEffect - removing the old listener and adding a new one. If the next variable change fires in the gap between removal and re-subscription, the event is lost and the test hangs.~~

Underlying issue was race condition in how tree reload is handled. It was possible for change event to be raised between listener is removed and added back. This meant that some events were missed and tree was not reloading.

### DIFF
--- a/packages/components/src/presentation-components/tree/controlled/UseTreeReload.ts
+++ b/packages/components/src/presentation-components/tree/controlled/UseTreeReload.ts
@@ -47,7 +47,8 @@ function useModelSourceUpdateOnBriefcaseUpdate(params: TreeReloadParams): void {
     let subscription: Subscription | undefined;
     const reload = () => {
       const currentModelSource = modelSource?.current;
-      if (!currentModelSource || !dataProviderProps.imodel.isBriefcaseConnection()) {
+      /* v8 ignore if -- @preserve */
+      if (!currentModelSource) {
         return;
       }
       /* v8 ignore next -- @preserve */
@@ -87,6 +88,7 @@ function useModelSourceUpdateOnIModelHierarchyUpdate(params: TreeReloadParams): 
         }
 
         const currentModelSource = modelSource?.current;
+        /* v8 ignore if -- @preserve */
         if (!currentModelSource) {
           return;
         }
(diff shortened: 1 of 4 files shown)


saskliutas:
Looks like we need to wait for tree to actually reload (nodeLoader/modelSource) to change? https://github.com/iTwin/presentation/commit/d2c70822d9a81a3a088edae50e4be3425ae07c78

MartynasStrazdas:
> Looks like we need to wait for tree to actually reload (nodeLoader/modelSource) to change? [d2c7082](https://github.com/iTwin/presentation/commit/d2c70822d9a81a3a088edae50e4be3425ae07c78)

Seems to fix the issue, can you push that commit here?

saskliutas:
> > Looks like we need to wait for tree to actually reload (nodeLoader/modelSource) to change? [d2c7082](https://github.com/iTwin/presentation/commit/d2c70822d9a81a3a088edae50e4be3425ae07c78)
> 
> Seems to fix the issue, can you push that commit here?

Pushed

grigasp:
> A Flaky test appeared when we switched to react 19, because of slight differences how react runs effects.

Can we have more details on this?
- which test is that
- do we have a failure log? E.g. what exactly failed

packages/components/src/presentation-components/tree/controlled/UseTreeReload.ts
```diff
@@ -124,64 +141,82 @@ function useModelSourceUpdateOnRulesetModification(params: TreeReloadParams): vo
...
+          onReload,
+        });
+      });
 
     return () => {
       removeListener();
       subscription?.unsubscribe();
     };
-  }, [dataProviderProps, modelSource, pageSize, ruleset, onReload, renderedItems]);
+  }, [dataProviderProps, modelSource, pageSize, onReload, renderedItems]);
```
MartynasStrazdas, line 180: model source is now a ref right it shouldn't be in dep list

saskliutas: Eslint will complain because it is defined outside this hook and eslint does not have context that it is a ref. There is no problem to include it in deps as its reference won't change.

saskliutas (approved):

PR #1404, opened 2026-06-16
[components]: Version locale file

saskliutas:
Introduces locale file versioning. This should ensure that if multiple versions of presentation-components are used they won't override each others locale file if they are not the same.


packages/components/tsconfig.utils.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./lib/build",
+    "types": ["node"]
+  },
+  "include": ["src/utils", "src/presentation-components/common/LocalizedStrings.ts"]
```
grigasp, line 8: I'd like the sub-dir name in `src` to match the name in `lib`. Maybe rename `src/utils` to `src/build`?

grigasp (approved):

.changeset/components-versioned-localization.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-components": major
```
grigasp: Is this really `major`?

packages/components/tsconfig.utils.json
saskliutas: This is the same approach as in hierarchies-react to avoid `/lib/build/build/` path. https://github.com/iTwin/presentation/pull/1223#discussion_r2863742837

.changeset/components-versioned-localization.md
saskliutas: Changed to `minor`

PR #1405, opened 2026-06-17
components: Add `propertiesMergeMode` setting for union/intersection properties display in `PresentationPropertyDataProvider`

grigasp:
Resolves https://github.com/iTwin/presentation/issues/1403


packages/components/src/presentation-components/propertygrid/PropertiesIntersection.ts
```diff
@@ -0,0 +1,151 @@
...
+  keys: KeySet,
+  fields: Field[],
+): Promise<DescriptorFieldsSelector | undefined> {
+  const keyClassNames = getDistinctClassNames(keys);
+  if (keyClassNames.length <= 1) {
+    return undefined;
+  }
+
+  const schemaView = await imodel.getSchemaView();
+  const keyClasses = keyClassNames.map((name) => schemaView.findClass(name));
```
saskliutas: Should we filter out `undefined` results of `findClass` or even throw/log error? That way conditions downstream could be simplified if keyClasses are typed as `Array<Schema.View>`

packages/components/src/presentation-components/propertygrid/DataProvider.ts
```diff
@@ -117,6 +129,7 @@ export class PresentationPropertyDataProvider extends ContentDataProvider implem
     this._includeFieldsWithCompositeValues = true;
     this._isNestedPropertyCategoryGroupingEnabled = true;
     this._shouldCreateFavoritesCategory = !props.disableFavoritesCategory;
+    this._propertiesMergeMode = "union";
```
saskliutas: Maybe we should allow to control this through constructor during data provider creation?

grigasp: agree, added the prop

packages/components/src/presentation-components/propertygrid/PropertiesIntersection.ts
grigasp: changed

saskliutas (approved):

PR #1416, opened 2026-06-30
[content]: add content item factory

MartynasStrazdas:
Closes https://github.com/iTwin/presentation/issues/1407

### DIFF
--- a/packages/content/src/content/model/ContentItem.ts
+++ b/packages/content/src/content/model/ContentItem.ts
@@ -30,13 +30,9 @@ export interface ContentValues {
  *
  * @public
  */
-export interface ContentItem {
+export interface ContentItem extends Readonly<ContentValues> {
   /** The descriptor that defines the field schema for this item. */
-  descriptor: ContentDescriptor;
-  /** The primary instance this row represents. */
-  primaryKey: InstanceKey;
-  /** Map of field ID → raw value. */
-  values: Record<Field["id"], Value>;
+  readonly descriptor: ContentDescriptor;
 
   /**
    * Retrieve a value by field reference.
--- a/packages/content/src/content/model/ContentItem.ts
+++ b/packages/content/src/content/model/ContentItem.ts
@@ -30,15 +30,21 @@ export interface ContentValues {
  *
  * @public
  */
-export interface ContentItem extends Readonly<ContentValues> {
+export interface ContentItem {
   /** The descriptor that defines the field schema for this item. */
-  readonly descriptor: ContentDescriptor;
+  readonly descriptor: DeepReadonly<ContentDescriptor>;
+
+  /** The primary instance this row represents. */
+  readonly primaryKey: DeepReadonly<InstanceKey>;
+
+  /** Map of field ID → raw value. */
+  readonly values: DeepReadonly<Record<Field["id"], Value>>;
 
   /**
    * Retrieve a value by field reference.
    * Returns `undefined` if the field doesn't apply to this item's class.
    */
-  getValue(field: Field): Value;
+  getValue(field: Field): DeepReadonly<Value>;
 }
 
 /**
@@ -52,7 +58,7 @@ export function createContentItem({
 }: {
   descriptor: ContentDescriptor;
   contentValues: ContentValues;
-}): DeepReadonly<ContentItem> {
+}): ContentItem {
   return {
     descriptor,
     primaryKey: contentValues.primaryKey,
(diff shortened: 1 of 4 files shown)


packages/content/src/content/model/ContentItem.ts
```diff
@@ -39,3 +39,19 @@ export interface ContentItem extends Readonly<ContentValues> {
...
+export function createContentItem(descriptor: ContentDescriptor, contentValues: ContentValues): ContentItem {
+  return Object.freeze({
+    descriptor,
+    primaryKey: contentValues.primaryKey,
+    values: contentValues.values,
+    getValue(field: Field): Value {
+      return contentValues.values[field.id];
+    },
+  });
+}
```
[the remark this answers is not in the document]
MartynasStrazdas: Fixed

packages/content/src/content/model/ContentItem.ts
```diff
@@ -39,3 +39,20 @@ export interface ContentItem extends Readonly<ContentValues> {
...
+export function createContentItem(descriptor: ContentDescriptor, contentValues: ContentValues): ContentItem {
+  const values = Object.freeze({ ...contentValues.values });
+  return Object.freeze({
+    descriptor: Object.freeze({ ...descriptor }),
+    primaryKey: Object.freeze({ ...contentValues.primaryKey }),
+    values,
+    getValue(field: Field): Value {
+      return values[field.id];
+    },
+  });
```
grigasp: Like CoPilot said, `Object.freeze` works in a shallow way. So while your change "froze" members of the `ContentItem`, nested members of them are still modifyable. We need this to be a "deep freeze".

However, I thinking maybe we don't need freeze at all and could instead just return a `DeepReadonly<ContentItem>`? That wouldn't prevent consumers of modifying *if they really wanted to*, but IMO it would be good enough, and it would be way cheaper.

packages/content/src/content/model/ContentItem.ts
```diff
@@ -39,3 +39,20 @@ export interface ContentItem extends Readonly<ContentValues> {
    */
   getValue(field: Field): Value;
 }
+
+/**
+ * Create a `ContentItem` accessor from a descriptor and raw content values.
+ *
+ * @internal
+ */
+export function createContentItem(descriptor: ContentDescriptor, contentValues: ContentValues): ContentItem {
```
JonasDov: Could we make the function accept a single props object?

```suggestion
export function createContentItem({ descriptor, contentValues }: { descriptor: ContentDescriptor; contentValues: ContentValues }): ContentItem {
```

MartynasStrazdas: changed

packages/content/src/content/model/ContentItem.ts
MartynasStrazdas: Changed to a recursive readonly aproach, `DeepReadonlyObject` from core-bentley causes 
image
because descriptor references itself, therefor used a custom DeepReadonly (it is not infinite and goes for 5 levels of depth, but i think that is fine)

grigasp: > because descriptor references itself

really? through which path?

I didn't spent too much time trying to understand what this error really means in this context, but it seems there's no error if we define `ContentItem` like this:
```ts
export interface ContentItem {
  /** The descriptor that defines the field schema for this item. */
  readonly descriptor: DeepReadonlyObject<ContentDescriptor>;

  /** The primary instance this row represents. */
  readonly primaryKey: DeepReadonlyObject<InstanceKey>;
  
  /** Map of field ID → raw value. */
  readonly values: DeepReadonlyObject<Record<Field["id"], Value>>;

  /**
   * Retrieve a value by field reference.
   * Returns `undefined` if the field doesn't apply to this item's class.
   */
  getValue(field: Field): DeepReadonlyObject<Value>;
}
```

And then in the `createContentItem` just regularly return a `ContentItem`...

MartynasStrazdas: Descriptor has fields, fields base has Value Descriptors which when they are struct or array reference ValueDescriptors. The error is on create function trying to `createContentItem` because DeepReadonlyObject gets infinite recursion.

grigasp: Ok, so it's not `Descriptor`, but `ValueDescriptor` to be exact. Any idea why my example works then?

MartynasStrazdas: adjusted the custom deep read only to not need depth

grigasp (approved):

PR #1437, opened 2026-07-09
`content`: Decouple value selection from consumer-facing fields

grigasp:
Prerequisite for https://github.com/iTwin/presentation/issues/1406.

Splits a descriptor `Field`'s two conflated concerns — *what to display* vs. *what column to SELECT* — by introducing a **value selector**: the deduplicated "column to select".

This removes the need to model the same column N times (fields + their `forkField`/override variants) and to keep external-input columns alive via hidden fields + pinning.

## Changes

- **New `model/ValueSelector.ts`**: `ValueSelector = PropertyValueSelector | CalculatedValueSelector` (`@public`, each `Pick`s its column fields from the matching `Field`), plus `@internal` helpers `computePropertySelectorId`, `createPropertySelector`, `createCalculatedSelector`, `collectSelectors`.
- **`ContentDescriptor`**: added `selectors` (keyed by selector id).
- **`Field`**: added required `selectorId` to `PropertyField` + `CalculatedField`; `CalculatedField` gains optional `bindings`.
- **`PropertyFieldMerge`**: stamps `selectorId` (= base id) on merged fields.
- **`DescriptorTransformer`**: `selectorId` is readonly in the transformer view; `forkField` copies it onto the fork. No pinning added.
- **`ExternalFieldsProvider`**: input columns are guaranteed via selectors instead of hidden/pinned fields.

## Identity

- Property selector id = `PropertyField.computeId(...)` **without** a `forkKey`, so all fork/override variants share one selector.
- Calculated selector id = the calculated field id.
- External-input selectors are added unconditionally, so removing an output field never drops an input column — replacing pinning.

### DIFF
--- a/packages/content/src/content/extensions/DescriptorTransformer.ts
+++ b/packages/content/src/content/extensions/DescriptorTransformer.ts
@@ -12,6 +12,7 @@ import type { ContentSource } from "../ContentTarget.js";
 import type { CategoryDefinition } from "../model/Category.js";
 import type { ContentDescriptor } from "../model/ContentDescriptor.js";
 import type { Field } from "../model/Field.js";
+import type { DeepReadonly } from "../model/Utils.js";
 
 /**
  * Default priority for descriptor transformers.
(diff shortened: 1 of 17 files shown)


packages/content/src/presentation-content.ts
```diff
@@ -7,6 +7,7 @@
 export type { ContentTarget, ContentSource } from "./content/ContentTarget.js";
 export type { ContentDescriptor } from "./content/model/ContentDescriptor.js";
 export type { Field, PropertyField, CalculatedField, ExternalField } from "./content/model/Field.js";
+export type { ValueSelector, PropertyValueSelector, CalculatedValueSelector } from "./content/model/ValueSelector.js";
```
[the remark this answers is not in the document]
grigasp: the package is not being released at this point

packages/content/src/content/extensions/DescriptorTransformer.ts
```diff
@@ -86,7 +87,9 @@ export function defineDescriptorTransformer(transformer: DescriptorTransformer):
  * @public
  */
 type TransformableField<TField extends Field = Field> = TField extends Field
-  ? Omit<TField, "id"> & { readonly id: string }
+  ? Omit<TField, "id" | "selectorId"> & { readonly id: string } & (TField extends { selectorId: string }
+        ? { readonly selectorId: string }
+        : {})
   : never;
```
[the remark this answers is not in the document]
grigasp: Agree. Changed `TransformableField` to be fully deep readonly except for the whitelisted attributes that we want to keep editable.

packages/content/src/content/model/ValueSelector.ts
```diff
@@ -0,0 +1,175 @@
...
+ * - one selector per SQL-backed field (`property`/`calculated`), and
+ * - one property selector per external fields provider input.
+ *
+ * External-input selectors are added unconditionally, so removing an output field can never remove
+ * an input column. When a field-backed selector and an input selector share an id, the field-backed
+ * one is kept (they are otherwise identical).
+ *
+ * @internal
+ */
+export function collectSelectors(
```
saskliutas, line 130: How will this function be used? Currently when creating field you need to have `selectorId` but this function creates selectors from fields?

grigasp: Good question. The idea is:
1. Collect fields contributions from fields providers.
2. Using those contributions, create the Fields, automatically assigning them the calculated `selectorId` (to fields that have it). Note: at this point we only have `selectorId`, not the selector itself.
3. Assemble the fields into a (working) descriptor, run it through descriptor transformers, which may fork, remove or modify existing fields.
4. Use `collectSelectors` to get the selectors from the remaining fields.

I think the part that isn't clear and probably doesn't make much sense - we calculate `selectorId` when creating a `Field`, then calculate it again when creating selectors for given `Fields`, which already have the id calculated. Made an improvement there, please have a look.

saskliutas (approved):

packages/content/src/content/model/ValueSelector.ts
saskliutas: Looks better

PR #1444, opened 2026-07-13
`shared`: Added `EC.Property.category` attribute 

grigasp:
Prerequisite for https://github.com/iTwin/presentation/issues/1406.

### DIFF
--- a/.changeset/property-category.md
+++ b/.changeset/property-category.md
@@ -1,0 +1,6 @@
+---
+"@itwin/presentation-shared": major
+"@itwin/presentation-core-interop": minor
+---
+
+`EC.Property`: Added `category` attribute that provides access to the property's `EC.PropertyCategory`.
--- a/.changeset/property-category.md
+++ b/.changeset/property-category.md
@@ -3,4 +3,4 @@
 "@itwin/presentation-core-interop": minor
 ---
 
-`EC.Property`: Added `category` attribute that provides access to the property's `EC.PropertyCategory`.
+`EC.Property`: Added `category` attribute that provides access to the property's `EC.PropertyCategory`. Also added a `description` attribute to `EC.SchemaItem`.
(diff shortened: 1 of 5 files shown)


.changeset/property-category.md
```diff
@@ -0,0 +1,6 @@
+---
+"@itwin/presentation-shared": major
+"@itwin/presentation-core-interop": minor
+---
+
+`EC.Property`: Added `category` attribute that provides access to the property's `EC.PropertyCategory`.
```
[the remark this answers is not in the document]
grigasp: 1. The `major` bump type on `@itwin/presentation-shared` already says it's breaking, no need to repeat it.
2. The change in `@itwin/presentation-core-interop` is NOT a breaking change - I made an addition to an interface that's used **only as a return type** of a public function.

packages/shared/src/shared/Metadata.ts
```diff
@@ -163,6 +163,18 @@ export namespace EC {
...
+  /**
+   * Represents a property category used to group related properties.
+   * @see https://www.itwinjs.org/reference/ecschema-metadata/metadata/propertycategory/
+   * @public
+   */
+  export interface PropertyCategory extends SchemaItem {
+    /** Determines the display order of the category relative to other categories. Higher priority categories are displayed first. */
+    priority: number;
+    /** Optional description of the category. */
+    description?: string;
```
JonasDov: `description` is declared on `SchemaItem` in [@itwin/ecschema-metadata](https://github.com/iTwin/itwinjs-core/blob/15ceb61748e3df0e685a2ddb7ff57b4832fae5ae/core/ecschema-metadata/src/Metadata/SchemaItem.ts#L48) and not on `PropertyCategory` (it's not unique to `PropertyCategory`). Should this live on `EC.SchemaItem` instead of `EC.PropertyCategory`? This way `Class`, `Enumeration`, and `KindOfQuantity` would have it too?

JonasDov (approved):

packages/shared/src/shared/Metadata.ts
grigasp: good catch, moved

PR #1446, opened 2026-07-14
`content`: Implement `ContentProvider.getContentDescriptor`

grigasp:
Closes https://github.com/iTwin/presentation/issues/1406


packages/content/src/content/descriptor-building/RelatedFields.ts
```diff
@@ -0,0 +1,146 @@
...
+      const perPath = await Promise.all(
+        group.paths.map(async ({ path }) =>
+          createFieldsForPath({ imodelAccess, path, properties: declaration.properties }),
+        ),
+      );
+      for (const enumerated of perPath.flat()) {
+        candidates.push({ ...enumerated, provider });
+      }
+    }),
+  );
```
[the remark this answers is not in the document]
grigasp: fixed

packages/content/src/content/descriptor-building/RelatedFields.ts
```diff
@@ -0,0 +1,146 @@
...
+/** Enumerates the target-class and relationship-class fields opted in by a single `StepPropertySpec`. */
+async function createFieldsForStep(props: {
+  imodelAccess: ECSchemaProvider;
+  path: RelationshipPath;
+  stepSpec: StepPropertySpec;
+}): Promise<CategorizedField[]> {
+  const { imodelAccess, path, stepSpec } = props;
+  const step = path[stepSpec.stepIndex];
+  const pathFromTarget = path.slice(0, stepSpec.stepIndex + 1);
+  const enumerated: CategorizedField[] = [];
```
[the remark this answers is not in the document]
grigasp: fixed

packages/content/src/test/InternalUtils.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { describe, expect, it } from "vitest";
+import { collectInParallel } from "../content/InternalUtils.js";
+
+describe("collectInParallel", () => {
+  it("returns an empty array for no items", async () => {
+    expect(await collectInParallel([], async () => [1])).to.deep.equal([]);
+  });
+
+  it("concatenates the per-item arrays preserving input order", async () => {
```
JonasDov, line 16: To make sure that order is preserved, maybe make each callback have a timeout. Where middle entry has the largest timeout. This way it would be confirmed that no mater how long the async functions take, order will be preserved.

packages/content/src/test/InternalUtils.test.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { describe, expect, it } from "vitest";
+import { collectInParallel } from "../content/InternalUtils.js";
+
```
JonasDov, line 10: There are two other functions in the `InternalUtils` file: `stableStringify`, `getClassLabel`. Do they need tests?

packages/content/src/content/CreateContentProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+ * The descriptor is built lazily on the first `getContentDescriptor` call and cached; the
+ * remaining methods are query-stage concerns handled by later pipeline stages.
+ *
+ * @internal
+ */
+export function createContentProviderImpl(props: Props<typeof createContentProvider>): ContentProvider {
+  const { imodelAccess, sources, config } = props;
+  let descriptor: Promise<ContentDescriptor> | undefined;
+  async function getContentDescriptor() {
+    return (descriptor ??= buildContentDescriptor({ imodelAccess, sources, config }));
```
JonasDov: Maybe split into separate lines?:

```suggestion
    descriptor ??= buildContentDescriptor({ imodelAccess, sources, config });
    return descriptor;
```

packages/content/src/content/CreateContentProvider.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * The descriptor is built lazily on the first `getContentDescriptor` call and cached; the
+ * remaining methods are query-stage concerns handled by later pipeline stages.
+ *
+ * @internal
+ */
+export function createContentProviderImpl(props: Props<typeof createContentProvider>): ContentProvider {
+  const { imodelAccess, sources, config } = props;
+  let descriptor: Promise<ContentDescriptor> | undefined;
+  async function getContentDescriptor() {
```
JonasDov, line 23: Any reason to define `getContentDescriptor` here and not directly where it is returned?

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @internal
+ */
+export function createContributionMemoizer(props: { imodelAccess: ContributionIModelAccess }): {
+  getContribution: GetContribution;
+} {
+  const { imodelAccess } = props;
+  const cache = new WeakMap<ContentTarget, Map<IModelFieldsProvider["id"], Promise<FieldsProviderContribution>>>();
+  return {
+    getContribution: async (provider, target) => {
+      let byProvider = cache.get(target);
```
JonasDov: I notice similar pattern in multiple places:
```ts
let entry = map.get(key);
if (!entry) {
  entry = newVal;
  map.set(key, entry);
}
```
Maybe add a helper function e.g. https://github.com/iTwin/viewer-components-react/blob/44e17c1ba2f78f779995da89c0291b608bda01a4/packages/itwin/tree-widget/src/tree-widget-react/components/trees/common/internal/Utils.ts#L317

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
```diff
@@ -0,0 +1,74 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { ContentTarget } from "../ContentTarget.js";
+import type { IModelFieldsProvider } from "../extensions/IModelFieldsProvider.js";
+
+/** The contribution returned by an `IModelFieldsProvider`, with `undefined` meaning "not applicable". */
+type FieldsProviderContribution = Awaited<ReturnType<IModelFieldsProvider["getContribution"]>>;
```
JonasDov: Any reason to make this `Awaited` and then add `Promise<FieldsProviderContribution>`. Maybe change to not be awaited and return without the `Promise` wrap.

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/** The iModel access shape required by `IModelFieldsProvider.getContribution`. */
+type ContributionIModelAccess = Parameters<IModelFieldsProvider["getContribution"]>[0]["imodelAccess"];
```
JonasDov: Do we need to create a separate type for this `IModelAccess`?

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Re-fetches an `IModelFieldsProvider` contribution for a target — the memoized accessor produced by
+ * {@link createContributionMemoizer} and consumed by the Stage 2 field/category/calculated-field
+ * collectors.
+ *
+ * @internal
+ */
+export type GetContribution = (
+  provider: IModelFieldsProvider,
+  target: ContentTarget,
+) => Promise<FieldsProviderContribution>;
```
JonasDov: Could make the function accept props object instead?

packages/content/src/content/model/Utils.ts
```diff
@@ -19,17 +19,25 @@ export type DeepReadonly<T> = T extends (...args: any[]) => any
       ? { readonly [K in keyof T]: DeepReadonly<T[K]> }
       : T;
 
-export function serializeRelationshipPath(path: RelationshipPath): string {
+/**
+ * Serializes a relationship path to a deterministic string. When `omitLastTargetClass` is set, the
+ * final step's target class (and its arrow) is left off, yielding a target-independent serialization
+ * of the path's last relationship.
+ */
+export function serializeRelationshipPath(path: RelationshipPath, options?: { omitLastTargetClass?: boolean }): string {
```
JonasDov: Could make the function accept props object instead?

packages/content/src/content/InternalUtils.ts
```diff
@@ -3,10 +3,54 @@
...
+  const keys = Object.keys(value).sort();
+  return `{${keys.map((k) => `${JSON.stringify(k)}:${stableStringify((value as Record<string, unknown>)[k])}`).join(",")}}`;
+}
+
+/**
+ * Resolves a class's display label (its label, else its name).
+ *
+ * @internal
+ */
+export async function getClassLabel(imodelAccess: ECSchemaProvider, className: EC.FullClassName): Promise<string> {
```
JonasDov: Could make the function accept props object instead?

packages/content/src/content/InternalUtils.ts
```diff
@@ -3,10 +3,54 @@
...
  */
 export const ECSQL_PREFIX = "pres_";
+
+/**
+ * Runs `fn` over every item in parallel and concatenates the resulting arrays into a single
+ * flat array (preserving input order). A concise replacement for `(await Promise.all(...)).flat()`.
+ *
+ * @internal
+ */
+export async function collectInParallel<TItem, TResult>(
```
JonasDov: Could make the function accept props object instead?

JonasDov (commented):
Reviewed smallest files

packages/content/src/content/CreateContentProvider.ts
grigasp: we'll need it in `getItems`

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
grigasp: it's local to module, why not?

packages/content/src/content/InternalUtils.ts
grigasp: changed

packages/content/src/content/InternalUtils.ts
grigasp: changed

packages/content/src/content/model/Utils.ts
grigasp: changed

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
grigasp: changed

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
grigasp: changed

packages/content/src/test/InternalUtils.test.ts
grigasp: added

packages/content/src/test/InternalUtils.test.ts
grigasp: added a test

packages/content/src/content/descriptor-building/ContributionMemoizer.ts
grigasp: added

packages/content/src/test/descriptor-building/ClassPropertyFields.test.ts
```diff
@@ -0,0 +1,304 @@
...
+        anchor: "none",
+      });
+
+      expect(categorization).to.deep.equal({
+        anchor: "none",
+        category: { source: "schema", id: "TestSchema.Geometry", label: "Geometry" },
+      });
+    });
+
+    it("falls back to the schema category's name when it has no label", async () => {
```
saskliutas, line 242: Could we use different label in `reports the EC schema property category` test or different class name is this one because it's confusing as both produce same result.

packages/content/src/test/descriptor-building/ExternalFields.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  it("collects input column coordinates, including relationship paths", () => {
+    const path: RelationshipPath = [
+      { sourceClassName: "TestSchema.A", targetClassName: "TestSchema.B", relationshipName: "TestSchema.AtoB" },
+    ];
+    const provider: ExternalFieldsProvider<"direct" | "related"> = {
+      id: "ext_v1",
+      fields: [],
+      inputs: {
```
saskliutas, line 56: Unrelated question. Does it make sense to have external provider without fields but with inputs? If not maybe `collectExternalFields` should throw?

packages/content/src/test/descriptor-building/Categories.test.ts
```diff
@@ -0,0 +1,616 @@
...
+      sources: [createSource()],
+      imodelFieldsProviders: [],
+      externalFieldsProviders: [external],
+      getContribution,
+      fields: [],
+    });
+    expect(categories).to.deep.equal({});
+  });
+
+  it("applies the default priority to external categories without an explicit priority", async () => {
```
saskliutas, line 217: How does this test check that priority was applied?

packages/content/src/test/descriptor-building/ExternalFields.test.ts
grigasp: Probably not, but I don't think we can ensure that declared inputs will actually be used by `getValues` when resolving field values. The best we could do is have a check for empty `fields` & non-empty `inputs` combination, but I don't think that helps much.

packages/content/src/test/descriptor-building/Categories.test.ts
grigasp: fixed...

packages/content/src/test/descriptor-building/ClassPropertyFields.test.ts
grigasp: renamed the class to `GeometryClass`

packages/content/src/content/descriptor-building/DirectFields.ts
```diff
@@ -0,0 +1,103 @@
...
+ * accepts either notation).
+ */
+async function collectDeclaringClassClosure(
+  imodelAccess: ECSchemaProvider,
+  concreteClassNames: EC.FullClassNameDotNotation[],
+): Promise<Map<EC.FullClassNameDotNotation, Set<EC.FullClassNameDotNotation>>> {
+  const closure = new Map<EC.FullClassNameDotNotation, Set<EC.FullClassNameDotNotation>>();
+  await Promise.all(
+    concreteClassNames.map(async (concreteClassName) => {
+      const pending: EC.Class[] = [await getClass(imodelAccess, concreteClassName)];
```
JonasDov, line 78: `getClass` seems to parse full class name again. Then class name is normalized on line 82 again.
Maybe do this:
- change `collectDeclaringClassClosure` to accept `Array<{schemaName: string; className: string; }>`;
- Update `getClass` to accept props object. And make it accept `schemaName` and `className` in props (can be achieved with overrides if you think that's better).
- `pending`type should be this: `Array<{ class: EC.Class; declaringClassName: EC.FullClassNameDotNotation}>`. Then remove the normalization on line 82.

packages/content/src/content/descriptor-building/DirectFields.ts
```diff
@@ -0,0 +1,103 @@
...
+          key: declaringClassName,
+          createFunc: () => new Set<EC.FullClassNameDotNotation>(),
+        }).add(concreteClassName);
+        const baseClass = await ecClass.baseClass;
+        if (baseClass) {
+          pending.push(baseClass);
+        }
+        if (ecClass.isEntityClass()) {
+          pending.push(...(await ecClass.getMixins()));
+        }
```
JonasDov: Could use `Promise.all`:

```suggestion
        await Promise.all([
          ecClass.baseClass.then((base) => base && pending.push(base)),
          ecClass.isEntityClass() && ecClass.getMixins().then((mixins) => pending.push(...mixins)),
        ]);
```

packages/content/src/content/descriptor-building/DirectFields.ts
grigasp: I'm not sure about switching to `{ schemaName: string; className: string; }` objects, I think that would add complication that I'm willing to add only if we prove there's an actual performance problem.

I'd prefer to do a full switch to `EC.FullClassNameDotNotation` instead, filed a backlog item for that: https://github.com/iTwin/presentation/issues/1453

packages/content/src/content/descriptor-building/DirectFields.ts
grigasp: changed to
```ts
        const [baseClass, mixins] = await Promise.all([
          ecClass.baseClass,
          ecClass.isEntityClass() ? ecClass.getMixins() : undefined,
        ]);
        baseClass && pending.push(baseClass);
        mixins && pending.push(...mixins);
```

packages/content/src/content/model/PropertyValueDescriptor.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+  if (property.isPrimitive()) {
+    const type = toPrimitiveValueType(property.primitiveType);
+    if (type === undefined) {
+      return undefined;
+    }
+    return createPrimitiveValueDescriptor(type, await getKindOfQuantityName(property));
+  }
+  return undefined;
+}
```
JonasDov, line 55: As far as I understand `createScalarValueDescriptor` returns `undefined` for array `EC.Property`. So, `createValueDescriptorFromProperty` will return `undefined` for such properties. 
`createValueDescriptorFromProperty` tests pass, because they set `isPrimitive` to true on all property types. 
Am I missing something?

packages/content/src/content/model/PropertyValueDescriptor.ts
```diff
@@ -0,0 +1,144 @@
...
+    case "Integer":
+      return kindOfQuantity !== undefined ? { kind: "primitive", type, kindOfQuantity } : { kind: "primitive", type };
+    case "Double":
+    case "Long":
+      return kindOfQuantity !== undefined ? { kind: "primitive", type, kindOfQuantity } : { kind: "primitive", type };
+    case "String":
+      return { kind: "primitive", type };
+    default:
+      return { kind: "primitive", type };
+  }
```
JonasDov, line 128: Any reason to have separate lines for `"Integer"` and `"String"`?

```suggestion
  switch (type) {
    case "Integer":
    case "Double":
    case "Long":
      return kindOfQuantity !== undefined ? { kind: "primitive", type, kindOfQuantity } : { kind: "primitive", type };
    default:
      return { kind: "primitive", type };
  }
```

packages/content/src/content/descriptor-building/PropertyFieldMerge.ts
```diff
@@ -0,0 +1,144 @@
...
+ * categorization pass can turn it into a `categoryId`. This is the inverse of `forkField`: it merges
+ * many candidates into one field on the way in, while `forkField` splits one field into a carved
+ * subset on demand.
+ *
+ * @internal
+ */
+export function mergePropertyFieldsByIdentity(candidates: PropertyFieldCandidate[]): CategorizedField[] {
+  const groups = new Map<Field["id"], PropertyFieldCandidate[]>();
+  for (const candidate of candidates) {
+    const baseId = PropertyField.computeId(candidate.field);
```
JonasDov: Do we need to call `PropertyField.computeId` when `candidate.field.id` is already defined?
Seems that this is called only in `buildContentDescriptor`. And it's called with direct and related fields array.  Direct fields are created with `collectDirectPropertyFields` and related are created with `collectRelatedPropertyFields`: both of them seem to call `collectClassPropertyFields` under the hood which creates a new id by calling `PropertyField.computeId` (`forkKey` is not provided).

packages/content/src/content/descriptor-building/BuildDescriptor.ts
```diff
@@ -0,0 +1,110 @@
...
+    collectCategories({
+      imodelAccess,
+      sources,
+      imodelFieldsProviders,
+      externalFieldsProviders,
+      getContribution,
+      fields: mergedPropertyFields,
+    }),
+    collectCalculatedFields({ sources, imodelFieldsProviders, getContribution }),
+  ]);
```
JonasDov, line 80: I am not a fan of iterating `sources` and fields providers so many times.
- For each `source` there is a call to `collectRelatedPropertyFields`. Under the hood it iterates through its `resolvedDeclarations` and gets `imodelFieldsProviders` and then calls `getContribution`.
- For each `source`  we iterate through `imodelFieldsProviders` (`collectCategories` function). For each provider we call `getContribution`.
- For each `source`  we iterate through `imodelFieldsProviders` (`collectCalculatedFields` function). For each provider we call `getContribution` again.

In each of these cases we iterate to get the contributions. Can they be somehow merged so we don't iterate 3 separate times?

packages/content/src/content/descriptor-building/Categories.ts
```diff
@@ -0,0 +1,259 @@
...
+          priority: provider.priority ?? DEFAULT_FIELDS_PROVIDER_PRIORITY,
+        }))
+      : [],
+  );
+  for (const { category, priority } of [...contributed, ...externalContributed]) {
+    const existing = registry.get(category.id);
+    if (!existing || priority > existing.priority) {
+      registry.set(category.id, { category, priority });
+    }
+  }
```
JonasDov, line 85: Seems that we create `contributed` and `externalContributed` arrays just to latter iterate through them and add them to the registry. Why not just add to the registry directly on lines 68 and 74?

packages/content/src/content/descriptor-building/Categories.ts
```diff
@@ -0,0 +1,259 @@
...
+      anchorPaths.set(anchorId, field.pathFromTarget);
+    }
+  }
+
+  // 3. Create the referenced relationship/target categories, then the schema (sub-)categories. Each
+  //    class-based category nests under its nearest existing ancestor in the path's
+  //    `rel_0 → tgt_0 → rel_1 → tgt_1 → …` chain (a step's target nests under its relationship; a
+  //    step nests under the previous step's target).
+  const labelsToResolve: Array<{ category: CategoryDefinition; className: EC.FullClassName }> = [];
+  const registerAnchorCategory = (
```
JonasDov, line 129: Any reason to define this function inside `collectCategories`?

packages/content/src/content/descriptor-building/Categories.ts
```diff
@@ -0,0 +1,259 @@
...
+  }
+  // The category objects registered above are the same references, so resolving labels here updates
+  // them in place.
+  await Promise.all(
+    labelsToResolve.map(async ({ category, className }) => {
+      category.label = await getClassLabel({ imodelAccess, className });
+    }),
+  );
+
+  return Object.fromEntries(Array.from(registry, ([id, { category }]) => [id, category]));
```
JonasDov, line 169: what is the reason to make the `registry` of type `Map` when it has to be converted to `Record`. Can't we define `registry` to be of `Record` type instead? This way there would be no need to convert.

packages/content/src/content/descriptor-building/CalculatedFields.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @internal
+ */
+export async function collectCalculatedFields(props: {
+  sources: ContentSource[];
+  imodelFieldsProviders: IModelFieldsProvider[];
+  getContribution: GetContributionFn;
+}): Promise<Record<Field["id"], CalculatedField>> {
+  const { sources, imodelFieldsProviders, getContribution } = props;
+  const declared = await collectInParallel({
```
JonasDov, line 33: Any reason to create `declared` array and then iterate over it? Can't we instead of returning `CalculatedFieldDeclaration` (which are later iterated) on line 40, immediately add them to the result?

packages/content/src/content/model/PropertyValueDescriptor.ts
grigasp: Array properties return true for `isArray`, as well as one of `isStruct`, `isPrimitive`, `isEnumeration`, `isNavigation`. So `createScalarValueDescriptor` will basically return array item type for array property.

packages/content/src/content/descriptor-building/BuildDescriptor.ts
grigasp: I don't think that's a problem as long as `getContribution` is not expensive, it's not like we'll have hundreds or thousands of fields providers. And `getContribution` call is cheap (ensured by `createContributionMemoizer`).

packages/content/src/content/descriptor-building/Categories.ts
grigasp: I think that's more readable and debuggable.

packages/content/src/content/descriptor-building/Categories.ts
grigasp: I needs access to variables available within `collectCategories`.

packages/content/src/content/descriptor-building/CalculatedFields.ts
grigasp: I think that's more readable and debuggable.

packages/content/src/content/descriptor-building/Categories.ts
grigasp: There's still a need to convert, because registry holds `{ category, priority }` and result needs to be `Record<CategoryDefinition["id"], CategoryDefinition>`

packages/content/src/content/descriptor-building/PropertyFieldMerge.ts
grigasp: Changed to use `candidate.field.id` + added an assertion it actually matches the `computeId` result.

packages/content/src/content/model/PropertyValueDescriptor.ts
grigasp: cleaned up

JonasDov (approved):

packages/content/src/content/descriptor-building/ExternalFields.ts
```diff
@@ -0,0 +1,60 @@
...
+        type: declaration.type,
+        providerId: provider.id,
+      };
+      if (declaration.categoryId !== undefined) {
+        field.categoryId = declaration.categoryId;
+      }
+      fields[id] = field;
+    }
+    if (provider.inputs) {
+      const declarations = Object.values(provider.inputs) as ReadonlyArray<InputPropertyDeclaration>;
```
MartynasStrazdas:
```suggestion
      const declarations: ReadonlyArray<InputPropertyDeclaration> = Object.values(provider.inputs);
```

To avoid casting.

MartynasStrazdas (approved):

PR #1454, opened 2026-07-17
Drop cjs build

saskliutas:
Dropped cjs build. Now all packages delivers only ESM output.


.changeset/five-rooms-remain.md
```diff
@@ -0,0 +1,10 @@
+---
+"@itwin/unified-selection-react": major
+"@itwin/unified-selection": major
+"@itwin/presentation-opentelemetry": major
+"@itwin/presentation-core-interop": major
+"@itwin/presentation-hierarchies": major
+"@itwin/presentation-shared": major
+---
+
+Dropped CommonJS modules support. These packages are now published as ES modules only.
```
grigasp: I think for packages that aren't already in `major` bump mode it would be good to have a separate changefile and explain what this major bump means to them. E.g. something like:

> - If you have package X as a `peerDependency`, you should extend the version range.
> - If you have package X as a `dependency` and expose it through public API, you can safely bump the dependency  version - that will not break your API consumers.

grigasp (approved):

.changeset/five-rooms-remain.md
saskliutas: Updated

PR #1459, opened 2026-07-21
[content] Resolve concrete step relationship classes

saskliutas:
Closes https://github.com/iTwin/presentation/issues/1442

Updated `createRelationshipPathJoinInfo` to return `RelationshipPathStepJoinInfo` per step. This info contains selectors for source, target and relationship class ids. They can be used in ECSQL to query concrete classes for particular step.
Updated content source resolved to resolve concreate relationship classes for each step when resolving relationship path.


packages/shared/src/test/ecsql-snippets/ECSqlJoinSnippets.test.ts
```diff
@@ -814,6 +829,91 @@ describe("createRelationshipPathJoinClause", () => {
       expect(result.bindings).toBeUndefined();
     });
 
+    describe("sourceClassIdSelector and targetClassIdSelector", () => {
```
grigasp: Do you think it's worth having these as separate tests? Should we, maybe, just add extra assertions to the tests that already exist?

packages/shared/src/shared/ecsql-snippets/ECSqlJoinSnippets.ts
```diff
@@ -88,6 +120,8 @@ interface RelationshipPathJoinInfo {
    * relationship + target.
    */
   joins: RelationshipJoinInfo[];
+  /** Per-path-step resolution info, one entry per input path step, in path order. */
+  steps: RelationshipPathStepJoinInfo[];
```
grigasp, line 124: I'm wondering if we could make this structure more useful... Currently `joins` and `steps` are not related in any way, but, in fact, each step has one or more joins. So maybe we should restructure this a bit to have more value?

packages/content/src/content/ResolveContentSources.ts
```diff
@@ -76,17 +76,36 @@ function buildTargetFilter(target: ContentTarget): {
...
+// Resolves a relationship path into its JOIN clause plus the per-step concrete relationship/target
+// `ECClassId` selectors.
+async function resolveJoin(
+  schemaProvider: ECSchemaProvider,
+  path: JoinRelationshipPath,
+): Promise<{
+  joins: string;
+  bindings?: Record<string, ECSqlBinding>;
+  relationshipClassIdSelectors: string[];
+  targetClassIdSelectors: string[];
```
grigasp: Maybe have `selectors: Array<{ relationshipClassId: string; targetClassId: string }>`? That way we're sure that the number of them is the same. Also, at least to me, it reads better.

packages/content/src/content/ResolveContentSources.ts
```diff
@@ -117,19 +136,21 @@ const originalStrategy: ResolutionQueryStrategy = {
...
+      bindings: joinBindings,
+      relationshipClassIdSelectors,
+      targetClassIdSelectors,
+    } = await resolveJoin(schemaProvider, joinPath);
     const targetFilter = buildTargetFilter(target);
     const whereClause = targetFilter.where ? `WHERE ${targetFilter.where}` : "";
     const allBindings = { ...joinBindings, ...targetFilter.bindings };
     const ecsql = `
-      SELECT GROUP_CONCAT(DISTINCT ec_classname([this].[ECClassId], 's.c')), ${buildClassNameColumns(joinPath)}
+      SELECT GROUP_CONCAT(DISTINCT ec_classname([this].[ECClassId], 's.c')), ${buildClassNameColumns(targetClassIdSelectors)}, ${buildClassNameColumns(relationshipClassIdSelectors)}
```
grigasp: Probably not a big difference, but I'd personally prefer if we listed selectors in `rel1, tgt1, rel2, tgt2, ...` order instead of `...all_targets, ...all_rels`. This would probably be easier to achieve with my previous suggestion.

packages/content/src/test/Content.test.ts
```diff
@@ -19,21 +19,42 @@ import type {
...
         bindings: undefined,
-      })),
+      }),
+      createRelationshipPathJoinClause: (info: {
+        steps: Array<{ relationshipClassIdSelector: string }>;
+        bindings?: unknown;
+      }) => ({
+        joins: FIXED_JOINS,
+        bindings: info.bindings,
+        relationshipClassIdSelectors: info.steps.map((step) => step.relationshipClassIdSelector),
```
grigasp: doesn't look like `relationshipClassIdSelectors`  should be part of the result

.changeset/yummy-lands-think.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-shared": minor
```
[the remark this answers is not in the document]
grigasp: This is an addition to a return-only type, which is not directly exported. We don't consider such changes breaking.

packages/shared/src/test/ecsql-snippets/ECSqlJoinSnippets.test.ts
saskliutas: Removed separate tests

packages/shared/src/shared/ecsql-snippets/ECSqlJoinSnippets.ts
saskliutas: Moved `joins` into `RelationshipPathStepJoinInfo` so each step carries its joins

packages/content/src/content/ResolveContentSources.ts
saskliutas: Changed

packages/content/src/content/ResolveContentSources.ts
saskliutas: Changed

packages/content/src/test/Content.test.ts
saskliutas: Removed

packages/shared/src/test/ecsql-snippets/ECSqlJoinSnippets.test.ts
```diff
@@ -814,6 +829,91 @@ describe("createRelationshipPathJoinClause", () => {
...
+      it("selects source and target `ECClassId` for a link-table step", async () => {
```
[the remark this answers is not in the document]
saskliutas: Updated other tests to assert on selectors

grigasp (approved):

PR #1462, opened 2026-07-23
`content`: Base query building

grigasp:
Closes https://github.com/iTwin/presentation/issues/1409.

### DIFF
--- a/packages/content/src/content/Content.ts
+++ b/packages/content/src/content/Content.ts
@@ -76,7 +76,7 @@ interface ContentSortSpec {
 type ContentValueFilterTarget =
   | {
       /**
-       * The property field to filter on. Filter can only target primitive and struct properties.
+       * The property field to filter on. Filter can only target primitive, navigation and struct properties.
        * Array properties are not supported.
        */
       field: PropertyField;
--- a/packages/content/src/content/Content.ts
+++ b/packages/content/src/content/Content.ts
@@ -41,7 +41,9 @@ import type {
   ECSchemaProvider,
   ECSqlQueryExecutor,
   InstanceKey,
-  Value,
+  Point2dValue,
+  Point3dValue,
+  PrimitiveValue,
 } from "@itwin/presentation-shared";
 import type { ContentSource, ContentTarget } from "./ContentTarget.js";
 import type { DescriptorTransformer } from "./extensions/DescriptorTransformer.js";
@@ -66,13 +68,10 @@ interface ContentSortSpec {
 }
 
 /**
- * A value filter applied during query building (Stage 3).
- * Adds a WHERE clause to the final query — does not affect which fields
- * exist in the descriptor (only which rows are returned).
- *
+ * Attributes shared by all content value filters.
  * @public
  */
-export interface ContentValueFilter {
+interface ContentValueFilterBase {
   /** The field to filter on. */
   field: PropertyField | CalculatedField;
   /**
@@ -81,11 +80,37 @@ export interface ContentValueFilter {
    * Omit for scalar fields.
    */
   member?: string;
-  /** The filter operator. */
-  operator: ValueFilterOperator;
-  /** The value(s) to compare against. */
-  value: Value;
 }
+
+/** @public */
+type ScalarValueFilterOperator = Exclude<ValueFilterOperator, "is-null" | "is-not-null" | "is-in" | "is-not-in">;
+
+/**
+ * A value filter applied during query building (Stage 3).
+ * Adds a WHERE clause to the final query — does not affect which fields
+ * exist in the descriptor (only which rows are returned).
+ *
+ * @public
+ */
+export type ContentValueFilter =
+  | (ContentValueFilterBase & {
+      /** The filter operator. */
+      operator: ScalarValueFilterOperator;
+      /** The scalar value to compare against. */
+      value: PrimitiveValue;
+    })
+  | (ContentValueFilterBase & {
+      /** The filter operator. */
+      operator: "is-in" | "is-not-in";
+      /** The values to compare against. */
+      value: Exclude<PrimitiveValue, Point2dValue | Point3dValue>[];
+    })
+  | (ContentValueFilterBase & {
+      /** The filter operator. */
+      operator: "is-null" | "is-not-null";
+      /** Null checks do not accept values. */
+      value?: never;
+    });
 
 /** @public */
 type ValueFilterOperator =
@@ -98,7 +123,8 @@ type ValueFilterOperator =
   | "greater-than"
   | "greater-than-or-equal"
   | "like"
-  | "is-in";
+  | "is-in"
+  | "is-not-in";
 
 /**
  * Request options passed alongside the descriptor when loading values.
--- a/packages/content/src/content/Content.ts
+++ b/packages/content/src/content/Content.ts
@@ -103,8 +103,8 @@ export type ContentValueFilter =
   | (ContentValueFilterTarget & {
       /** The filter operator. */
       operator: ScalarValueFilterOperator;
-      /** The scalar value to compare against. */
-      value: PrimitiveValue;
+      /** The scalar value to compare against. Points are filtered per-coordinate via `member`, so a whole-point value is not accepted. */
+      value: Exclude<PrimitiveValue, Point2dValue | Point3dValue>;
     })
   | (ContentValueFilterTarget & {
       /** The filter operator. */
--- a/packages/content/src/content/Content.ts
+++ b/packages/content/src/content/Content.ts
@@ -68,8 +68,9 @@ interface ContentSortSpec {
 }
 
 /**
- * The field a value filter targets. `member` (composite access) is only valid for property fields —
- * calculated fields are scalar expressions, so the type disallows a member on them.
+ * The field targeted by a value filter. Filters may target calculated or property fields.
+ * When targeting a struct or point property field, specify `member` to identify the member to compare.
+ *
  * @public
  */
 type ContentValueFilterTarget =
--- a/packages/content/src/content/Content.ts
+++ b/packages/content/src/content/Content.ts
@@ -74,12 +74,16 @@ interface ContentSortSpec {
  */
 type ContentValueFilterTarget =
   | {
-      /** The property field to filter on. */
+      /**
+       * The property field to filter on. Filter can only target primitive and struct properties.
+       * Array properties are not supported.
+       */
       field: PropertyField;
       /**
        * For composite property fields (structs, points), the member to compare.
        * Example: `"x"` for a Point3d field, `"Street"` for an Address struct.
        * Omit for scalar fields.
+       * For struct properties and `Point2d`/`Point3d`, the member must be provided — filtering on a struct or point as a whole is not supported.
        */
       member?: string;
     }
--- a/packages/content/src/content/Content.ts
+++ b/packages/content/src/content/Content.ts
@@ -68,19 +68,26 @@ interface ContentSortSpec {
 }
 
 /**
- * Attributes shared by all content value filters.
+ * The field a value filter targets. `member` (composite access) is only valid for property fields —
+ * calculated fields are scalar expressions, so the type disallows a member on them.
  * @public
  */
-interface ContentValueFilterBase {
-  /** The field to filter on. */
-  field: PropertyField | CalculatedField;
-  /**
-   * For composite fields (structs, points), the member to compare.
-   * Example: `"x"` for a Point3d field, `"Street"` for an Address struct.
-   * Omit for scalar fields.
-   */
-  member?: string;
-}
+type ContentValueFilterTarget =
+  | {
+      /** The property field to filter on. */
+      field: PropertyField;
+      /**
+       * For composite property fields (structs, points), the member to compare.
+       * Example: `"x"` for a Point3d field, `"Street"` for an Address struct.
+       * Omit for scalar fields.
+       */
+      member?: string;
+    }
+  | {
+      /** The calculated field to filter on. Calculated fields are scalar, so `member` does not apply. */
+      field: CalculatedField;
+      member?: never;
+    };
 
 /** @public */
 type ScalarValueFilterOperator = Exclude<ValueFilterOperator, "is-null" | "is-not-null" | "is-in" | "is-not-in">;
@@ -93,19 +100,19 @@ type ScalarValueFilterOperator = Exclude<ValueFilterOperator, "is-null" | "is-no
  * @public
  */
 export type ContentValueFilter =
-  | (ContentValueFilterBase & {
+  | (ContentValueFilterTarget & {
       /** The filter operator. */
       operator: ScalarValueFilterOperator;
       /** The scalar value to compare against. */
       value: PrimitiveValue;
     })
-  | (ContentValueFilterBase & {
+  | (ContentValueFilterTarget & {
       /** The filter operator. */
       operator: "is-in" | "is-not-in";
       /** The values to compare against. */
       value: Exclude<PrimitiveValue, Point2dValue | Point3dValue>[];
     })
-  | (ContentValueFilterBase & {
+  | (ContentValueFilterTarget & {
       /** The filter operator. */
       operator: "is-null" | "is-not-null";
       /** Null checks do not accept values. */
--- a/packages/content/src/content/query/BaseQuery.ts
+++ b/packages/content/src/content/query/BaseQuery.ts
@@ -511,7 +511,10 @@ function resolveSelector(props: {
     }
     case "property": {
       // Composite access (e.g. a struct member or point `x`) addresses a member of the property column.
-      const memberSuffix = member ? `.[${member}]` : "";
+      // For a point property, emit the validated, canonical coordinate spelling; other composite
+      // members are addressed verbatim.
+      const resolvedMember = isPointType(field.type) ? resolvePointMember(field.type.type, member) : member;
+      const memberSuffix = resolvedMember ? `.[${resolvedMember}]` : "";
       const alias = resolvePropertyAlias({ field, relatedClassAliases, isRelationshipClass });
       return { selector: `[${alias}].[${field.propertyName}]${memberSuffix}`, type };
     }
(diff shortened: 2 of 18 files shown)


packages/content/src/content/query/TargetFilter.ts
```diff
@@ -0,0 +1,47 @@
...
+    const alias = target.instanceFilter.primaryClassAlias ?? PRIMARY_CLASS_ALIAS;
+    const expression = substituteExpressionAlias({
+      expression: target.instanceFilter.expression,
+      fromAlias: alias,
+      toAlias: PRIMARY_CLASS_ALIAS,
+    });
+    clauses.push(expression);
+    if (target.instanceFilter.bindings) {
+      Object.assign(bindings, target.instanceFilter.bindings);
+    }
```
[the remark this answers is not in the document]
saskliutas: Shouldn't be an issue. Only one binding might be present with `TARGET_FILTER_JOIN_ALIAS` name. It is unlikely it would be used by consumer as well.

packages/content/src/content/query/ValueFilters.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  props.filters.forEach((filter, filterIndex) => {
+    const resolved = props.resolveSelector(filter.field, filter.member);
+    // Navigation properties are structs whose target instance id lives in a `.Id` member, so filter
+    // against that member rather than the raw navigation column.
+    const selector =
+      filter.field.type.kind === "navigation" ? { ...resolved, selector: `${resolved.selector}.[Id]` } : resolved;
+    const clause = buildFilterClause({ filter, selector, filterIndex, bindings });
+    clauses.push(clause);
+  });
```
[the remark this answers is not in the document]
saskliutas: Updated to ignore member on navigation property.

packages/content/src/content/Content.ts
```diff
@@ -65,27 +67,57 @@ interface ContentSortSpec {
...
+      member?: string;
+    }
+  | {
+      /** The calculated field to filter on. Calculated fields are scalar, so `member` does not apply. */
+      field: CalculatedField;
+      member?: never;
+    };
+
+/** @public */
+type ScalarValueFilterOperator = Exclude<ValueFilterOperator, "is-null" | "is-not-null" | "is-in" | "is-not-in">;
```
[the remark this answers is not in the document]
saskliutas: This is expected

packages/content/src/content/Content.ts
```diff
@@ -65,27 +67,57 @@ interface ContentSortSpec {
...
-  /** The filter operator. */
-  operator: ValueFilterOperator;
-  /** The value(s) to compare against. */
-  value: Value;
-}
+export type ContentValueFilter =
+  | (ContentValueFilterTarget & {
+      /** The filter operator. */
+      operator: ScalarValueFilterOperator;
+      /** The scalar value to compare against. Points are filtered per-coordinate via `member`, so a whole-point value is not accepted. */
```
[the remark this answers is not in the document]
saskliutas: This package is not released yet. No need to document changes to API that is not released yet.

packages/content/src/content/Content.ts
```diff
@@ -65,27 +67,57 @@ interface ContentSortSpec {
   direction: "asc" | "desc";
 }
 
+/**
+ * The field a value filter targets. `member` (composite access) is only valid for property fields —
+ * calculated fields are scalar expressions, so the type disallows a member on them.
+ * @public
```
JonasDov: What do you think about rephrasing to something like:

```suggestion
 * The field targeted by a value filter. Filters may target calculated or property fields.
 * When targeting a struct or point property field, specify `member` to identify the member to compare.
 *
 * @public
```

packages/content/src/content/query/ValueFilters.ts
```diff
@@ -0,0 +1,126 @@
...
+   * automatically based on the field's type.
+   */
+  resolveSelector: (field: PropertyField | CalculatedField, member?: string) => ValueFilterSelector;
+}): { where?: string; bindings?: Record<string, ECSqlBinding> } {
+  const clauses: string[] = [];
+  const bindings: Record<string, ECSqlBinding> = {};
+
+  props.filters.forEach((filter, filterIndex) => {
+    // ignore `member` on navigation properties, since they are structs whose target instance id lives in a `.Id` member
+    const resolved = props.resolveSelector(filter.field, filter.field.type.kind === "navigation" ? "" : filter.member);
```
JonasDov: should `member` be `undefined` for navigation properties and not empty string?
```suggestion
    const resolved = props.resolveSelector(filter.field, filter.field.type.kind === "navigation" ? undefined : filter.member);
```

packages/content/src/content/query/ValueFilters.ts
```diff
@@ -0,0 +1,126 @@
...
+    // Navigation properties are structs whose target instance id lives in a `.Id` member, so filter
+    // against that member rather than the raw navigation column.
+    const selector =
+      filter.field.type.kind === "navigation" ? { ...resolved, selector: `${resolved.selector}.[Id]` } : resolved;
+    const clause = buildFilterClause({ filter, selector, filterIndex, bindings });
+    clauses.push(clause);
+  });
+
+  const where = clauses.length > 1 ? clauses.map((clause) => `(${clause})`).join(" AND ") : clauses[0];
+  return { ...(where ? { where } : undefined), ...(Object.keys(bindings).length > 0 ? { bindings } : undefined) };
```
JonasDov: A few notes:
1. As far as I understand if clauses is empty, then bindings is also empty. And there can't be a case where bindings is not empty and clauses is empty.
2. We might access array index which does not exist with `clauses[0]`. 
Maybe change to:

```suggestion
  const where = clauses.length === 1 ? clauses[0] : clauses.map((clause) => `(${clause})`).join(" AND ");
  if (where === "") {
    return {};
  }
  return { where, ...(Object.keys(bindings).length > 0 ? { bindings } : undefined) };
```

packages/content/src/content/query/TargetFilter.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  return {
+    ...(joins ? { joins } : undefined),
+    ...(clauses.length > 0 ? { where: clauses.join(" AND ") } : undefined),
```
JonasDov: Should we wrap clauses with `()` if there is more than one of them?

packages/content/src/content/query/QueryLimits.ts
```diff
@@ -0,0 +1,138 @@
...
+ */
+export const SQLITE_MAX_COMPOUND_SELECT = 500;
+
+/**
+ * Splits `terms` into chunks small enough to be UNION-ed into a single compound SQLite query without
+ * exceeding {@link SQLITE_MAX_COMPOUND_SELECT}. Each returned chunk has at most `maxPerChunk` terms.
+ *
+ * @internal
+ */
+export function chunkCompoundSelects<T>(terms: T[], maxPerChunk: number = SQLITE_MAX_COMPOUND_SELECT): T[][] {
```
JonasDov, line 34: Seems that this function is not used anywhere except tests. Do we need it / will we need it later?

packages/content/src/content/query/BaseQuery.ts
```diff
@@ -0,0 +1,672 @@
...
+  }
+  return `${quoted.slice(0, -1).join(", ")}, or ${quoted[quoted.length - 1]}`;
+}
+
+/**
+ * Merges `source` bindings into `target`, throwing on a duplicate name. Internal binding names are
+ * `ECSQL_PREFIX`-scoped and thus collision-free by construction, so a clash signals a real bug (e.g.
+ * a query filterer or instance filter reusing a reserved/duplicate parameter name).
+ */
+function mergeBindings(target: Record<string, ECSqlBinding>, source: Record<string, ECSqlBinding> | undefined): void {
```
JonasDov: could you make this function accept props object?

packages/content/src/content/Content.ts
```diff
@@ -65,27 +67,57 @@ interface ContentSortSpec {
...
+type ContentValueFilterTarget =
+  | {
+      /** The property field to filter on. */
+      field: PropertyField;
+      /**
+       * For composite property fields (structs, points), the member to compare.
+       * Example: `"x"` for a Point3d field, `"Street"` for an Address struct.
+       * Omit for scalar fields.
+       */
+      member?: string;
```
JonasDov, line 89: As I understand member **must** be provided for `Point2d`, `Point3d` and structs? If so, it should be documented.

packages/content/src/content/Content.ts
```diff
@@ -65,27 +67,57 @@ interface ContentSortSpec {
...
 
+/**
+ * The field a value filter targets. `member` (composite access) is only valid for property fields —
+ * calculated fields are scalar expressions, so the type disallows a member on them.
+ * @public
+ */
+type ContentValueFilterTarget =
+  | {
+      /** The property field to filter on. */
+      field: PropertyField;
```
JonasDov, line 82: Should probably mention that array property fields are not supported

packages/content/src/content/query/ValueFilters.ts
saskliutas: Cleaned up this piece to explicitly handle cases where no clauses are produced.

packages/content/src/content/query/TargetFilter.ts
saskliutas: There is always only single clause. Updated code accordingly.

packages/content/src/content/query/QueryLimits.ts
saskliutas: This will be needed in the next stage when fields are selected.

packages/content/src/content/query/BaseQuery.ts
saskliutas: Don't see a value in this change. this is function internal to the module that is not exported.

packages/content/src/content/Content.ts
saskliutas: Updated docs to specify the member is required for structs and points

packages/content/src/content/Content.ts
saskliutas: Updated docs to mention that array properties are not supported

JonasDov: comment says: `The property field to filter on. Filter can only target primitive and struct properties.` Aren't navigation properties supported?

packages/content/src/content/query/ValueFilters.ts
JonasDov: Do we want to return empty bindings object?

saskliutas: I don't see any value in returning `undefined` instead.

packages/content/src/content/Content.ts
saskliutas: Yes, updated docs. For some reason thought that primitive prop = navigation prop. 🫤

packages/content/src/test/query/BaseQuery.test.ts
```diff
@@ -0,0 +1,989 @@
...
+        includeRelatedJoins: true,
+        cardinalityHints,
+      });
+
+      expect(result.additional).to.be.undefined;
+      expect(result.anchor.paths.map((p) => p.path)).to.deep.equal([path]);
+    });
+  });
+
+  describe("value filters on split paths", () => {
```
JonasDov, line 866: In `BaseQuery.ts` line:
`const anchorJoinPaths = unionPaths([...anchorPaths.map((resolved) => resolved.path), **...filterPaths]);`
joins every filter path directly onto the anchor. When filter path has 1:Many relationship the query can produce duplicate rows. E.g. test:
```ts
it("[gap 1] builds an anchor query that may return the same primary once per matching related instance", async () => {
  const path = [makeStep(primaryClass, "TestSchema.RelMany", "TestSchema.Many")];
  const field = makePropertyField({
    propertyName: "Name",
    propertyClassName: "TestSchema.Many",
    pathFromTarget: path,
    valueClassNames: ["TestSchema.Many"],
  });

  const result = await buildBaseQuery({
    schemaProvider,
    source: makeSource([path]),
    includeRelatedJoins: true,
    filters: [{ field, operator: "is-equal", value: "A" }],
  });

  const anchorQuery = `
    SELECT [this].[ECInstanceId]
    ${result.anchor.parts.from}
    ${result.anchor.parts.joins}
    WHERE ${result.anchor.parts.where}
  `;
  expect(trimWhitespace(anchorQuery)).to.equal(
    trimWhitespace(`
      SELECT [this].[ECInstanceId]
      FROM [TestSchema].[Primary] [this]
      OUTER JOIN (
        SELECT [${ECSQL_PREFIX}r0].*
        FROM [TestSchema].[RelMany] [${ECSQL_PREFIX}r0]
        INNER JOIN [TestSchema].[Many] [${ECSQL_PREFIX}t0] ON [${ECSQL_PREFIX}t0].[ECInstanceId] = [${ECSQL_PREFIX}r0].[TargetECInstanceId]
      ) [${ECSQL_PREFIX}r0] ON [${ECSQL_PREFIX}r0].[SourceECInstanceId] = [this].[ECInstanceId]
      OUTER JOIN [TestSchema].[Many] [${ECSQL_PREFIX}t0] ON [${ECSQL_PREFIX}t0].[ECInstanceId] = [${ECSQL_PREFIX}r0].[TargetECInstanceId]
      WHERE [${ECSQL_PREFIX}t0].[Name] = :${ECSQL_PREFIX}vf0
    `),
  );
  // For example, assume TestSchema.Primary(0x1) is related through TestSchema.RelMany to both
  // TestSchema.Many(0x2, Name = "A") and TestSchema.Many(0x3, Name = "A"). The relationship JOIN
  // produces these rows before projection:
  //   this.ECInstanceId | r0.TargetECInstanceId | t0.Name
  //   0x1              | 0x2                  | A
  //   0x1              | 0x3                  | A
  // Both rows pass the Name filter, so SELECT [this].[ECInstanceId] returns [0x1, 0x1].
  // An existential or de-duplicated subquery would keep
  // the anchor at one row per TestSchema.Primary instance.
});
```

JonasDov, line 866: With a lot of filter paths there can be more than `SQLITE_MAX_JOIN_TABLES` JOIN statements. E.g.:
```ts
it("[gap 2] keeps filter-required joins within SQLite's table limit", async () => {
  const paths = Array.from({ length: 32 }, (_, index) => [
    makeStep(primaryClass, `TestSchema.Rel${index}`, `TestSchema.Target${index}`),
  ]);
  const filters: ContentValueFilter[] = paths.map((path, index) => ({
    field: makePropertyField({
      propertyName: "Name",
      propertyClassName: `TestSchema.Target${index}`,
      pathFromTarget: path,
      valueClassNames: [`TestSchema.Target${index}`],
    }),
    operator: "is-equal",
    value: "A",
  }));

  const result = await buildBaseQuery({
    schemaProvider,
    source: makeSource(paths),
    includeRelatedJoins: false,
    filters,
  });
  // Filter-required paths are added directly to the anchor after related paths are partitioned by
  // the join budget. Each link-table path adds two top-level OUTER JOINs: one derived relationship
  // table and one target table. With the primary FROM table, 32 paths therefore use 65 tables and
  // exceed SQLite's 64-table limit instead of being split into separate query groups.
  const outerJoinCount = result.anchor.parts.joins.match(/\bOUTER JOIN\b/g)?.length ?? 0;

  expect(1 + outerJoinCount).to.be.at.most(64);
});
```

packages/content/src/test/query/BaseQuery.test.ts
```diff
@@ -0,0 +1,989 @@
...
+        expression: "this.CodeValue || this.UserLabel",
+        selectorId: "calc",
+      };
+      const filters: ContentValueFilter[] = [{ field, operator: "like", value: "A%" }];
+
+      const result = await buildBaseQuery({ schemaProvider, source: makeSource([]), filters });
+
+      expect(result.anchor.parts.where).to.equal(`[this].CodeValue || [this].UserLabel LIKE :${ECSQL_PREFIX}vf0`);
+    });
+
```
JonasDov, line 592: Calculated field filters neither include the expression’s bindings nor parenthesize compound expressions, causing missing-parameter errors or incorrect operator precedence. E.g.:
```ts
it("[gap 3] includes calculated expression bindings in the base query", async () => {
  const field: CalculatedField = {
    kind: "calculated",
    id: "calc",
    label: "Calc",
    type: { kind: "primitive", type: "Double" },
    expression: "this.Length * :scale",
    bindings: { scale: { type: "double", value: 2 } },
    selectorId: "calc",
  };

  const result = await buildBaseQuery({
    schemaProvider,
    source: makeSource([]),
    filters: [{ field, operator: "greater-than", value: 10 }],
  });

  // The generated WHERE references both `scale` and `vf0`, but the current bindings contain only
  // `vf0`. The calculated field's `scale` binding must also be carried into the base query.
  expect({
    where: result.anchor.parts.where,
    bindings: result.anchor.parts.bindings,
  }).to.deep.equal({
    where: `[this].Length * :scale > :${ECSQL_PREFIX}vf0`,
    bindings: {
      [`${ECSQL_PREFIX}vf0`]: { type: "double", value: 10 },
    },
  });

  // Expected once calculated field bindings are included:
  // expect(result.anchor.parts.bindings).to.deep.equal({
  //   scale: { type: "double", value: 2 },
  //   [`${ECSQL_PREFIX}vf0`]: { type: "double", value: 10 },
  // });
});

it("[gap 3] parenthesizes a calculated expression before applying the filter operator", async () => {
  const field: CalculatedField = {
    kind: "calculated",
    id: "calc",
    label: "Calc",
    type: { kind: "primitive", type: "Boolean" },
    expression: "this.FlagA OR this.FlagB",
    selectorId: "calc",
  };

  const result = await buildBaseQuery({
    schemaProvider,
    source: makeSource([]),
    filters: [{ field, operator: "is-equal", value: true }],
  });

  // Currently generated: `[this].FlagA OR [this].FlagB = :vf0`, which compares only FlagB and then
  // ORs that result with FlagA. Expected: parenthesize the calculated expression before comparing it.
  expect(result.anchor.parts.where).to.equal(`[this].FlagA OR [this].FlagB = :${ECSQL_PREFIX}vf0`);

  // Expected once calculated expressions are parenthesized:
  // expect(result.anchor.parts.where).to.equal(`([this].FlagA OR [this].FlagB) = :${ECSQL_PREFIX}vf0`);
});
```

JonasDov (commented):
Asked AI to review, here are a few comments it left

packages/content/src/test/query/BaseQuery.test.ts
saskliutas: Fixed this

packages/content/src/test/query/BaseQuery.test.ts
saskliutas: Created issue for this case https://github.com/iTwin/presentation/issues/1464
This might require quite substantial changes and would be better to solve with full-stack-tests running actual queries.

saskliutas: Similar to the previous issue. Added this to issue https://github.com/iTwin/presentation/issues/1464 as it might be solved with the same solution.

JonasDov (approved):

PR #1481, opened 2026-08-13
`content-react`: Introduce package for consuming `presentation-content` in UI

### LINKED ISSUE #1474: `content-react`: Introduce package for consuming `presentation-content` in React applications
Introduce `content-react` package (similar to `hierarchies-react`) that would deliver utilities and custom hooks for convenient usage of `content` package in React applications.

saskliutas:
Closes https://github.com/iTwin/presentation/issues/1474

Initial setup of `@itwin/presentation-content-react` package. It builds on top of `@itwin/presentation-content` and provides utilities and hooks for consuming content in UI. Initial API provides `useInstanceFilterPropertiesInfo` for suing content descriptor to populate filter building UI.

- Added `presentation-content` based filter builder test component to test-app.
- Added basic BisCore fields provider to make sure content descriptor matches native content result more closely.


packages/content-react/src/content-react/InstanceFilterProperties.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export interface InstanceFilterProperty {
+  /** The descriptor field used when building a filter for this property. */
+  field: PropertyField;
+  /** Stable identity of the descriptor field. */
+  id: PropertyField["id"];
+  /** Text suitable for displaying the property. */
+  label: PropertyField["label"];
+  /** Value shape used by consumers to select operators and value editors. */
+  type: PropertyField["type"];
```
grigasp: Aren't `id`, `label` and `type` the same as the corresponding attributes on `field`? Is there value in having them duplicated here?

packages/content-react/src/content-react/InstanceFilterProperties.ts
```diff
@@ -0,0 +1,146 @@
...
+  /** Whether the property is reached through one or more relationships from the primary instance. */
+  isRelated: boolean;
```
grigasp: Similar to the above comment - it looks like `isRelated = !!field.pathFromTarget.length`

packages/content-react/src/test/InstanceFilterProperties.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+describe("createInstanceFilterProperties", () => {
+  it("uses the target primary class when source resolution has no concrete classes", () => {
+    const source = createSource({ resolvedPrimaryClasses: [], resolvedPaths: [] });
+    const directField = createPropertyField({ id: "direct", valueClassNames: [classA] });
+
+    const result = createInstanceFilterProperties({ descriptor: createDescriptor(source, [directField]) });
+
+    expect(result.classes).toEqual([{ name: classA, label: classA }]);
+    expect(result.properties[0].availableClassNames).toEqual([classA]);
```
grigasp: Looking at the test setup it looks like `classA` comes from the field, when in fact it comes from the source. I think it would be good to:
1. Pass `targetPrimaryClass` to `createSource`, even if it's going to match the default value.
2. Use a different class for `createPropertyField` to make it clear where the resulting class comes from.

packages/content-react/src/content-react/InstanceFilterProperties.ts
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Concrete primary classes that can supply the property.
+   *
+   * Direct fields use their descriptor value classes. Related fields use the target classes recorded
+   * on resolved relationship paths matching the field's path.
+   */
+  availableClassNames: EC.FullClassNameDotNotation[];
```
grigasp: Not sure "available" is the best term here 🤔 Maybe "primary" would be better?

packages/content-react/src/test/InstanceFilterProperties.test.ts
```diff
@@ -0,0 +1,152 @@
...
+      id: "related-a",
+      pathFromTarget: pathAC,
+      valueClassNames: [classB],
+    });
+    const relatedFieldFromB = createPropertyField({
+      id: "related-b",
+      pathFromTarget: pathBC,
+      valueClassNames: [classC],
+    });
+    const hiddenField = createPropertyField({ id: "hidden", hidden: true, valueClassNames: [classA] });
```
grigasp: do we need the hidden field in this test? looks like the "excludes hidden, non-property, and direct fields unavailable to source classes" test exercises that path.

apps/test-app/frontend/src/components/filter-builder/FilterBuilder.tsx
```diff
@@ -0,0 +1,216 @@
...
+    ...embeddedConfiguration,
+    imodelFieldsProviders: [...(embeddedConfiguration.imodelFieldsProviders ?? []), bisCoreContentCustomization],
+  };
+  const sources = await resolveContentSources({ imodelAccess, targets: [target], config: configuration });
+  const descriptor = await createContentProvider({
+    imodelAccess,
+    sources,
+    config: configuration,
+  }).getContentDescriptor();
+  return { descriptor: descriptor as ContentDescriptor };
```
grigasp: What do you think about defining a `ReadonlyContentDescriptor` in `@itwin/presentation-content` and using it as this function's return type, as well as input to `LoadedContentFilterBuilder` and `content-react` functions?

packages/content-react/src/content-react/InstanceFilterProperties.ts
grigasp: Also, do you think we should consider having this information to `PropertyField`? That way we wouldn't even need `InstanceFilterProperty` - we could use `PropertyField` directly.

apps/test-app/frontend/src/components/filter-builder/FilterBuilder.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const propertyRenderer = useCallback(
+    (id: string) => {
+      const property = visibleProperties.find((item) => item.id === id);
+      if (!property) {
+        return id;
+      }
+      return (
+        <span className="filter-builder-property">
+          {property.label}
+          {property.isRelated ? (
```
grigasp: what about direct properties that are categorized?

packages/content-react/src/content-react/InstanceFilterProperties.ts
saskliutas: Added `primaryClasses` to the `PropertiesField` in https://github.com/iTwin/presentation/pull/1489

packages/content-react/src/content-react/InstanceFilterProperties.ts
saskliutas: Removed

packages/content-react/src/content-react/InstanceFilterProperties.ts
saskliutas: Removed

packages/content-react/src/test/InstanceFilterProperties.test.ts
saskliutas: This test might be unnecessary after https://github.com/iTwin/presentation/pull/1489

packages/content-react/src/test/InstanceFilterProperties.test.ts
saskliutas: Removed

apps/test-app/frontend/src/components/filter-builder/FilterBuilder.tsx
saskliutas: Introduced `ReadonlyContentDescriptor` and `Readonly*Field` types to make it more convenient to pass content result around

apps/test-app/frontend/src/components/filter-builder/FilterBuilder.tsx
saskliutas: Changed to always use category.

packages/content/src/presentation-content.ts
```diff
@@ -5,8 +5,16 @@
...
+export type { ContentDescriptor, ReadonlyContentDescriptor } from "./content/model/ContentDescriptor.js";
+export type {
+  Field,
+  CalculatedField,
+  ExternalField,
+  ReadonlyField,
+  ReadonlyPropertyField,
+  ReadonlyCalculatedField,
+  ReadonlyExternalField,
+} from "./content/model/Field.js";
```
grigasp: maybe also export `PropertyField` for consistency?

packages/content-react/src/content-react/InstanceFilterProperties.ts
grigasp: Just to confirm, are you going to remove `InstanceFilterProperty` when https://github.com/iTwin/presentation/pull/1489 merges and this PR merges with `next`? Seems like quite a bit of code that's responsible for determining `availableClassNames` would go away.

saskliutas: Yes, I will cleanup this PR after https://github.com/iTwin/presentation/pull/1489 is merged.

packages/content/src/presentation-content.ts
saskliutas: It is exported. However, there is `PropertyField` namespace so `export type` does not work for it.

packages/content-react/src/content-react/InstanceFilterProperties.ts
saskliutas: Cleaned up to use `PropertyField.primaryClassNames`.

packages/content-react/src/content-react/InstanceFilterProperties.ts
```diff
@@ -57,30 +41,21 @@ export interface CreateInstanceFilterPropertiesProps {
...
  * @public
  */
 export function createInstanceFilterProperties({
   descriptor,
 }: CreateInstanceFilterPropertiesProps): InstanceFilterProperties {
   const candidateClassNames = getCandidateClassNames(descriptor.sources);
-  const candidateClassNameSet = new Set(candidateClassNames);
-
-  const properties = Object.values(descriptor.fields).flatMap((field): InstanceFilterProperty[] => {
+  const properties = Object.values(descriptor.fields).flatMap((field): ReadonlyPropertyField[] => {
```
grigasp: seems like we could use `filter` instead of `flatMap`

grigasp (approved):

packages/content-react/src/content-react/InstanceFilterProperties.ts
saskliutas: Changed to `filter`

PR #1491, opened 2026-08-19
Deprecate `ECClassHierarchyInspector` and `createCachingECClassHierarchyInspector`

### LINKED ISSUE #1488: `shared`: Consider dropping `ECClassHierarchyInspector` in favor of `ECSchemaProvider`
The new `ECSchemaProvider` implementation introduced in https://github.com/iTwin/presentation/pull/1394 already has the information, necessary for `ECClassHierarchyInspector.classDerivesFrom`, so it could provide the answers more quickly, without the possible need of a network request.

We should consider exposing the `classDerivesFrom` function on the `ECSchemaProvider` interface and removing `ECClassHierarchyInspector` and `createCachingECClassHierarchyInspector`.

grigasp:
Closes https://github.com/iTwin/presentation/issues/1488

### DIFF
--- a/.changeset/schema-provider-class-derives-from.md
+++ b/.changeset/schema-provider-class-derives-from.md
@@ -1,0 +1,46 @@
+---
+"@itwin/presentation-core-interop": minor
+"@itwin/presentation-hierarchies": major
+"@itwin/presentation-shared": major
+---
+
+`ECSchemaProvider`: Added a `classDerivesFrom` method for checking whether one ECClass is the same as, or derives from, another. `createECSchemaProvider` (in `@itwin/presentation-core-interop`) implements it using the class hierarchy information it already loads, so the answer is returned synchronously once the hierarchy has been loaded and no additional round-trips to the iModel are needed.
+
+Also, `ECClassHierarchyInspector` and `createCachingECClassHierarchyInspector` have been deprecated. Because `ECSchemaProvider` now exposes `classDerivesFrom` directly, a separate class hierarchy inspector is no longer needed when setting up iModel access:
+
+```ts
+// Before:
+import { createCachingECClassHierarchyInspector } from "@itwin/presentation-shared";
+import { createECSchemaProvider, createECSqlQueryExecutor } from "@itwin/presentation-core-interop";
+
+const schemaProvider = createECSchemaProvider(imodel);
+const imodelAccess = {
+...schemaProvider,
+...createCachingECClassHierarchyInspector({ schemaProvider }),
+...createECSqlQueryExecutor(imodel),
+};
+
+// After:
+import { createECSchemaProvider, createECSqlQueryExecutor } from "@itwin/presentation-core-interop";
+
+const imodelAccess = {
+...createECSchemaProvider(imodel), // now also provides `classDerivesFrom`
+...createECSqlQueryExecutor(imodel),
+};
+```
+
+**Breaking changes:**
+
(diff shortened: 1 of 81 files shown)


.changeset/schema-provider-class-derives-from.md
```diff
@@ -0,0 +1,46 @@
...
+...createECSchemaProvider(imodel), // now also provides `classDerivesFrom`
+...createECSqlQueryExecutor(imodel),
+};
+```
+
+**Breaking changes:**
+
+- `ECSchemaProvider` now requires a `classDerivesFrom` method. Objects created via `createECSchemaProvider` get it automatically, so most consumers don't need to change anything. Only custom, hand-written `ECSchemaProvider` implementations need to add the method.
+
+- Renamed the `classHierarchyInspector` prop to `imodelAccess` on `createClassBasedInstanceLabelSelectClauseFactory`, `createBisInstanceLabelSelectClauseFactory` (both in `@itwin/presentation-shared`) and `createPredicateBasedHierarchyDefinition` (in `@itwin/presentation-hierarchies`). The prop's type is unchanged, so the value passed to it doesn't need to change - only the prop name:
```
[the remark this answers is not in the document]
grigasp: 1. The two packages with breaking changes got `major` bump, so the changelog will show up under "Major" section, highlighting the fact that these are breaking changes.
2. The 2 breaking changes are under "Breaking changes" topic - no need to have individual "Breaking:" prefix.

.changeset/schema-provider-class-derives-from.md
```diff
@@ -0,0 +1,46 @@
...
+...createECSqlQueryExecutor(imodel),
+};
+
+// After:
+import { createECSchemaProvider, createECSqlQueryExecutor } from "@itwin/presentation-core-interop";
+
+const imodelAccess = {
+...createECSchemaProvider(imodel), // now also provides `classDerivesFrom`
+...createECSqlQueryExecutor(imodel),
+};
```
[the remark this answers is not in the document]
grigasp: fixed

packages/core-interop/src/core-interop/Metadata.ts
```diff
@@ -101,6 +113,19 @@ export function createECSchemaProvider(
       const svSchema = schemaView.getSchema(name);
       return svSchema ? createECSchemaFromSchemaView(svSchema, { schemaView, classHierarchyResolver }) : undefined;
     },
+    classDerivesFrom(
+      derivedClassFullName: EC.FullClassNameDotNotation,
+      candidateBaseClassFullName: EC.FullClassNameDotNotation,
+    ): Promise<boolean> | boolean {
+      // A class always derives from itself. This matches the semantics of the deprecated `ECClassHierarchyInspector`.
+      if (derivedClassFullName === candidateBaseClassFullName) {
```
JonasDov: Should we `localeLowerCase` in this check?

packages/core-interop/src/core-interop/Metadata.ts
```diff
@@ -101,6 +113,19 @@ export function createECSchemaProvider(
       const svSchema = schemaView.getSchema(name);
       return svSchema ? createECSchemaFromSchemaView(svSchema, { schemaView, classHierarchyResolver }) : undefined;
     },
+    classDerivesFrom(
+      derivedClassFullName: EC.FullClassNameDotNotation,
+      candidateBaseClassFullName: EC.FullClassNameDotNotation,
+    ): Promise<boolean> | boolean {
```
JonasDov: Why not always return a promise?

JonasDov (approved):

packages/core-interop/src/core-interop/Metadata.ts
grigasp: I think let's leave it as-is and move that conversation to https://github.com/iTwin/presentation/issues/1492.

PR #1507, opened 2026-08-25
`content`: Implement `getInstanceKeys` functionality and associated tests for content provider

### LINKED ISSUE #1412: `content`: `getInstanceKeys` SELECT builder
**Goal:** `SELECT ECInstanceId, ec_classname(ECClassId)` over the base query — no descriptor needed.

**Work:** SELECT primary key columns over the base query; map rows → `InstanceKey`.

**Tests:** keys shape with/without filters; multi-source.

**Dependencies:** Stage 1 sources (no descriptor), #1409.

grigasp:
Closes https://github.com/iTwin/presentation/issues/1412.

Moves `eachValueFrom` from `presentation-hierarchies`, where it was an internal utility, to `presentation-shared`, where it becomes `@public`, so we can reuse it in other packages. Note: the API is exposed without `rxjs` dependency (it's still added to `package.json` as a devDependency for tests).

### DIFF
--- a/.changeset/fresh-streams-yield.md
+++ b/.changeset/fresh-streams-yield.md
@@ -1,0 +1,5 @@
+---
+"@itwin/presentation-shared": minor
+---
+
+Add `eachValueFrom` for consuming subscribable streams through async iteration.
--- a/apps/full-stack-tests/src/content/GetInstanceKeys.test.ts
+++ b/apps/full-stack-tests/src/content/GetInstanceKeys.test.ts
@@ -1,0 +1,97 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { createContentProvider, resolveContentSources } from "@itwin/presentation-content";
+import { buildTestECDb } from "../ECDbUtils.js";
+import { initialize, terminate } from "../IntegrationTests.js";
+import { importSchema } from "../SchemaUtils.js";
+import { createContentIModelAccess, getPropertyFieldsByName } from "./Utils.js";
+
+import type { InstanceKey } from "@itwin/presentation-shared";
+
+async function collect<T>(values: AsyncIterable<T>): Promise<T[]> {
+  const result: T[] = [];
+  for await (const value of values) {
+    result.push(value);
+  }
+  return result;
+}
+
+function expectKeys(actual: InstanceKey[], expected: InstanceKey[]) {
+  expect(actual).toHaveLength(expected.length);
+  expect(actual).toEqual(expect.arrayContaining(expected));
+}
+
+describe("Content", () => {
+  describe("getInstanceKeys", () => {
+    beforeAll(async () => {
+      await initialize();
+    });
+
+    afterAll(async () => {
+      await terminate();
+    });
+
+    it("gets matching keys across sources and returns concrete polymorphic classes", async () => {
+      using setup = await buildTestECDb(async (builder, testName) => {
+        const schema = await importSchema(
+          testName,
+          builder,
+          `
+            <ECEntityClass typeName="A">
+              <ECProperty propertyName="Score" typeName="int" />
+            </ECEntityClass>
+            <ECEntityClass typeName="B">
+              <ECProperty propertyName="Score" typeName="int" />
+            </ECEntityClass>
+            <ECEntityClass typeName="Base" modifier="Abstract" />
+            <ECEntityClass typeName="Derived">
+              <BaseClass>Base</BaseClass>
+            </ECEntityClass>
+          `,
+        );
+        const a1 = builder.insertInstance(schema.items.A.fullName, { score: 1 });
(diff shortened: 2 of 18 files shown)


packages/shared/src/presentation-shared.ts
```diff
@@ -6,6 +6,7 @@
 export * as ECSql from "./shared/ecsql-snippets/index.js";
 
 export { ConcatenatedValue, ConcatenatedValuePart } from "./shared/ConcatenatedValue.js";
+export { eachValueFrom } from "./shared/EachValueFrom.js";
```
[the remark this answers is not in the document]
grigasp: nah, this is a common pattern in our packages

packages/content/src/content/query/GetInstanceKeys.ts
```diff
@@ -0,0 +1,60 @@
...
+              parts.where && `WHERE ${parts.where}`,
+            ]
+              .filter((fragment) => fragment)
+              .join(" "),
+            bindings: parts.bindings,
+          },
+          { rowFormat: "Indexes" },
+        );
+        return from(reader).pipe(
+          map((row) => ({ id: row[0], className: normalizeFullClassName(row[1]) })),
```
JonasDov: Do we need to normalize here? I think the returned className format is defined above with `'s.c'`?

packages/content/src/content/query/GetInstanceKeys.ts
```diff
@@ -0,0 +1,60 @@
...
+        }),
+      ),
+      mergeMap(({ anchor: { parts } }) => {
+        const reader = props.imodelAccess.createQueryReader(
+          {
+            ecsql: [
+              "SELECT [this].[ECInstanceId], ec_classname([this].[ECClassId], 's.c')",
+              parts.from,
+              parts.joins,
+              parts.where && `WHERE ${parts.where}`,
```
JonasDov: Is it expected that `parts.from` and `parts.joins` has `FROM` and `JOIN` added, when `parts.where` does not have `WHERE`

packages/content/src/content/query/GetInstanceKeys.ts
```diff
@@ -0,0 +1,60 @@
...
+      mergeMap(async (source) =>
+        buildBaseQuery({
+          schemaProvider: props.imodelAccess,
+          source,
+          queryFilterers: props.queryFilterers,
+          filters: props.filters,
+        }),
+      ),
+      mergeMap(({ anchor: { parts } }) => {
+        const reader = props.imodelAccess.createQueryReader(
```
JonasDov: A new query will be executed for each source. If two sources have the produce the same keys, then duplicate values will be returned by this function. Should we do something about this?

.changeset/fresh-streams-yield.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/presentation-shared": minor
+---
+
+Add `eachValueFrom` for consuming subscribable streams through async iteration.
```
JonasDov: What about `getInstanceKeys` change. It is not mentioned.

apps/full-stack-tests/src/content/GetInstanceKeys.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { createContentProvider, resolveContentSources } from "@itwin/presentation-content";
+import { buildTestECDb } from "../ECDbUtils.js";
+import { initialize, terminate } from "../IntegrationTests.js";
+import { importSchema } from "../SchemaUtils.js";
+import { createContentIModelAccess, getPropertyFieldsByName } from "./Utils.js";
+
+import type { InstanceKey } from "@itwin/presentation-shared";
+
+async function collect<T>(values: AsyncIterable<T>): Promise<T[]> {
```
JonasDov: I think we have at least one more implementation of `collect` in `CreateContentProvider.test.ts‎`. Should we add this function to `test-utilities` package?

.changeset/fresh-streams-yield.md
grigasp: the change is for an unreleased package - no need to mention

packages/content/src/content/query/GetInstanceKeys.ts
grigasp: Good question. 

Since the number of instance keys could be huge, I don't want to keep a set of all retrieved keys in memory, and that's the only way to ensure uniqueness if we're handling this with separate query per source.

The only solution is to `UNION ALL` all queries and use grouping on top. That's doable, but adds complication and is likely to make the performance worse. 

So for now I chose to document that duplicates are possible. We can revisit if someone sees that as a problem.

packages/content/src/content/query/GetInstanceKeys.ts
grigasp: that's how `buildBaseQuery` returned them, but I changed it to return all `from`, `join` and `where` with appropriate clauses

packages/content/src/content/query/GetInstanceKeys.ts
grigasp: agree, no need

apps/full-stack-tests/src/content/GetInstanceKeys.test.ts
grigasp: changed to use the one from `test-utilities`

JonasDov (approved):

PR #1517, opened 2026-08-28
Stabilize performance benchmarks by forcing GC before each measurement

grigasp:
Backporting https://github.com/iTwin/presentation/pull/1515.

The hierarchy benchmarks showed a bimodal ~1.5-2x run-to-run variance. Its cause is a V8 GC promotion cascade: leftover setup garbage in the young generation occasionally gets promoted to old space when the measured load starts churning short-lived per-node objects, bloating old space and making every subsequent Scavenge an order of magnitude costlier.

Force a full GC after setup (before the timed section) so each measurement starts from a clean heap, and run the test scripts with `--expose-gc` so `global.gc` is available. Also trigger the benchmark workflows on perf tests' `package.json` changes so this change is exercised in CI.

### DIFF
--- a/.github/workflows/benchmark-hierarchies.yml
+++ b/.github/workflows/benchmark-hierarchies.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - apps/performance-tests/hierarchies/**
       - apps/performance-tests/util/**
+      - apps/performance-tests/package.json
       - packages/core-interop/**
       - packages/hierarchies/**
       - packages/models-tree/**
--- a/.github/workflows/benchmark-hierarchies.yml
+++ b/.github/workflows/benchmark-hierarchies.yml
@@ -6,8 +6,8 @@ on:
     branches:
       - master
     paths:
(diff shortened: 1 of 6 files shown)


JonasDov (changes requested):
Hierarchies benchmark tests seem to have gotten worse. Is that expected?

grigasp:
> Hierarchies benchmark tests seem to have gotten worse. Is that expected?

Not really, I ran the tests locally with/without gc an gc'ed runs were always faster. So at this point it looks like a different machine / different time kind of problem.

I would add that the last run was historically very quick:

| test | diagram |
| --- | --- |
| filtering filters with 50000 paths  | image  |
| hide if no children required to finalize root, w/o children  | image |
| models tree full (Baytown)  | image  |
| models tree creates initial filtered view for 50k target items  | image |
| grouping by multiple attributes  | image  |
| flat 50k elements list  | image |

JonasDov (approved):

